PR #3, opened 2026-06-19
chore: update domain name

ckerr:
use transmissiontorrent.com instead of transmissionbt.com


ckerr:
note: `webapp / test-generated-files (pull_request)` is failing because this PR modifies `package.json` to change the domain name. This is intentional & nonblocking

PR #17, opened 2026-06-23
refactor: remove `tr_lib_init()`

ckerr:
`tr_lib_init()` is only there to ensure that curl gets initialized. Let's just lazy-load it instead when we're about to use curl for the first time.


tearfur (changes requested):
This will partially re-introduce upstream issue 5698. It's not safe to initialize curl after creating threads on some systems.

PR #27, opened 2026-06-25
chore: bump libb64

tearfur:
Raised minimum CMake version.

Fix these warnings:

```
CMake Deprecation Warning at third-party/libb64/CMakeLists.txt:1 (cmake_minimum_required):
  Compatibility with CMake < 3.10 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
  to tell CMake that the project requires at least <min> but has been updated
  to work with policies introduced by <max> or earlier.
```

```
CMake Warning (dev) at third-party/libb64/src/CMakeLists.txt:1 (add_library):
  Policy CMP0069 is not set: INTERPROCEDURAL_OPTIMIZATION is enforced when
  enabled.  Run "cmake --help-policy CMP0069" for policy details.  Use the
  cmake_policy command to set the policy and suppress this warning.

  INTERPROCEDURAL_OPTIMIZATION property will be ignored for target 'libb64'.
```


ckerr:
@tearfur Why was CMake 3.16.3 chosen?

It's not a blocker if there's good reason, but CMake 3.10.0 came out in Nov 2017 and 3.16.3 came out in Jan 2020, so _if_ it was an arbitrary choice, that's 2+ years of CMake versions that we could be supporting w/o risk wrt libb64's build scripts.

tearfur:
In our project the highest CMake version requirement among all submodules and the main project is the effective version requirement.

And our highest effective CMake version requirement is 3.16.3.

ckerr:
Makes sense to me, I hadn't thought to look.  3.16.3 isn't even new to libb64 -- wide-integer already requires it

| Package | cmake_minimum_required |
| --- | --- |
| dht | 3.10 |
| fast_float | 3.9 |
| fmt | 3.8...3.28 |
| googletest | 3.16 |
| libb64 | 3.16.3 FATAL_ERROR |
| libdeflate | 3.7 |
| libevent | 3.10 FATAL_ERROR |
| libevent/test-export | 3.10 |
| libnatpmp | 3.5 |
| libpsl | 3.10 |
| librecycle | 3.16 |
| libutp | 3.12 FATAL_ERROR |
| miniupnp/miniupnpc | 3.14 FATAL_ERROR |
| sigslot | 3.10 |
| small | 3.15 |
| utfcpp | 3.14...3.27 |
| wide-integer | 3.16.3 |

PR #29, opened 2026-06-25
build: skip C/C++ formatting checks by default if version unmatch

tearfur:
Cherry pick upstream PR 8925.

<details>
<summary>Original PR description</summary>

Closes `#8864`.

This PR aims to solve the problems contributors might face when they have a different `clang-format` release than the one we're targeting (currently 20).

- `clang-format` result might have changed under the same config over releases. Contributors might have their pre-commit hook fail on existing code that's passing formatting checks in our CI.

With this PR, `code_style.sh` will by default skip C/C++ formatting checks/fixes if `clang-format` version does not match our target, and print a warning in the console instead. You can force the check anyway by adding `--force`.

The script used to use whatever `clang-format` version it found and the result does not always match the version we're targeting in our CI.

P.S. `#7542` does something similar to this PR and I was against it back then. Now I changed my mind because I am feeling first hand how troublesome it is to deal with mismatched versions. Yes I know, I'm a man of unwavering beliefs and excellent foresight :)
</details>

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -105,7 +105,7 @@ jobs:
         id: check-for-diffs
         working-directory: .
         run: |
-          ./code_style.sh
+          ./code_style.sh --force
           set +e
           git diff --exit-code > style.diff
           echo "differs=$?" >> $GITHUB_OUTPUT
--- a/code_style.sh
+++ b/code_style.sh
@@ -1,7 +1,6 @@
 #!/usr/bin/env bash
 
-# Usage: ./code_style.sh
-# Usage: ./code_style.sh --check
+# Usage: ./code_style.sh [--check] [--force]
 
 set -o noglob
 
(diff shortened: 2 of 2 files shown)


tearfur:
@ckerr This PR mainly affects developers, and so I thought I should get a second opinion from you.

ckerr:
> @ckerr This PR mainly affects developers, and so I thought I should get a second opinion from you.

:thinking:

What I really care about is that there's a simple way to get the right version of clang-format into developers' hands so that they're not blocked and don't have to jump through a lot of hoops.

I don't have strong feelings about this PR but am _slightly_ against since IMO it attacks a symptom rather than the root cause.

I asked Claude what it thinks and it came up with https://github.com/ssciwr/clang-format-wheel, a well-maintained project that has installable clang-tidy versions and apparently works on macOS, Windows, and nix. Maybe we could try that in `code_style.sh`?

1. replace `clang_format_exe_names` with a hardcoded explicit version eg `20.1.8`
2. match on the exact version, not just the major version as we currently do
3. if we don't find the right version, spit back a message saying 'Try `pip install clang-format==20.1.8`'?

*Edited to Add:* I don't feel strongly about the pip approach either. If you prefer this PR and want to merge it, I'm OK with that.

tearfur:
I can make the script print some suggestions as to where to install `clang-format` conveniently, but I want to keep `code_style.sh`'s default behaviour loose.

We have the code style CI as the final guardrail anyway.

PR #39, opened 2026-06-28
fix: don't throw in `tr_dht_impl` destructor

tearfur:
Extract the code in `~tr_dht_impl()` to a deleter since they can throw exceptions.

There should be no change in behaviour as long as no exceptions are thrown.


tearfur:
Closing this one in favour of #50 since deleters in `std::unique_ptr` are supposed to be `noexcept`.

PR #55, opened 2026-06-30
feat: recent paths

ckerr:
Support a 'recent paths' feature in RPC, GTK, and Qt.

- The GTK client had this before, but it was broken by 9f77ef9c7ae32c1dcb7be79fca338f3ea1fb62fe when we added support for `Gtk::FileChooserNative`, which doesn't guarantee support for shortcut folders.
- First time supported in the Qt client
- First time supported for remote/RPC clients

Notes: The "Add torrent" and "Move torrent" dialogs'  "recent directories" feature is now supported on all desktop clients.


ckerr:
CC @tearfur for review on the libtransmission/ parts (and whatever else you want to review, for that matter)

libtransmission-app/prefs.h
```diff
@@ -197,6 +197,8 @@ class Prefs final
     StringType script_torrent_done_filename_;
     StringType script_torrent_done_seeding_filename_;
     StringType socket_diffserv_;
+    std::vector<StringType> recent_download_paths_ = { download_dir_ }; // must be declared after download_dir_
+    std::vector<StringType> recent_relocate_paths_ = { download_dir_ }; // must be declared after download_dir_
```
tearfur: Just a question, as this doesn't cause bugs AFAICT.

Why do these 2 and `download_dir_` need to mirror the default value from `session-settings.h` instead of default initializing like the others?

tearfur (approved):
`libtransmission*/` and docs LGTM.

libtransmission-app/prefs.h
ckerr: Answered in https://github.com/transmissiontorrent/transmission/pull/72 which I'll land before this PR

PR #65, opened 2026-07-02
build: bump to clang-tidy 22 on Linux

tearfur:
Fix new warnings, disable new checks we don't need and checks that are bugged.

### DIFF
--- a/libtransmission-app/app.h
+++ b/libtransmission-app/app.h
@@ -3,8 +3,7 @@
 // or any future license endorsed by Mnemosaic LLC.
 // License text can be found in the licenses/ folder.
 
-#ifndef TR_APP_APP_H
-#define TR_APP_APP_H
+#pragma once
 
 namespace tr::app
 {
@@ -13,5 +12,3 @@ namespace tr::app
 void init();
 
 } // namespace tr::app
-
-#endif // TR_APP_APP_H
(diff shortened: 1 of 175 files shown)


libtransmission-app/app.h
```diff
@@ -3,7 +3,8 @@
 // or any future license endorsed by Mnemosaic LLC.
 // License text can be found in the licenses/ folder.
 
-#pragma once
+#ifndef TR_APP_APP_H
+#define TR_APP_APP_H
```
ckerr: If you feel strongly about `portability-avoid-pragma-once` I'll go along with it, but my vote is to keep `#pragma once`.

The clang-tidy rationale for `portability-avoid-pragma-once` is to support older compilers; but we're already requiring C++20, this doesn't affect us.  The `#ifndef -> #define -> #endif` pattern is error-prone.

ckerr (changes requested):
LGTM other than the pragma question. OK to merge if you feel strongly about it

libtransmission-app/app.h
tearfur: No strong feeling about it, I was just following the conclusion we reached at https://github.com/transmission/transmission/pull/7877#discussion_r2599034697.

Changed back to `#pragma once` now.

cb0e87839d3645dd57fb8c2a6c474779b7d84cc3
198921bce4813277163aa48db83cb61e4a535db9
99bd2d6ec0d0bfb734f3ea78ca816ed15484c03d

ckerr: Oog I forgot I said that, fair enough 🫠

ckerr (approved):

PR #77, opened 2026-07-04
deps: bump bundled libdeflate to v1.25 from v1.20

ckerr:
When not using an OS-provided copy of libdeflate, default to a bundled libdeflate v1.25 instead of v1.20.

See [NEWS.md](https://github.com/transmissiontorrent/libdeflate/blob/master/NEWS.md) for the list of changes. Most changes listed between 1.20 and 1.25 are `semver/patch`, the rest are `semver/minor`

Notes: Bumped bundled version of [libdeflate](https://github.com/ebiggers/libdeflate) to v1.25 from v1.20 for upstream bugfixes.


ckerr:
marked as `notes:none` since libdeflate was removed in #106

PR #91, opened 2026-07-09
refactor: change crypto utils to use `std::span`

tearfur:
Fixes `cppcoreguidelines-pro-bounds-pointer-arithmetic` clang-tidy warnings in `crypto-utils*` code.

I haven't made up my mind whether to enable `cppcoreguidelines-pro-bounds-pointer-arithmetic` globally yet. My opinion on pointer arithmetic is a bit similar to `goto`: Most of the time they are fragile, hard-to-maintain code, but sometimes they are the only tool that does the job well.

There should be no change in behaviour.

### DIFF
--- a/libtransmission/crypto-utils-ccrypto.cc
+++ b/libtransmission/crypto-utils-ccrypto.cc
@@ -154,13 +154,11 @@ tr_sha256_digest_t tr_sha256::finish()
 
 // ---
 
-bool tr_rand_buffer_crypto(void* buffer, size_t length)
+bool tr_rand_buffer_crypto(std::span<std::byte> const buffer)
 {
-    if (length == 0) {
+    if (buffer.empty()) {
         return true;
     }
 
-    TR_ASSERT(buffer != nullptr);
-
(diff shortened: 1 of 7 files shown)


ckerr (approved):

ckerr:
> I haven't made up my mind whether to enable cppcoreguidelines-pro-bounds-pointer-arithmetic globally yet. My opinion on pointer arithmetic is a bit similar to goto: Most of the time they are fragile, hard-to-maintain code, but sometimes they are the only tool that does the job well.

I agree with all that. Also, since this codebase was all C until a few years ago, there's bound to be leftover pointer arithmetic that's been shipping for a decade without any problem.

I'm fine with enabling this warning (and most other warnings), but `cppcoreguidelines-pro-bounds-pointer-arithmetic` isn't a high priority to me personally.

tearfur:
The more `cppcoreguidelines-pro-bounds-pointer-arithmetic` warnings I look at, the more they look overly restrictive to me. I'm not going to enable it.

Issue #104, opened 2026-07-11
Migrate blocklist downloading + unzipping into libtransmission


ckerr:
This is a subissue of "Support using transmission-gtk as a remote control" #82

One more step of unifying system settings is to have a consistent blocklist-download-and-install behavior across the macOS, GTK, and Qt UIs.`BlocklistDownloader.mm` does unzip/untar by calling shell commands. This approach won't work cross-platform, so let's add  [libarchive](https://github.com/transmissiontorrent/libarchive) as a third-party dependency to do it. libarchive is well-maintained, uses CMake, and has a compatible license. It also ships pre-installed on macOS.

However, libarchive requires zlib. This was already an _implicit_ dependency via libcurl (all platforms) and Qt (transmission-qt). Let's make it an explicit dependency instead and drop libdeflate as redundant.

- [x] Make our zlib dep explicit, remove libdeflate #106
- [x] Add vendorized libarchive dep #108
- [x] Move `BlocklistDownloader.mm` to `libtransmission/`, add C API for macOS client #113
- [x] Expose new blocklist methods to RPC #133

tearfur:
Apparently macOS's blocklist downloader code will not work if the blocklist is hosted at localhost, see upstream issue 8892.

I'm guessing you plan to re-implement the `blocklist_update` RPC method with libarchive, while mimicking current macOS behaviour, so I'm mentioning this in case you're planning to migrate the macOS code to RPC 1-to-1.

PR #107, opened 2026-07-11
deps: blocklist pt 2: add libarchive

ckerr:
deps: add vendored libarchive dependency

Part [2](https://github.com/transmissiontorrent/transmission/pull/106) of issue #104 to move the macOS version of the blocklist downloader into libtransmission.

---
    
Add third-party/libarchive and pin to latest stable v3.8.8 for to make consistent, cross-platform tar/zip/gzip decompression available to libtransmission for the upcoming blocklist downloader.
    
CMake: cmake/Findlibarchive.cmake plus a tr_add_external_auto_library block that builds libarchive statically with only the tar/zip/raw formats and the gzip (system zlib) filter enabled; linked PRIVATE into libtransmission. The parent prefix path is forwarded so the sub-build's find_package(ZLIB) resolves the Windows-staged zlib.
    
Xcode: a new static-library "archive" target compiling the same 123 sources the CMake build selects, driven by a hand-authored Darwin config.h under third-party/libarchive-config/ (HAVE_CONFIG_H). Linked into libtransmission with a target dependency.


ckerr:
Wrong branch

PR #109, opened 2026-07-12
perf: simplify tr_arc4

ckerr:
Minor optimizations to the `tr_arc4` header:

- Use `std::span` instead of pointer-and-length
- Use `std::swap` and `std::itoa` instead of bespoke code
- Remove an unused constructor
- Hoist `i_` and `j_ ` into locals across the loop in `process()`, giving about 5% throughput gain in `-O2` on all encrypted peer traffic. *eta:* more info @ https://github.com/transmissiontorrent/transmission/pull/109#discussion_r3566502647 

AI disclosure: I have reviewed and tested all changes in this PR.

### DIFF
--- a/libtransmission/tr-arc4.h
+++ b/libtransmission/tr-arc4.h
@@ -5,8 +5,11 @@
 #pragma once
 
 #include <array>
-#include <cstddef> // size_t
+#include <cstddef> // size_t, std::byte
 #include <cstdint> // uint8_t
+#include <numeric> // std::iota
+#include <span>
+#include <utility> // std::swap
 
 /**
  * This is a tiny and reusable implementation of alleged RC4 cipher.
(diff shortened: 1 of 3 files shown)


libtransmission/tr-arc4.h
```diff
@@ -26,55 +29,48 @@
...
+    // tgt must hold at least std::size(src) bytes; src and tgt may be equal
+    constexpr void process(std::span<std::byte const> src, std::span<std::byte> tgt)
     {
-        for (size_t i = 0; i != n_bytes; ++i) {
-            tgt[i] = src[i] ^ arc4_next();
+        auto i = i_;
+        auto j = j_;
+
+        for (size_t k = 0, n = std::size(src); k < n; ++k) {
+            tgt[k] = src[k] ^ std::byte{ next(i, j) };
```
tearfur, line 53: Can be simplified further.

```suggestion
            tgt[k] = src[k] ^ std::byte{ next(i_, j_) };
```

libtransmission/tr-arc4.h
```diff
@@ -26,55 +29,48 @@
...
-    {
-        auto const tmp = s_[i];
-        s_[i] = s_[j];
-        s_[j] = tmp;
-    }
-
-    constexpr uint8_t arc4_next()
+    // state is passed by reference so that process() can keep it
+    // in locals across its hot loop instead of touching members
+    constexpr uint8_t next(uint8_t& i, uint8_t& j)
```
tearfur, line 70: After simplifying `process()`, `next()` no longer needs to take arguments, as all call sites are passing in `i_` and `j_`.

tearfur (commented):
The changes look correct as-is.

Just a few suggestion to simplify the code further.

libtransmission/tr-arc4.h
ckerr: I had the same instinct you did, but this hoisting is intentional. I should have added a code comment to make this explicit. I'll do that now.

Here's the AI's finding, in its usual too-verbose style:

> Consistent +4–5% across three runs, identical ciphertext. 
>
> The short version
uint8_t is unsigned char on every mainstream platform, and character types are the one family the C++ aliasing rules say may alias anything. So when process() writes a byte through tgt, the compiler must assume that store might have overwritten this->i_ or this->j_. That forces it to reload and re-store the cipher state from memory on every iteration of the loop. Copying i_/j_ into locals removes that obligation, because a local whose address never escapes provably can't be aliased by any pointer.
>
>The longer version
> 1. What the compiler is obligated to assume. The compiler must generate code that is correct for every legal caller — including a pathological one like cipher.process(buf, n, reinterpret_cast<uint8_t*>(&cipher)), which aims the output buffer at the cipher's own state. Normally, type-based alias analysis lets it rule this out: a store through a double* can't legally modify an int, so the compiler ignores the possibility. But the standard ([basic.lval]/11) explicitly exempts char, unsigned char, and std::byte — a store through those may touch any object. Since uint8_t is unsigned char, the store tgt[k] = ... in the original loop is, as far as the compiler can prove, a potential write to i_, j_, or s_.
>
> 2. What that costs in the loop. In the original (tr-arc4.h:51), each iteration reads and writes the members i_/j_ and then stores to tgt. Because that store may have clobbered the members, any register copies of i_/j_ are potentially stale at the top of the next iteration — so the compiler reloads them from memory. Symmetrically, their updated values must be flushed to memory before the store. Here's GCC 14 at -O2; %rdi is this, so 256(%rdi) is i_ and 257(%rdi) is j_:
>
> That's 2 loads + 2 stores per byte that exist purely to defend against a caller that doesn't exist. In the hoisted version the loop body has zero accesses to offsets 256/257 — i and j live in registers, and GCC writes both back with a single 16-bit store after the loop (movw %dx, 256(%rdi)).
>
> 3. One honest nuance for the reviewer. In the pathological self-aliasing call, the hoisted version would behave differently from the original — that's precisely why the compiler isn't allowed to make this transformation and a human is: it's a contract decision ("tgt never points into the cipher object") that only the author can sign off on. It's also not a novel trick — OpenSSL's rc4_enc.c uses exactly this formulation (x = key->x; y = key->y; ... key->x = x; key->y = y;) for the same reason.
>
> Here's the tab-friendly pair, both live and verified:
>
> Variant A — original, state in members: https://godbolt.org/z/afrKaazjG
> Variant B — hoisted, state in locals: https://godbolt.org/z/YWcTeezxM
> I set them up specifically so the tab-flip comparison reads cleanly:
>
> Both pages define the same function name (arc4::process), the same class, the same compiler and flags (GCC 16.1, -O2 -std=c++20), and a near-identical header comment — so when you flip tabs, the page layout stays put and your eye lands on the only thing that actually changed.
What the flip shows: in variant A's loop, four instructions touching BYTE PTR [rdi+256] / [rdi+257] (the i_/j_ reloads at the top and flushes before the tgt[k] store); in variant B's loop, those offsets never appear — the state lives in registers and is written back once after the loop.
Variant A's header hyperlinks to variant B (godbolt short URLs are content-addressed, so a true mutual link is impossible — B's header just points at "the link shared with this one").
Before minting, I verified through godbolt's compile API that these exact pages on that exact compiler produce the claimed assembly: 4 state accesses inside A's loop, 0 inside B's.

ckerr: Added in  e07072f:

```diff
     // tgt must hold at least std::size(src) bytes; src and tgt may be equal
     constexpr void process(std::span<std::byte const> src, std::span<std::byte> tgt)
     {
+        // See https://github.com/transmissiontorrent/transmission/pull/109
+        // tldr: the compiler thinks the in-loop assignment call could
+        // overwrite `i_` or `j_`, so it reloads them each time we loop.
+        // Hosting them into locals avoids that work.
         auto i = i_;
         auto j = j_;
```

tearfur: Damn the AI is citing some deep magic.

Makes sense though.

tearfur (approved):

PR #111, opened 2026-07-12
refactor: convert `tr_torrentStat()` to use `std::span`

tearfur:
What the title says.


ckerr (approved):
LGTM.

One thing, WDYT about using "Notes: Moved from C++17 to C++20." for release notes? That's the text we've used on about 20 other PRs, and `std::span`ification seems to fit that category.

If you don't think it fits, the existing no-notes is fine

tearfur:
I forgot about that. I will update my other `std::span` PRs.

PR #113, opened 2026-07-12
feat: blocklists pt 3: add untar/unzip support in libt

ckerr:
Part [3](https://github.com/transmissiontorrent/transmission/pull/106) of issue #104 to move the macOS version of the blocklist downloader into libtransmission.

This PR:

- Adds untar/unzip support for downloaded blocklists in libtransmission
- Adds automatic blocklist updating (eg weekly) to libtransmission
- Removes the macOS-only blocklist downloader. The macOS UI now uses libtransmission for this.

So this has a lot of changes going on under the hood, but (if everything was done correctly) no user-visible changes yet. Just that untar/unzip is happening in `libtransmission/` instead of in `macosx/` now.

RPC will be migrated in part 4.

*edit:* `+1,460 -711` looks scary but the size of the production code is mostly unchanged, just moving ~400 LOC from `macosx/` to `libtransmission/`. The big increase is coming from new test coverage.

Notes: Tarred and gzipped blocklist downloads are now supported on all desktop clients.


ckerr:
CC @tearfur not sure if this refactor is interesting to you or not? Review welcomed but not required

libtransmission/blocklist-download.cc
```diff
@@ -0,0 +1,384 @@
...
+} // namespace
+
+// Decompress a downloaded blocklist. Supports txt, tar, gz, tgz formats.
+// Omit other libarchive formats for YAGNI, we can add later if/when needed.
+std::string decompress(std::string_view const body)
+{
+    auto* const arc = archive_read_new();
+    if (arc == nullptr) {
+        return {};
+    }
```
tearfur, line 62: Wrap `arc` into a unique pointer so you have freedom to early return anywhere and flatten all those nested braces.

```suggestion
    auto* const arc = archive_read_new();
    if (arc == nullptr) {
        return {};
    }
    auto const arc_uniq = std::unique_ptr<struct archive, decltype(&archive_read_free)>{ arc, archive_read_free };
```

libtransmission/blocklist-download.cc
```diff
@@ -0,0 +1,384 @@
...
+    archive_read_support_filter_gzip(arc);
+    archive_read_support_format_tar(arc);
+    archive_read_support_format_zip(arc);
+    archive_read_support_format_raw(arc);
```
tearfur: The libarchive wiki and man pages both warn against enabling the `raw` format with any other formats.

https://github.com/libarchive/libarchive/wiki/FormatRaw#caveat-dont-mix-_raw_-with-other-handlers:

> If you are using the raw handler, you should generally not enable any other handler

ARCHIVE_READ_FORMAT(3):

> Using the “raw” handler together with any other handler will often work but can produce surprising results.

tearfur (changes requested):
I've taken long enough. Here's the first batch of comments.

libtransmission/blocklist-download.cc
ckerr: Yep, that does simplify the return branches. Updated in 219e4498f4cf0df4bf11cba952d90d0e2d7897f4

libtransmission/blocklist-download.cc
ckerr: I missed that. Good catch. Fixed in c099bcfbefeeac8f8d7da1594bc4070fe293d7a0

libtransmission/blocklist-download.cc
```diff
@@ -0,0 +1,429 @@
...
+    struct archive_entry* entry = nullptr;
+    for (;;) {
+        // ARCHIVE_WARN still hands back a usable entry; only EOF or a fatal error stops us
+        if (auto const header = archive_read_next_header(arc, &entry); header != ARCHIVE_OK && header != ARCHIVE_WARN) {
+            break;
+        }
+
+        // a blocklist is a single file; skip directories and take the
+        // first file we find
+        if (archive_entry_filetype(entry) == AE_IFDIR) {
```
tearfur:
```suggestion
        if (archive_entry_filetype(entry) != AE_IFREG) {
```

I'd rather we find out we need to add more file types rather than accidentally reading from somewhere we shouldn't have.

libtransmission/blocklist-download.h
```diff
@@ -0,0 +1,96 @@
...
+// (gzip, tar (possibly gzipped), zip) and return the first regular file's
+// contents; a plain, unwrapped list is returned as-is. Exposed here so the
+// format matrix can be tested directly, without a session or a download.
+[[nodiscard]] std::string decompress(std::string_view body);
+
+// Normalize a configured blocklist URL for fetching: trim surrounding
+// whitespace and, when no scheme is present, assume https (a bare host like
+// "list.example.com/bl" would otherwise fail at libcurl). Returns empty for an
+// empty or whitespace-only URL. Exposed here so it can be tested directly.
+[[nodiscard]] std::string normalize_blocklist_url(std::string_view url);
```
tearfur: Is there a bug report somewhere documenting a case curl's scheme guessing doesn't cover? Otherwise I don't see the point of having this.

libtransmission/blocklist-download.h
```diff
@@ -0,0 +1,96 @@
...
+    // at all if cancel() is called first.
+    void update(tr_blocklist_update_func on_done);
+
+    // Abandon the most recent in-flight update: its `on_done` will not fire.
+    void cancel();
+
+    // (Re)arm or disarm the periodic auto-update timer to match the
+    // current settings (blocklist enabled + URL set + updates enabled).
+    // Safe to call from any thread.
+    void restart_timer();
```
tearfur: We should rename this method to reflect that it can also stop the timer.

libtransmission/blocklist-download.cc
```diff
@@ -0,0 +1,429 @@
...
+    }
+
+    auto const content = decompress(response.body);
+
+    // don't use empty responses
+    if (!has_rule_lines(content)) {
+        result.status = tr_blocklist_update_status::InvalidData;
+        deliver(*pending, result);
+        return;
+    }
```
tearfur: This defeats the purpose of 5cea498eb66795c541474f1376016112d66ce8f8.

To justification for 5cea498eb66795c541474f1376016112d66ce8f8 was that there were 2 separate issues from upstream that specifically asked for the ability to unblock all peers by serving an empty (but valid) blocklist.

- 7942
- 8074

Having said that, reading the conversation history again, looks like I forgot to ask about their use-case.

If you want to keep this behaviour, IMO:

1. We probably should figure why those 2 people wanted that behaviour.
2. The change should be made in `parseFile()` in `blocklist.cc` instead of scattering+duplicating some of the logic here. In general, `blocklist-download.cc` should not concern itself with the content of the downloaded file IMO.

P.S. This check let's through a file containing only meaningless garbage, which will then be rejected in `blocklist.cc`.

libtransmission/blocklist-download.cc
```diff
@@ -0,0 +1,429 @@
...
+
+    auto error = std::string{};
+    if (auto const n_rules = pending->mediator->set_blocklist_content(content, error); !n_rules) {
+        result.status = tr_blocklist_update_status::SaveError;
+        result.error = std::move(error);
+    } else if (*n_rules == 0U) {
+        result.status = tr_blocklist_update_status::InvalidData;
```
tearfur: Noting that this is somewhat related to my previous comment.

What we do with this branch depends on what we decide about empty blocklists. So I'm just leaving this here for now.

libtransmission/blocklist-download.cc
```diff
@@ -0,0 +1,429 @@
...
+        url.remove_prefix(2U);
+    }
+    return fmt::format("https://{:s}", url);
+}
+
+// Per-request state, kept alive by the tr_web fetch callback during the fetch.
+struct Pending {
+    Updater::Mediator* mediator = nullptr;
+    tr_blocklist_update_func on_done;
+    bool cancelled = false;
```
tearfur: If you want to save some space, looks like `cancelled` can be subsumed into `on_done`, since all `cancelled` does is prevent invoking `on_done`.

OTOH, keeping `cancelled` might make debugging the states easier. So this is entirely up to your preference.

libtransmission/blocklist-download.cc
```diff
@@ -0,0 +1,429 @@
...
+            fmt::arg("error_code", err.code()));
+        return std::nullopt;
+    }
+
+    auto const n_rules = tr_blocklistSetContent(&session_, filename);
+    tr_sys_path_remove(filename);
+
+    // edge case: nullopt from the parser means "no valid rules".
+    // Collapse it to 0 so the Updater reports InvalidData
+    return n_rules.value_or(0U);
```
tearfur: Related to my comment about empty blocklists.

libtransmission/blocklist-download.h
```diff
@@ -0,0 +1,96 @@
...
+    // Superseded status if a newer update() takes over first. It is not invoked
+    // at all if cancel() is called first.
+    void update(tr_blocklist_update_func on_done);
+
+    // Abandon the most recent in-flight update: its `on_done` will not fire.
+    void cancel();
+
+    // (Re)arm or disarm the periodic auto-update timer to match the
+    // current settings (blocklist enabled + URL set + updates enabled).
+    // Safe to call from any thread.
```
tearfur: I had to spend some time to figure this out.

Might be worthy to note down that this is thread-safe *with the condition* that only one other thread (a.k.a. the main thread) will ever invoke this method, which is the case currently.

The risk with more than one thread calling this method is with `tr_sessionClose()`. The lambda that this method queues might end up being executed *after* `this` was tore down.

CMakeLists.txt
```diff
@@ -579,6 +579,17 @@ tr_add_external_auto_library(ARCHIVE libarchive
         -DENABLE_CNG:BOOL=OFF
         -DENABLE_WIN32_XMLLITE:BOOL=OFF)
 
+# archive.h decorates its API with __declspec(dllimport) on Windows unless this
+# is defined. We always link the static libarchive, so tell consumers not to
+# expect a DLL (harmless elsewhere; only __declspec on MSVC cares).
+set_property(TARGET libarchive::libarchive APPEND PROPERTY INTERFACE_COMPILE_DEFINITIONS LIBARCHIVE_STATIC)
```
tearfur: Modern syntax.

```suggestion
target_compile_definitions(libarchive::libarchive
    INTERFACE
        LIBARCHIVE_STATIC)
```

CMakeLists.txt
```diff
@@ -579,6 +579,17 @@ tr_add_external_auto_library(ARCHIVE libarchive
...
+
+if(WIN32)
+    # libarchive's archive_util.c seeds its RNG with BCryptGenRandom (bcrypt.h),
+    # so anything linking the static archive.lib must also link bcrypt.
+    set_property(TARGET libarchive::libarchive APPEND PROPERTY INTERFACE_LINK_LIBRARIES bcrypt)
```
tearfur: Modern syntax.

```suggestion
    target_link_libraries(libarchive::libarchive
        INTERFACE
            bcrypt)
```

libtransmission/blocklist-download.h
ckerr: ... I think there's still some value here but it's in the wrong place.

RFC 3986 requires a scheme for absolute URLs, but `tr_urlParse()` currently doesn't require them. Let's fix  `tr_urlParse()` and remove this normalize method.

libtransmission/blocklist-download.cc
ckerr: 5154d0e578784a814b40073dac18d1a6625e59f9

libtransmission/blocklist-download.h
ckerr: 6c712709bc8cbc69c4a21da005cf3a05350ad818

libtransmission/blocklist-download.h
ckerr: de4c2284f0d9946a07dd27b32c85aed99d8e06e4

libtransmission/blocklist-download.cc
ckerr: I *think* I've followed the intent of 5cea498 here in 3bf1554f587c077b9627e508ebbb5e15f1569278 but I'm not positive & would appreciate a second look here

libtransmission/blocklist-download.cc
ckerr: let's go with YAGNI for now. c564285221f532bddaf970d215c81ecc5f7d8524

libtransmission/blocklist-download.h
ckerr: Updated in 096a47e548673ed08b7a57b5f3f64dc846a06aa8 to be a little clearer. I still don't love this but am not sure how to make this simpler w/o dropping important info. :man_shrugging:

CMakeLists.txt
ckerr: fc609bac1dc92bd8b034a59a91427bbaea01b0ac

CMakeLists.txt
ckerr: fc609bac1dc92bd8b034a59a91427bbaea01b0ac

tests/libtransmission/blocklist-download-test.cc
```diff
@@ -0,0 +1,660 @@
...
+    return out;
+}
+
+void asGzip(archive* arc)
+{
+    archive_write_set_format_raw(arc);
+    archive_write_add_filter_gzip(arc);
+}
+
+void asTar(archive* arc)
```
tearfur: We should test other tar variants as well. Especially gnutar, since it is the default format created by GNU tar.

There's also v7tar and pax.

libtransmission/blocklist-download.cc
tearfur: I think this is right now. 👍🏻

tests/libtransmission/blocklist-download-test.cc
ckerr: gnutar, v7tar, ustar added in 26f4c180de4be20b607546a2225a095b13260645

tgz added in dc0288c76d26436973614ec7562e43d6077e202b

tearfur: Sweet.

tearfur (approved):
libtransmission, CMake and test code LGTM.

P.S. Hopefully my review comments was worth the 2 weeks wait, because I was already full sending it.

PR #117, opened 2026-07-13
refactor: convert peer mgr API to take `std::span` parameters

tearfur:

There should be no change in behaviour.

### DIFF
--- a/libtransmission/peer-mgr.h
+++ b/libtransmission/peer-mgr.h
@@ -614,11 +614,6 @@ struct tr_pex {
     uint8_t flags = 0;
 };
 
-constexpr bool tr_isPex(tr_pex const* pex)
-{
-    return pex != nullptr && pex->socket_address.is_valid();
-}
-
 [[nodiscard]] tr_peerMgr* tr_peerMgrNew(tr_session* session);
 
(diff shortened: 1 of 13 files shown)


libtransmission/peer-mgr.h
```diff
@@ -627,7 +622,7 @@ void tr_peerMgrFree(tr_peerMgr* manager);
 
 void tr_peerMgrAddIncoming(tr_peerMgr* manager, std::shared_ptr<tr_peer_socket> socket);
 
-size_t tr_peerMgrAddPex(tr_torrent* tor, tr_peer_from from, tr_pex const* pex, size_t n_pex);
+size_t tr_peerMgrAddPex(tr_torrent* tor, tr_peer_from from, std::span<tr_pex const> pex);
```
ckerr, line 648: needs an `#include <span>` at top of file

ckerr (approved):
One minor missing `#include <span>` but otherwise LGTM

libtransmission/peer-mgr.h
ckerr: Oh wait, I think it's already there iff #116  lands first :smile:

PR #121, opened 2026-07-14
fix: lost wakeup that could hang tr_web shutdown

ckerr:
Sample test error:

```
The following tests FAILED:
	554 - LT.WebTest.userAgentFromMediatorIsSent (Timeout)
Errors while running CTest
```

`~Impl()` and `startShutdown()` stored `deadline_ns_` and notified without holding `tasks_mutex_,` so the notify could be lost if it fired in the gap before the curl worker began its condvar wait. Worse, the wait treated a pending shutdown as a reason to keep sleeping, so nothing could ever wake it and `join()` hung forever.

Fixed by:

- Setting the deadline under the mutex
- Making a pending shutdown stop the wait
- Moving the exit check and fetch()'s deadline check under the mutex.

Added a stress test destroying a tr_web after each fetch.

This fixes a bug in production code; but since (I think?) it would only be exercised in cases where `tr_web` had a short lifespan, I think this only affects tests so I'm marking as `notes:none`.

### DIFF
--- a/libtransmission/web.cc
+++ b/libtransmission/web.cc
@@ -892,10 +892,9 @@ tr_web::tr_web(Mediator& mediator)
 {
 }
 
-tr_web::~tr_web()
-{
-    impl_->startShutdown(0ms);
-}
+// ~Impl() itself starts an immediate shutdown before joining the curl
+// thread, so there's nothing to do here beyond destroying the members
+tr_web::~tr_web() = default;
 
 std::unique_ptr<tr_web> tr_web::create(Mediator& mediator)
 {
--- a/libtransmission/web.cc
+++ b/libtransmission/web.cc
@@ -230,24 +230,30 @@ public:
 
     ~Impl()
     {
-        deadline_ns_ = to_ns(mediator.now());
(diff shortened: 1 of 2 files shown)


libtransmission/web.cc
```diff
@@ -755,17 +761,19 @@ class tr_web::Impl
...
-            }
-
             if (auto lock = std::unique_lock{ tasks_mutex_ }; lock.owns_lock()) {
-                // sleep until there's something to do
+                // sleep until there's something to do.
+                // NB: a pending shutdown must stop the wait; if it prolonged
+                // it instead, the loop could sleep through its exit condition
                 auto const stop_waiting = [this]() {
-                    return !is_idle() || !deadline_exists();
+                    return !is_idle() || deadline_exists();
```
tearfur, line 769: Adding a bit of context for the record.

Looks like the `deadline_exists()` condition was accidentally flipped in f03fc9270ba466348a4e9798386d36f8af8a6043, partly causing this regression.

tearfur (approved):

PR #128, opened 2026-07-15
ci: faster BSD tests

ckerr:
The emulated BSD jobs are three of the slowest bottlenecks.

This PR tries to improve this with these BSD-specific changes:

- The biggie is don't do Qt / GTK builds there. Neither UI has BSD-specific code, and there are other jobs already exercising GTK3, GTK4, Qt5, Qt6 builds. The real benefit to the `Sanity / BSD` runs is building libtransmission on BSD.
- Be explicit about how much memory & how many CPUs the emulator can use. (Based on the `ubuntu-latest` runner specs from GitHub Actions hosted runners docs)
- Don't upgrade packages when we don't need to.
- Don't compile with debug info; these binaries are built to be thrown away anyway.
- Don't link with LTO; these binaries are built to be thrown away anyway.

*edit:* piggybacking on this PR to fix a CI limit I hadn't expected: we're hitting GitHub's 10GB cache limit, which causes Qt to have to be rebuilt more often on Windows. :skull:  Switch the builds-we-are-never-going-to-use-for-testing-or-nightlies to `CMAKE_BUILD_TYPE=Release` to reduce their ccache size by removing debug info. Add ccache pruning rules.

### DIFF
--- a/.github/actions/prepare-deps-win32/action.yml
+++ b/.github/actions/prepare-deps-win32/action.yml
@@ -40,7 +40,7 @@ runs:
       id: restore-cache
       with:
         path: ${{ env.DEPS_PREFIX }}
-        key: ${{ github.job }}-${{ inputs.arch }}-${{ steps.cache-key.outputs.hash }}
+        key: win-deps-${{ inputs.arch }}-${{ steps.cache-key.outputs.hash }}
     - name: Build Dependencies
       if: steps.restore-cache.outputs.cache-hit != 'true'
       shell: pwsh
@@ -58,4 +58,4 @@ runs:
       id: cache
       with:
         path: ${{ env.DEPS_PREFIX }}
-        key: ${{ github.job }}-${{ inputs.arch }}-${{ steps.cache-key.outputs.hash }}
+        key: win-deps-${{ inputs.arch }}-${{ steps.cache-key.outputs.hash }}
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -1589,7 +1589,6 @@ jobs:
         uses: vmactions/freebsd-vm@v1
         with:
           usesh: true
-          cpu: 4
           mem: 12288
           prepare: |
             set -ex
@@ -1769,7 +1768,6 @@ jobs:
         uses: vmactions/openbsd-vm@v1
         with:
           usesh: true
-          cpu: 4
           mem: 12288
           prepare: |
             set -ex
@@ -1849,7 +1847,6 @@ jobs:
         uses: vmactions/netbsd-vm@v1
         with:
           usesh: true
-          cpu: 4
           mem: 12288
           prepare: |
             set -ex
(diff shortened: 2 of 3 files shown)


ckerr:
This should probably land before anything else, since the cache exhaustion issue is blocking CI

.github/workflows/actions.yml
```diff
@@ -1805,6 +1849,8 @@ jobs:
         uses: vmactions/netbsd-vm@v1
         with:
           usesh: true
+          cpu: 4
```
tearfur: AFAICT we don't need to manually specify CPU cores. They already automatically detect the number of cores and use them all.

This applies to all BSD runners.

https://github.com/anyvm-org/anyvm/blob/3cca386905e9f843270f3a80101c3adb50dd9e5e/anyvm.py#L2636-L2637

```
  --cpu <num>            Number of CPU cores (Default: host cores, capped at
                         8 with hardware acceleration, 2 under TCG).
```

https://github.com/anyvm-org/anyvm/blob/3cca386905e9f843270f3a80101c3adb50dd9e5e/anyvm.py#L4880

```python
    # Default configuration
    default_cpu = str(max(1, os.cpu_count() or 1))
    config = {
        'mem': "2048",
        'cpu': default_cpu,
```

.github/workflows/actions.yml
```diff
@@ -1863,13 +1901,42 @@ jobs:
...
           name: binaries-${{ github.job }}
           path: pfx/**/*
+
+  prune-ccache:
+    # The timestamped ccache keys used above are write-once: every push run
+    # saves a new snapshot, and the one it supersedes is never restored again
+    # but only gets deleted by LRU pressure at the repo's 10 GB cache cap --
+    # evicting still-useful caches (e.g. the prebuilt Windows deps) along the
+    # way. Sweep every ref and keep just the newest snapshot per key prefix.
+    # Runs only on pushes to main: PR runs from forks get a read-only token.
```
tearfur, line 1918: Making sure I'm reading this correctly. the idea is ccache entries will be pruned once a new entry superscedes it, or it is 7 days old, right?

ckerr: Yes, that's it exactly.

.github/workflows/actions.yml
ckerr: Good catch! Fixed e75ba7f0b18645390393c398c71db3db2c50a76d

.github/actions/prepare-deps-win32/action.yml
```diff
@@ -40,7 +40,7 @@ runs:
       id: restore-cache
       with:
         path: ${{ env.DEPS_PREFIX }}
-        key: ${{ github.job }}-${{ inputs.arch }}-${{ steps.cache-key.outputs.hash }}
+        key: win-deps-${{ inputs.arch }}-${{ steps.cache-key.outputs.hash }}
```
tearfur, line 43: Good catch. This should cut cache size for win deps by a big fraction with no tradeoff.

tearfur (approved):

PR #131, opened 2026-07-17
refactor: use vanilla css instead of sass

ckerr:
Our sass use was minimal: a few media-query mixins and some nesting. So just use vanilla css instead and drop sass. The media queries are now inlined & css nesting is flattened in esbuild for compatibility with older browsers.

This has some nice side-effects:

1. The only tool needed at packaging time is `esbuild`, which is ubiquitous, eg `sudo apt install esbuild`
1. `package.json.buildonly` and `.nvmrc` are moot & have been removed
1. `sass`, `stylelint-config-sass-guidelines`, `esbuild-sass-plugin` go away as devDependencies
1. We don't need to install Node.js anymore in CI
1. The CMake rules to rebuild the bundle don't need Node.js either


tearfur:
This PR is large, reviewing might take a few days.

tearfur:
Are there supposed to be any visible changes in the UI?

ckerr:
> Are there supposed to be any visible changes in the UI?

Shouldn't be, unless I made a mistake when migrating scss -> css. 

This PR should all just be

- scss -> css migration
- build process simplification made possible by removing scss
 - vendorizing the two JS deps to make life easier for Debian

CMakeLists.txt
```diff
@@ -74,7 +74,7 @@ option(ENABLE_DAEMON "Build daemon" ON)
 tr_auto_option(ENABLE_GTK "Build GTK client" AUTO)
 tr_auto_option(ENABLE_QT "Build Qt client" AUTO)
 tr_auto_option(ENABLE_MAC "Build Mac client" AUTO)
-tr_auto_option(REBUILD_WEB "Rebuild the web client's generated assets. Requires Node.js and network access." OFF)
+tr_auto_option(REBUILD_WEB "Rebuild the web client's generated assets. Requires esbuild." OFF)
```
tearfur:
```suggestion
option(REBUILD_WEB "Rebuild the web client's generated assets. Requires esbuild." OFF)
```

`REBUILD_WEB` doesn't need to use `tr_auto_option`, no sane build will explicitly set this option to `AUTO`.

In fact, it used to be `option` before 5fe95ad5b0ffe9e53c8bc8dd9082ec67ce9da807, where the commit changed its default to `AUTO`, before changing back to `OFF` shortly after at 7a686a53f3445fa6397c0abf2c0f66e2e2a5dcb9.

If we make this change, then...

CMakeLists.txt
```diff
@@ -723,35 +723,31 @@ target_compile_definitions(transmission::WideInteger
 
 set(TR_WEB_ASSETS ${PROJECT_SOURCE_DIR}/web/public_html)
 if(REBUILD_WEB)
```
tearfur: ...we don't need to handle all the extra steps that come with `tr_auto_option`, so we can simplify and subsume this entire `if(REBUILD_WEB)` block...

web/CMakeLists.txt
```diff
@@ -1,3 +1,20 @@
+if(CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR)
+    # standalone mode: `cmake -S web -B <build-dir>` builds just the webapp
+    # without configuring the rest of Transmission. CI uses this to check
+    # that the bundle can be rebuilt with nothing but cmake and esbuild.
+    cmake_minimum_required(VERSION 3.18 FATAL_ERROR)
+    project(transmission-web NONE)
+    set(TR_NAME transmission)
+    set(ESBUILD_MINIMUM 0.25.0) # keep in sync with the root CMakeLists.txt
+    find_program(ESBUILD esbuild REQUIRED)
```
tearfur: ...into here, and in the process dedupe the `ESBUILD_MINIMUM` definition.

tearfur (commented):
First batch of my comments. I might have more as I look into more parts of this PR.

CMakeLists.txt
ckerr: 5884568327f0b3edc8fa144867f1388331554b65

CMakeLists.txt
ckerr: OO this simplifies the whole thing quite a bit. :astonished: 

5884568327f0b3edc8fa144867f1388331554b65

web/CMakeLists.txt
ckerr: 5884568327f0b3edc8fa144867f1388331554b65

ckerr:
fedora-rawhide CI is red because of an upstream issue. Fix is @ https://github.com/transmissiontorrent/transmission/pull/132

tearfur (approved):

PR #133, opened 2026-07-18
feat: blocklists pt 4: use new blocklist updater in RPC; update Qt client

### LINKED ISSUE #104: Migrate blocklist downloading + unzipping into libtransmission

ckerr:
Fixes #104.

Part [4](https://github.com/transmissiontorrent/transmission/pull/106) of issue #104 to move the macOS version of the blocklist downloader into libtransmission.

- Add `update_blocklist_v2` method to RPC. The older v1 returns a rule count; v2 returns an object with a count and with status/error messages
- Update Qt client prefs dialog to use the `update_blocklist_v2`. Copy behavior and error strings from macOS client.
- Add Qt PrefsDialog tests.

Notes: The "Enable automatic blocklist updates" feature is now supported on all desktop clients.

Notes: Tarred and gzipped blocklist downloads are now supported on all desktop clients.


docs/rpc-spec.md
```diff
@@ -713,6 +716,28 @@ Request parameters: none
 
 Response parameters: a number `blocklist_size`
 
+Downloads the blocklist at `blocklist_url`, decompresses it (gzip, tar, and zip
+archives are handled, as is plain text), installs it, and returns the new rule
+count. On failure it returns a JSON-RPC error.
+
+#### 4.3.1 Blocklist update with a structured result
+Method name: `blocklist_update_v2`
```
tearfur: Do we need this new API? If the whole point was just to tell "superseded" apart from "succeeded", can we add a new JSON-RPC error for `blocklist_update` instead?

ckerr: This is a nice idea, I didn't think of that. :+1:

ckerr: e6e70b3d63d716ec20b00e702a1d3efafa5043b7

tearfur:
@ckerr Merge conflicts.

ckerr:
@tearfur rebased + force-pushed

docs/rpc-spec.md
tearfur: Oh, you dropped the "superseded" status altogether, I guess that works too.

What I had in mind was simply to add a new RPC error in `rpcimpl.h`.

tearfur (approved):

PR #28, opened 2026-06-25
chore: bump to clang-format 22

tearfur:
Raise clang-format 22 before PRs pile up and conflicts arise.

PR #137, opened 2026-07-20
chore: allow braceless single-line statement bodies

ckerr:
A followup to #28, when we change the brace style from Allman to Linux. This one is more opinionated, so I'm making it a standalone PR where it can get approved or rejected on its own.

tldr, I'd like to be able to omit braces in an `if` statement if the clause is only a single line:

```c++
if (foo) { // <-- braces required before this PR; still allowed with this PR
  bar();
}

if (foo) // <-- this PR: braceless allowed, but not required
  bar();

if (baz) { // <-- unchanged
  qux();
  thud();
}
```

This is 100% opinion and probably comes from me spending a lot of time in the Chromium codebase.

### DIFF
--- a/gtk/.clang-tidy
+++ b/gtk/.clang-tidy
@@ -18,6 +18,7 @@ Checks:
   - -cppcoreguidelines-pro-bounds-constant-array-index
   - -cppcoreguidelines-use-enum-class
   - hicpp-*
+  - -hicpp-braces-around-statements # alias of `readability-braces-around-statements`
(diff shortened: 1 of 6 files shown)


tearfur:
I am slightly against this PR. But I don't mind trying it out.

I came to appreciate the spacing provided by braces while working on Transmission, especially with the old style of always open brace on the next line.

ckerr:
To me this is a minor nice-to-have but it's not a big deal. If you're slightly against, I'm fine with closing this PR?

PR #143, opened 2026-07-22
ci: stop cache thrashing in GitHub Actions

ckerr:
Rework the GitHub Actions cache scripts to reduce cache misses & make CI runs faster.

The 10 GiB repo limit was getting filled with timestamped snapshots, causing longer CI runs because important cache items were crowded out.

- Save ccache snapshots only on `main`. PRs can read `main`'s cache, but `main` and other PRs cannot read PR-scoped caches. (This is a security rule set by GitHub.) Saving from PRs would duplicate the entire ccache in every PR.

- Use CCACHE_COMPRESSLEVEL to make the ccache items a little smaller.

- Remove superseded items from the cache.

- Verify each new ccache entry before deleting its predecessor. This preserves the last usable snapshot if an upload fails.

- Allow ccache families up to 1 GiB of room, needed to keep big jobs (eg Windows Qt) from purging items mid-build.

- Build cache keys from the jobs' platform, configuration, and IPO mode. This prevents a job from using cache items from incompatible configs.

- On Windows, also include the dependency type in cache keys to make the distinction between Qt builds and non-gui builds explicit.

- On Windows, use IPO only on the Windows x64 reference build. This way the reference build (which will become the nightly build) gets the optimization, but other CI jobs finish sooner.

- On macOS, run the from-tarball build w/bundled deps on macOS 26. The separate CMake job still builds on Intel with system dependencies, removing one Intel matrix leg while retaining both kinds of coverage.

- Use `DISCOVERY_MODE` and `DISCOVERY_TIMEOUT` in gtest to prevent a busy CI box from starving test discovery and failing with timeout.

- Install Windows dependency and packaging tools only as needed. This shortens warm-cache jobs and avoids unnecessary setup work.

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -11,7 +11,20 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
   cancel-in-progress: true
 
+permissions:
+  contents: read
+  # ccache jobs and the cache janitor delete superseded cache entries
+  actions: write
+
 env:
+  # See .github/CACHE_POLICY.md for cache classes, lifecycle, and budget.
+  CACHE_LIMIT_GIB: '10'
+  CACHE_WARNING_GIB: '9'
+  # Bumping CACHE_SCHEMA re-keys every ccache family; the janitor deletes
+  # keys that do not carry the current value.
+  CACHE_SCHEMA: cache-v2
+  CCACHE_COMPRESSLEVEL: '5'
+  CCACHE_DIR: ${{ github.workspace }}/.ccache
   GTEST_OUTPUT: xml:./
 jobs:
   what-to-make:
@@ -141,12 +154,7 @@ jobs:
           enable-qt: qt6
           use-sudo: true
       - name: Setup ccache
-        uses: hendrikmuhs/ccache-action@v1.2.23
-        with:
-          key: ${{ github.job }}
-          # only main's snapshots are saved; PR runs restore them via the
-          # base-branch fallback but must not multiply them per-ref
-          save: ${{ github.event_name == 'push' }}
+        uses: ./.github/actions/setup-ccache
       - name: Configure
         run: |
           cmake \
@@ -190,9 +198,8 @@ jobs:
           # disable container-overflow reports to avoid false errors
           # that are caused by Qt libraries not being asan instrumented
           extra-asan-options: detect_container_overflow=0
-      - name: Trim ccache
-        if: ${{ always() && github.event_name == 'push' }}
-        run: ccache --evict-older-than 7d
+      - name: Finalize ccache
+        uses: ./.github/actions/finalize-ccache
 
   sanitizer-tests-macos:
     runs-on: macos-26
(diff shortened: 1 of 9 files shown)


.github/workflows/actions.yml
```diff
@@ -868,8 +870,10 @@ jobs:
           name: binaries-${{ github.job }}-${{ matrix.arch }}-msi
           path: obj/dist/msi/*.msi
       - name: Trim ccache
-        if: ${{ always() && github.event_name == 'push' }}
-        run: ccache --evict-older-than 7d
+        # A successful build refreshes every live entry's timestamp, so
+        # anything older than a few days is stale; trim to stay under max-size.
+        if: ${{ success() && github.event_name == 'push' }}
```
tearfur:
```suggestion
        if: ${{ github.event_name == 'push' }}
```

tearfur (approved):
Small cosmetic nit, but LGTM otherwise.

tearfur (dismissed):
All these new changes aren't meant for this PR are they...? 😕

ckerr:
> All these new changes aren't meant for this PR are they...? 😕

Unfortunately they are.  I've been cycling on this on and off for hours. Basically we're trying to get about 25GB of cache data (ccache + ctcache + vcpkg + Qt builds on Windows + BSD images) down to under 10GB. So we're making PR CI runs only read cache, not write it (GitHub has a security measure that won't let `main` read cached data from PRs, so writing from both would double the storage use), adding a janitor to clean things in a consistent way everywhere, handling LTO timing issues on Windows that were causing timeouts, and dropping a redundant macOS build job. :scream: 

I've reviewed this whole PR and understand it, but this is *much* more vibe-codey than I'm comfortable with. I've been using agents for C++/ObjC++ code PRs, but usually *that* workflow is "I need A, B, C, D .... Z. Let's start with A... no not that way, change this, now let's move to B" micro-managing an agent to do what I tell it. This PR... this isn't that. This was me throwing my ideas, Sol's ideas, and Fable's ideas all against each other until both agents agreed that the result would be <10GB. The alternative would be to drop out more builds or architectures, but I don't see anything droppable  unless we want to drop one of the BSD flavors for about 600MB.

Fable thinks the expected cache size if this gets merged into `main` and runs CI there is about 9.1 GiB:

Projection from measured cache sizes:

| Family | Expected size |
| --- | ---: |
| 20 cache-v2 ccache families | 5.64 GiB |
| BSD VM images, one per OS/version | 1.82 GiB |
| Typed Windows dependencies, one per key | 0.56 GiB |
| Gradle and CodeQL | 0.27 GiB |
| ctcache | <0.001 GiB |
| Legacy untyped Windows dependencies | up to 0.84 GiB |
| **Immediate total** | **about 9.12 GiB** |

That leaves roughly **0.88 GiB** below the 10 GiB cap.

ckerr:
The PR is still complex, but I've tried to make it a little easier to digest:

1. Squashed it down to get rid of the iterative noise
2. Updated the commit message to be clearer
3. Updated the PR message to be clearer

ckerr:
@tearfur I would really like to land this one to (try to) improve CI's cache thrashing. So if it really is OK, then I'll take you up on your suggestion to merge it now.

Your reviews have been great though. If you do find any issues, feel free to create bug reports & assign to me if you want me to fix 'em :bow:

PR #158, opened 2026-07-28
fix: don't require quarks to be utf8

ckerr:
## Description

fix: Don't unconditionally convert quark strings to utf8.

Found another bug while testing the `transmission-show --scrape` regression discussed earlier today. This is an unreleased regression from 6c92d06 that unconditionally converted strings to utf8 in `tr_quark_new()`. This breaks the BitTorrent v2 dicts, which use arbitrary binary data as keys in the merkle trees. Every torrent that round-tripped through `tr_quark` lost those keys to U+FFFD and were corrupted.
    
This commit moves utf8-cleaning out of the quark code and into the consumers that require utf8-clean strings: JSON serialization and labels.

*edit:* marking as no-notes since bug never made it into production

## Checklist

### DIFF
--- a/libtransmission/variant-json.cc
+++ b/libtransmission/variant-json.cc
@@ -33,6 +33,7 @@
 
 #include "libtransmission/error.h"
 #include "libtransmission/quark.h"
+#include "libtransmission/string-utils.h"
 #include "libtransmission/tr-assert.h"
 #include "libtransmission/utils.h"
 #include "libtransmission/variant.h"
@@ -295,6 +296,14 @@ template<typename WriterT>
 struct JsonWriter {
     WriterT& writer;
 
+    using StringWriteFunc = bool (WriterT::*)(typename WriterT::Ch const*, rapidjson::SizeType, bool);
+
+    void WriteString(StringWriteFunc const func, std::string_view sv) const
+    {
+        auto const utf8 = tr_strv_to_utf8_string(sv);
+        (writer.*func)(std::data(utf8), std::size(utf8), false);
+    }
+
     void operator()(std::monostate /*unused*/) const
     {
     }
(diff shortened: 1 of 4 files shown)


libtransmission/variant-json.cc
```diff
@@ -296,6 +297,22 @@ template<typename WriterT>
...
     WriterT& writer;
 
+    using StringWriteFunc = bool (WriterT::*)(typename WriterT::Ch const*, rapidjson::SizeType, bool);
+
+    void WriteString(StringWriteFunc const func, std::string_view sv) const
+    {
+        // JSON requires utf8 strings, so ensure this is clean
+        auto cleaned = std::string{};
+        if (tr_strv_find_invalid_utf8(sv) != std::string_view::npos) {
+            cleaned = tr_strv_replace_invalid(sv);
```
tearfur: Small behaviour change on Apple machines here, because `tr_strv_replace_invalid()` is used instead of `tr_strv_to_utf8_string()`.

What's the reason for this?

tearfur:
Ah, luckily I decided to revert from `char8_t` to `char`. Fitting the merkle tree roots in quarks just would not have made sense.

tearfur (commented):
How about we add a Boolean parameter to `tr_quark_new` that controls whether to sanitise the string instead?

The only place (I think) that would disable sanitisation would be in here:

https://github.com/retransmission/retransmission/blob/9f6e7ac071626a962a8b38930afb6283f97482b3/libtransmission/variant-benc.cc#L182

```c++
   key_ = tr_quark_new(sv, false);
```

`tr_quark_new()` has *a lot* of indirect call sites, mainly via `tr_interned_string` and its methods. This PR doesn't even cover a fraction of them, and I can't vouch that nothing got broken.

libtransmission/variant-json.cc
ckerr: Brainfart. I'll update this to use `tr_strv_to_utf8_string()`.

ckerr:
> `tr_quark_new()` has _a lot_ of indirect call sites, mainly via `tr_interned_string` and its methods. This PR doesn't even cover a fraction of them, and I can't vouch that nothing got broken.

I've wanted to kill `tr_interned_string` for awhile now. Lemme take a stab at it today and see how it works.

tearfur:
Just to be clear, what I meant was I can't vouch that nothing got broken with the current changes. If you took my suggested approach, then I can be confident nothing new got broken.

Anyhow, I'll wait and see what you do with `tr_interned_string` before making a judgement again.

ckerr:
> Anyhow, I'll wait and see what you do with `tr_interned_string` before making a judgement again.

https://github.com/retransmission/retransmission/pull/167

libtransmission/variant-json.cc
ckerr: Fixed in the latest revision

tearfur (approved):

PR #161, opened 2026-07-29
refactor: use `std::to_array` in converters code

tearfur:
Purely code cleanup. No change in behaviour.

### DIFF
--- a/libtransmission-app/converters.cc
+++ b/libtransmission-app/converters.cc
@@ -32,7 +32,15 @@ template<typename T>
 inline constexpr bool HasTmGmtoffV = requires(T t) { t.tm_gmtoff; };
 
 template<typename T, size_t N>
-using Lookup = std::array<std::pair<std::string_view, T>, N>;
+using LookupTable = std::array<std::pair<std::string_view, T>, N>;
+
+// NOLINTBEGIN(modernize-avoid-c-arrays)
+template<typename T, size_t N>
+consteval LookupTable<std::remove_cv_t<T>, N> to_lookup(std::pair<std::string_view, T> (&&a)[N])
+{
+    return std::to_array(std::move(a));
+}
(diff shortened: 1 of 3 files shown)


ckerr (commented):
+1 on using `std::to_arary()` in most cases but there are a handful here where the hardcoded numbers were being used to keep these arrays in sync with the enums at compile time: the three arrays in converters.cc and the three in the tests.

If we want to use to_array for these six, we should probably also add something like

```c++
static_assert(std::size(ShowKeys) == ShowModeCount);
```

libtransmission-app/converters.cc
```diff
@@ -32,7 +32,15 @@ template<typename T>
 inline constexpr bool HasTmGmtoffV = requires(T t) { t.tm_gmtoff; };
 
 template<typename T, size_t N>
-using Lookup = std::array<std::pair<std::string_view, T>, N>;
+using LookupTable = std::array<std::pair<std::string_view, T>, N>;
+
+// NOLINTBEGIN(modernize-avoid-c-arrays)
+template<typename T, size_t N>
+consteval LookupTable<std::remove_cv_t<T>, N> to_lookup(std::pair<std::string_view, T> (&&a)[N])
```
ckerr, line 40: needs `#include <type_traits>` at top of header

tearfur:
> If we want to use to_array for these six, we should probably also add something like
> 
> ```c++
> static_assert(std::size(ShowKeys) == ShowModeCount);
> ```

I think I will do this. The original code will catch extra elements, but won't catch missing elements at compile time.

864b4a1986fe6ba6965babcde15bcdd5fae9bd57

libtransmission-app/converters.cc
tearfur: 1878dfa06630cf402a2933c20445e1e89b148422

ckerr (approved):

PR #165, opened 2026-07-30
refactor: remove `<fmt/format.h>` from `values.h`

ckerr:
## Description

Refactor `libtransmission/values.h` to not `#include <fmt/...>`. This cuts down the `#include` fan-out for pretty much every source file we have, since `values.h` is included by `types.h`.

## Checklist

### DIFF
--- a/libtransmission/values.h
+++ b/libtransmission/values.h
@@ -7,13 +7,10 @@
 
 #include <array>
 #include <cstddef> // size_t
-#include <cmath> // for std::fabs(), std::floor()
 #include <compare>
 #include <cstdint> // for uint64_t
 #include <string>
 #include <string_view>
-
-#include <fmt/format.h>
 
 namespace tr::Values
 {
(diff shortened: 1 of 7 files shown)


libtransmission/values.h
```diff
@@ -69,21 +51,50 @@ struct Config {
...
+        template<typename... Names> // NOLINTNEXTLINE(google-explicit-constructor)
+        Units(Base base, Names... names)
+        {
+            set_base(base);
+
+            auto idx = size_t{ 0U };
+            (set_name(idx++, names), ...);
+        }
+
+        using UnitsBase::display_name;
```
tearfur: Currently unused, but we could leave it in.

libtransmission/values.h
```diff
@@ -69,21 +51,50 @@ struct Config {
...
+        std::array<uint64_t, 5> multipliers_ = {};
+        Base base_ = {}; // NOLINT(bugprone-invalid-enum-default-initialization): overwritten by set_base()
+    };
+
+    template<typename UnitsEnum>
+    struct Units : UnitsBase {
+        template<typename... Names> // NOLINTNEXTLINE(google-explicit-constructor)
+        Units(Base base, Names... names)
+        {
+            set_base(base);
```
tearfur, line 69: While you're at it, let's add one more safeguard.


```suggestion
            static_assert(sizeof...(Names) == display_names_.size());

            set_base(base);
```

libtransmission/values.h
```diff
@@ -69,21 +51,50 @@ struct Config {
...
-            *fmt::format_to_n(std::data(display_names_[idx]), std::size(display_names_[idx]) - 1, "{:s}", name).out = '\0';
+            return UnitsBase::display_name(static_cast<size_t>(multiplier));
         }
 
-        std::array<std::array<char, 32>, 5> display_names_ = {};
-        std::array<uint64_t, 5> multipliers_;
-        Base base_ = {}; // NOLINT(bugprone-invalid-enum-default-initialization): overwritten by constructor
+        [[nodiscard]] constexpr auto multiplier(UnitsEnum multiplier) const noexcept
+        {
+            return multipliers_[static_cast<int>(multiplier)];
```
tearfur: *(nit)*

```suggestion
            return multipliers_[static_cast<size_t>(multiplier)];
```

libtransmission/values.h
ckerr: This was used in `rpcimpl.cc` but the use had a code smell of looping through until it got an empty string. I replaced it with a new method `display_names()` in e2ef438111e7fbfb565fc9c931dee97782f24db4

libtransmission/values.h
ckerr: 71e1706a227f2134c630188f764e18dc116e5f0a

libtransmission/values.h
ckerr: 71e1706a227f2134c630188f764e18dc116e5f0a

tearfur (approved):

PR #167, opened 2026-07-30
refactor: add `tr::shared_string` class

ckerr:
## Description

Replace `tr_interned_string` with `tr::shared_string`, which is similar to `tr_interned_string` except it's not tied to `tr_quark`: instead, it frees the string after the last `shared_string` that references it is destroyed.

## Checklist


libtransmission/peer-mgr.cc
```diff
@@ -33,7 +33,7 @@
 #include "libtransmission/clients.h"
 #include "libtransmission/crypto-utils.h"
 #include "libtransmission/handshake.h"
-#include "libtransmission/interned-string.h"
+#include "libtransmission/shared-string.h"
```
tearfur: *(nit)* not sorted

libtransmission/peer-msgs.cc
```diff
@@ -32,7 +32,7 @@
 #include "libtransmission/clients.h"
 #include "libtransmission/crypto-utils.h"
 #include "libtransmission/inout.h"
-#include "libtransmission/interned-string.h"
+#include "libtransmission/shared-string.h"
 #include "libtransmission/log.h"
```
tearfur, line 35: ditto

libtransmission/peer-msgs.h
```diff
@@ -14,7 +14,7 @@
 #include <cstddef> // for size_t
 #include <memory>
 
-#include "libtransmission/interned-string.h"
+#include "libtransmission/shared-string.h"
```
tearfur: ditto

libtransmission/shared-string.cc
```diff
@@ -0,0 +1,234 @@
...
+
+/** One pooled string, and the references to it that keep it alive. */
+struct string_pool_node {
+    explicit string_pool_node(std::string_view const key)
+        : value{ key }
+    {
+    }
+
+    std::atomic<size_t> refcount = 1U;
+    std::string value;
```
tearfur: *(perf)*

Every `sv()` call is reading 2 heap memory regions.
1. `tr::shared_string::node_`
2. Heap allocated string in `tr::detail::string_pool_node::value`.

We probably should store the string as part of the node? For example by using `small::string`.

If we're using C, I would have liked to use a flexible array member `char value[];` here, but we're on C++.

gtk/DetailsDialog.cc
```diff
@@ -1703,7 +1703,7 @@ void DetailsDialog::Impl::refreshTracker(std::vector<tr_torrent*> const& torrent
 
         // build the key to find the row
         gstr.str({});
-        gstr << torrent_id << '\t' << tracker.tier << '\t' << tracker.announce;
+        gstr << torrent_id << '\t' << tracker.tier << '\t' << tracker.announce.sv();
```
tearfur, line 1704: A big part of the diff was adding `.sv()` to expressions.

Was it a conscious choice not to add `operator std::sting_view` to `tr::shared_string`?

ckerr: Yes. This was to help me find all the places where we were implicitly converting the string so I could make sure we weren't holding any long-term references anywhere (which we were). This was necessary since these used to have a "permanently interned" guarantee which is going away in this PR

libtransmission/shared-string.cc
ckerr: Added the C++ version of flexible-array-member in e77799845afb8aabac7ca499cce1c13a0c8fa031

libtransmission/peer-msgs.h
ckerr: fixed eedd7282457c477834da56f8732ea0b2bfab982d

libtransmission/peer-msgs.cc
ckerr: fixed eedd7282457c477834da56f8732ea0b2bfab982d

libtransmission/peer-mgr.cc
ckerr: fixed eedd7282457c477834da56f8732ea0b2bfab982d

libtransmission/shared-string.cc
tearfur: Ooh nifty.

libtransmission/shared-string.cc
```diff
@@ -0,0 +1,272 @@
...
+ *
+ * The characters sit right after this header in the same heap block —
+ * create() over-allocates to make room — so a pooled string costs one
+ * allocation, not a node plus a separate string buffer, and readers
+ * dereference a single heap region.
+ */
+struct string_pool_node {
+    [[nodiscard]] static string_pool_node* create(std::string_view const key)
+    {
+        return new (::operator new(sizeof(string_pool_node) + std::size(key) + 1U)) string_pool_node{ key };
```
tearfur, line 36: From https://en.cppreference.com/cpp/memory/new/align_val_t:

> Both new-expression and delete-expression, when used with objects whose alignment requirement is greater than `__STDCPP_DEFAULT_NEW_ALIGNMENT__`, pass that alignment requirement as an argument of type `std::align_val_t` to the selected allocation/deallocation function.

```suggestion
        static_assert(alignof(string_pool_node) <= __STDCPP_DEFAULT_NEW_ALIGNMENT__);
        return new (::operator new(sizeof(string_pool_node) + std::size(key) + 1U)) string_pool_node{ key };
```

gtk/Torrent.cc
```diff
@@ -60,7 +60,9 @@ unsigned int build_torrent_trackers_hash(tr_torrent const& torrent)
     auto hash = uint64_t{};
 
     for (auto i = size_t{}, n = tr_torrentTrackerCount(&torrent); i < n; ++i) {
-        for (auto const ch : std::string_view{ tr_torrentTracker(&torrent, i).announce }) {
+        // the view has to outlive the loop below: it owns the text being hashed
```
tearfur: Now is a good time to replace the bespoke code here with `std::hash`. We don't even need this helper function anymore, we can invoke `std::hash` directly at the call site.

Maybe it'll be a good idea to add a `std::hash<tr::shared_string>` specialisation, which will wrap around `std::hash<std::string_view>`. If we have that, we can use `tr::shared_string` as `std::unordered_map/set` keys too.

tearfur (changes requested):
Merge conflict + some more minor comments.

libtransmission/shared-string.cc
ckerr: fixed in de53b10c935982eb75654f0f45a180df97fd9fef

gtk/Torrent.cc
ckerr: fixed in 0b3f92eca8efc8b02c176bc9b293ce849f338c0f and recycled `tr_hash_combine()` from `net.h` for the job

tearfur (approved):

tearfur:
Well that's awkward, somehow my email was listed as the main author in the commit 71d917ed90c618ee15c39eabfaeb7356ffaa3033.

Sloppy job from GitHub.

PR #175, opened 2026-08-04
perf: `tr_urlPercentEncode()` appends to a `std::string`

ckerr:
Avoid using tr_strbuf since it's a false savings here: every caller's URL turns into a std::string via either (1) tr_magnet_metainfo::magnet() or (2) tr_web::fetch().

## Description

Avoid using `tr_strbuf` in `tr_urlPercentEncode()` since it's a false savings here: every caller's URL turns into a `std::string` via either (1) `tr_magnet_metainfo::magnet()` or (2) `tr_web::fetch()`.

## Checklist

### DIFF
--- a/libtransmission/announcer-http.cc
+++ b/libtransmission/announcer-http.cc
@@ -34,7 +34,6 @@
 #include "libtransmission/session.h"
 #include "libtransmission/string-utils.h"
 #include "libtransmission/tr-assert.h"
-#include "libtransmission/tr-strbuf.h" // tr_strbuf, tr_urlbuf
 #include "libtransmission/types.h"
 #include "libtransmission/utils.h"
 #include "libtransmission/web-utils.h"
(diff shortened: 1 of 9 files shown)


libtransmission/magnet-metainfo.cc
```diff
@@ -152,24 +150,25 @@ std::optional<tr_sha256_digest_t> parseHash2(std::string_view sv)
 
 std::string tr_magnet_metainfo::magnet() const
 {
-    auto buf = tr_urlbuf{ "magnet:?xt=urn:btih:"sv, info_hash_string() };
+    auto buf = std::string{ "magnet:?xt=urn:btih:"sv };
```
tearfur:
```suggestion
    auto buf = "magnet:?xt=urn:btih:"s;
```

libtransmission/announcer-http.cc
```diff
@@ -480,10 +490,9 @@ void tr_tracker_http_scrape(tr_session const* session, tr_scrape_request const&
         response.rows[i].info_hash = request.info_hash[i];
     }
 
-    auto scrape_url = tr_pathbuf{};
-    scrape_url_new(scrape_url, request);
+    auto scrape_url = scrape_url_new(request);
     tr_logAddTrace(fmt::format("Sending scrape to libcurl: '{}'", scrape_url), request.log_name);
-    auto options = tr_web::FetchOptions{ scrape_url, onScrapeDone, d };
+    auto options = tr_web::FetchOptions{ std::move(scrape_url), onScrapeDone, d };
```
tearfur: `scrape_url` can be inlined.

```suggestion
    auto options = tr_web::FetchOptions{ scrape_url_new(request), onScrapeDone, d };
```

tearfur (approved):
The changes look correct.

If you feel like it, here are some minor code smells to fix, but they are not blockers by any means.

libtransmission/announcer-http.cc
tearfur: Oops now I see this was a dumb suggestion. I will revert this change, then merge.

PR #186, opened 2026-08-07
refactor: drop `tr_ctorMode`

ckerr:
## Description

This mode is both confusing and redundant. `tr_torrent::init()` already loads the defaults so we don't need `TR_FALLBACK` mode. Removing that only leaves one mode; so the enum can go away, too.

No behavior changes.

## Checklist

### DIFF
--- a/gtk/Session.cc
+++ b/gtk/Session.cc
@@ -695,20 +695,20 @@ namespace
 
 void core_apply_defaults(tr_ctor* ctor)
 {
-    if (!tr_ctorGetPaused(ctor, TR_FORCE, nullptr)) {
-        tr_ctorSetPaused(ctor, TR_FORCE, !gtr_pref_flag_get(TR_KEY_start_added_torrents));
+    if (!tr_ctorGetPaused(ctor, nullptr)) {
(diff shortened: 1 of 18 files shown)


gtk/Session.cc
```diff
@@ -695,20 +695,20 @@ namespace
 
 void core_apply_defaults(tr_ctor* ctor)
```
tearfur, line 696: Looks like the whole helper is redundant code.

- `start_added_torrents`: Covered in `tr_torrent::init()`.
- `trash_original_torrent_files`: Covered in `tr_ctor` constructor.
- `peer_limit_per_torrent`: Covered in `tr_torrent::init()`.
- `download_dir`: Covered in `tr_torrent::init()`.

tearfur (approved):
Having said that, PR is good as-is.

The redundant code in GTK is not new, we can leave it for another PR.

tearfur:
@ckerr I'll let you merge this one because of https://github.com/github/gh-stack/issues/371.

The commit author will be wrong if I merge.

gtk/Session.cc
ckerr: Some of this is handled in https://github.com/retransmission/retransmission/pull/190. 

I think others could be addressed in a followup

PR #199, opened 2026-08-11
refactor: extract group cell from XIB

lolgear:
To increase flexibility when editing group cell.

Extracting the group cell from the XIB unlocks:
2. Moving away from heavy, fragile XML-based XIB files, making future code reviews and layout modifications easier.
3. Preventing unexpected memory issues and layouts bugs caused by repetitive `awakeFromNib` calls.

### Results

This PR fixes a critical side-effect where `awakeFromNib` is invoked multiple times in `TorrentTableView` according to the [Apple documentation](https://apple.com).

Previously, `makeView(withIdentifier:owner:)` was called with `owner: self` (the table view itself). This caused `[TorrentTableView awakeFromNib]` to trigger repeatedly every time a new cell was instantiated (e.g., when the pool of cells was insufficient or when toggling between Compact and Small view settings). 

Removing the cell from the XIB completely solves these unexpected multiple invocations.

### Before

Снимок экрана 2026-02-17 в 21 45 06
Снимок экрана 2026-02-17 в 21 45 23

### After
Снимок экрана 2026-02-17 в 21 46 34
Снимок экрана 2026-02-17 в 21 46 29

### DIFF
--- a/macosx/GroupCell.mm
+++ b/macosx/GroupCell.mm
@@ -18,33 +18,31 @@
 
 - (void)configureViews
 {
-    __auto_type indicatorView = [[NSImageView alloc] init];
+    auto indicatorView = [[NSImageView alloc] init];
     indicatorView.imageScaling = NSImageScaleProportionallyDown;
     self.fGroupIndicatorView = indicatorView;
 
-    __auto_type titleField = [[NSTextField alloc] init];
+    auto titleField = [[NSTextField alloc] init];
     titleField.textColor = NSColor.secondaryLabelColor;
-    titleField.backgroundColor = NSColor.controlColor;
     titleField.lineBreakMode = NSLineBreakByTruncatingMiddle;
     self.fGroupTitleField = titleField;
+    titleField.allowsExpansionToolTips = YES;
 
-    __auto_type downloadView = [[NSImageView alloc] init];
+    auto downloadView = [[NSImageView alloc] init];
     downloadView.imageScaling = NSImageScaleProportionallyDown;
     self.fGroupDownloadView = downloadView;
 
-    __auto_type downloadField = [[NSTextField alloc] init];
+    auto downloadField = [[NSTextField alloc] init];
     downloadField.textColor = NSColor.secondaryLabelColor;
-    downloadField.backgroundColor = NSColor.textBackgroundColor;
     downloadField.lineBreakMode = NSLineBreakByClipping;
     self.fGroupDownloadField = downloadField;
 
-    __auto_type uploadAndRatioView = [[NSImageView alloc] init];
+    auto uploadAndRatioView = [[NSImageView alloc] init];
     uploadAndRatioView.imageScaling = NSImageScaleProportionallyDown;
     self.fGroupUploadAndRatioView = uploadAndRatioView;
 
-    __auto_type uploadAndRatioField = [[NSTextField alloc] init];
+    auto uploadAndRatioField = [[NSTextField alloc] init];
     uploadAndRatioField.textColor = NSColor.secondaryLabelColor;
-    uploadAndRatioField.backgroundColor = NSColor.textBackgroundColor;
     uploadAndRatioField.lineBreakMode = NSLineBreakByClipping;
     self.fGroupUploadAndRatioField = uploadAndRatioField;
 
@@ -52,7 +50,7 @@
         view.editable = NO;
         view.selectable = NO;
         view.bordered = NO;
-        view.font = [NSFont systemFontOfSize:11.0 weight:NSFontWeightBold];
+        view.font = [NSFont boldSystemFontOfSize:NSFont.smallSystemFontSize];
         view.drawsBackground = NO;
     }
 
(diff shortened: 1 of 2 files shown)


lolgear:
@ckerr 
Does `clang-format` have different style? ( `clang-format-20` )
It's a bit embarrassing.

lolgear:
@ckerr 
Also not sure how to transfer PRs between repositories.
I forked this repository and then I recreated branch with commit.
I mean "IF" these repositories may differ in some point of time.
And some things need to be synced between them.

ckerr:
> @ckerr Does `clang-format` have different style? ( `clang-format-20` ) It's a bit embarrassing.

This repo changed clang-format from 20 to 22 in #28 and switch brace style [to Linux from Allman](https://clang.llvm.org/docs/ClangFormatStyleOptions.html).

Just make sure you have clang-format-22 installed and run the `code_style.sh` script & it should just work

> Also not sure how to transfer PRs between repositories.
> I forked this repository and then I recreated branch with commit.

That sounds like a reasonable approach.

ckerr:
@lolgear I hope you're OK with AI-assisted code review? :sweat_smile: I am still getting up-to-speed with the macOS client. And TBH I will probably keep using AI as a second reviewer even with code I _am_ up-to-speed with :man_shrugging: 

1. The XIB prototype is still there, so the title's claim isn't what the code does. MainMenu.xib:51 still carries the full GroupCell prototype with all six outlet connections. The PR sidesteps it by renaming the lookup key to "NewGroupCell" rather than deleting it. That leaves two sources of truth — someone editing the XIB cell will see no effect — and keeps the XML the PR set out to remove. Delete the tableCellView identifier="GroupCell" node and use @"GroupCell" as the identifier.

    Relatedly, "NewGroupCell" is a name that only makes sense if you know the version it replaced. Once the XIB node is gone, @"GroupCell" is free — ideally as a shared static NSUserInterfaceItemIdentifier const so the two call sites can't drift.

2. The awakeFromNib fix is partial, and the actual bug is untouched. SmallTorrentCell (TorrentTableView.mm:234) and TorrentCell (TorrentTableView.mm:259) are still XIB prototypes fetched with owner:self, so -[TorrentTableView awakeFromNib] still fires repeatedly — far more often than the group path did, since torrent cells vastly outnumber group cells. What that method does is addObserver: for RefreshTorrentTable with no matching removal (TorrentTableView.mm:111), so observers accumulate and each refresh runs refreshTorrentTable N times. Moving that registration to initWithCoder: is a one-line fix that actually closes the bug today, independent of cell extraction. Please soften the PR description's "completely solves" — and the https://apple.com link is a placeholder, not documentation.

3. Layout fidelity vs. the XIB
    - Missing 8pt gap. XIB: uploadAndRatioView.leading == downloadField.trailing + 8. The PR uses constant 0, so the up-arrow/ratio icon butts against the download-speed text. This is the one geometric difference — every other constant (11, 5, 14, 16, ≥60, −5) matches.
    - Title field lost horizontalCompressionResistancePriority = 250. It now defaults to 750, tied with the two speed fields. The speed fields are pinned by required ≥60, so for short strings the tie is moot — but with a long speed string in a narrow window, nothing tells Auto Layout that the truncating-middle group name is the one meant to give way, and it may clip a speed field instead. Set it explicitly.
    - allowsExpansionToolTips = YES on the title is gone — truncated group names no longer expand on hover.
(imageScaling on the indicator changed from proportionallyUpOrDown to proportionallyDown, but discIconWithColor: produces 16×16 into a 14×14 frame, so it only ever scales down. No effect.)

4. Style and smaller items
    - if (self = [super initWithFrame:frameRect]) — both the local precedent (BaseFileNameCellView.mm:29) and TorrentTableView.mm:82 use the double-paren form that suppresses -Wparentheses. Not a build break (the Xcode project sets -w), but it's the house form.
    - translatesAutoresizingMaskIntoConstraints = false; — false in Objective-C, three lines below view.editable = NO;.
    - setContentHuggingPriority:251 — NSLayoutPriorityDefaultLow + 1 says what 251 means.
    - [NSFont systemFontOfSize:11.0 weight:NSFontWeightBold] hard-codes what the XIB expressed as smallSystemBold; [NSFont boldSystemFontOfSize:NSFont.smallSystemFontSize] tracks the system metric.
    - The three backgroundColor assignments are dead — every field also gets drawsBackground = NO. They also carry over an unexplained XIB inconsistency (title gets controlColor, the others textBackgroundColor). Just drop them.
    - Magic constants inline; BaseFileNameCellView.mm:10 puts them in named static CGFloat const k… at file scope.
    - GroupCell.h still declares all six properties IBOutlet. After the XIB node goes, the annotations are stale (the properties themselves must stay public — TorrentTableView sets them).

lolgear:
@ckerr 
A lot of words... oh...
But still manageable!
But some these are not applicable to this PR cause it is a part of series of "extractions".

lolgear:
> 1. The XIB prototype is still there, so the title's claim isn't what the code does. MainMenu.xib:51 still carries the full GroupCell prototype with all six outlet connections. The PR sidesteps it by renaming the lookup key to "NewGroupCell" rather than deleting it. That leaves two sources of truth — someone editing the XIB cell will see no effect — and keeps the XML the PR set out to remove. Delete the tableCellView identifier="GroupCell" node and use @"GroupCell" as the identifier.
>    Relatedly, "NewGroupCell" is a name that only makes sense if you know the version it replaced. Once the XIB node is gone, @"GroupCell" is free — ideally as a shared static NSUserInterfaceItemIdentifier const so the two call sites can't drift.
> 2. The awakeFromNib fix is partial, and the actual bug is untouched. SmallTorrentCell (TorrentTableView.mm:234) and TorrentCell (TorrentTableView.mm:259) are still XIB prototypes fetched with owner:self, so -[TorrentTableView awakeFromNib] still fires repeatedly — far more often than the group path did, since torrent cells vastly outnumber group cells. What that method does is addObserver: for RefreshTorrentTable with no matching removal (TorrentTableView.mm:111), so observers accumulate and each refresh runs refreshTorrentTable N times. Moving that registration to initWithCoder: is a one-line fix that actually closes the bug today, independent of cell extraction. Please soften the PR description's "completely solves" — and the https://apple.com link is a placeholder, not documentation.

Yes, first two are good to notice but these are not related to current stage.
It is better to cleanup xibs _after_ transition ( which can be switched in code right now by using different identifiers ).

> 3. Layout fidelity vs. the XIB
>    
>    * Missing 8pt gap. XIB: uploadAndRatioView.leading == downloadField.trailing + 8. The PR uses constant 0, so the up-arrow/ratio icon butts against the download-speed text. This is the one geometric difference — every other constant (11, 5, 14, 16, ≥60, −5) matches.

Ok, but still not sure if it is better.

>    * Title field lost horizontalCompressionResistancePriority = 250. It now defaults to 750, tied with the two speed fields. The speed fields are pinned by required ≥60, so for short strings the tie is moot — but with a long speed string in a narrow window, nothing tells Auto Layout that the truncating-middle group name is the one meant to give way, and it may clip a speed field instead. Set it explicitly.

Good point! Fixing it.

>    * allowsExpansionToolTips = YES on the title is gone — truncated group names no longer expand on hover.

Good point! Fixing it.

>      (imageScaling on the indicator changed from proportionallyUpOrDown to proportionallyDown, but discIconWithColor: produces 16×16 into a 14×14 frame, so it only ever scales down. No effect.)

Nothing to do.

> 4. Style and smaller items
>    
>    * if (self = [super initWithFrame:frameRect]) — both the local precedent (BaseFileNameCellView.mm:29) and TorrentTableView.mm:82 use the double-paren form that suppresses -Wparentheses. Not a build break (the Xcode project sets -w), but it's the house form.

Ok, discussable. Will keep it for now in old format.

>    * translatesAutoresizingMaskIntoConstraints = false; — false in Objective-C, three lines below view.editable = NO;.

Ok, changing.

>    * setContentHuggingPriority:251 — NSLayoutPriorityDefaultLow + 1 says what 251 means.

Ok, changing.

>    * [NSFont systemFontOfSize:11.0 weight:NSFontWeightBold] hard-codes what the XIB expressed as smallSystemBold; [NSFont boldSystemFontOfSize:NSFont.smallSystemFontSize] tracks the system metric.

Ok, changing.

>    * The three backgroundColor assignments are dead — every field also gets drawsBackground = NO. They also carry over an unexplained XIB inconsistency (title gets controlColor, the others textBackgroundColor). Just drop them.

Ok changing.

>    * Magic constants inline; BaseFileNameCellView.mm:10 puts them in named static CGFloat const k… at file scope.

Not related to this PR.

>    * GroupCell.h still declares all six properties IBOutlet. After the XIB node goes, the annotations are stale (the properties themselves must stay public — TorrentTableView sets them).

Good for this stage. After full transition they will be removed.

lolgear:
@ckerr 

Can AI post review comments to specific lines of code?
It would be convenient.

ckerr:
> @ckerr
> 
> Can AI post review comments to specific lines of code? It would be convenient.

I'm happy to make this be easier for both of us, but I'm not sure I that I understand what you're asking. Can you rephrase or give an example of "do A instead of B"?

ckerr:
> Yes, first two are good to notice but these are not related to current stage. It is better to cleanup xibs after transition (which can be switched in code right now by using different identifiers ).

OK.

Would you be alright with adding a `// TODO(lolgear): ...` in the code to make sure the task doesn't get lost later?

lolgear:
> > @ckerr
> 
> > 
> 
> > Can AI post review comments to specific lines of code? It would be convenient.
> 
> 
> 
> I'm happy to make this be easier for both of us, but I'm not sure I that I understand what you're asking. Can you rephrase or give an example of "do A instead of B"?

I mean

"Configure it to post inline / diff comments on specific lines of code, rather than just generating a high-level summary of the PR."

This feature called "A.I. powered inline comments".
I guess so..

ckerr:
Looks like this latest push matches the original's layout perfectly :+1: 

The only two items I see left for this PR:

1.  Add the `// TODO` comment we discussed
2. `translatesAutoresizingMaskIntoConstraints` is still `false` instead of `NO`

ckerr (approved):

ckerr:
> Label new pull requests / remove-on-approval (pull_request_review) Failing after 7s

:laughing:

Issue #200, opened 2026-08-11
feat: nonblocking disk IO


ckerr:
continuation of upstream issue 2462.

AI disclosure: I created this plan working interactively with Fable 5. In particular, it helped with number crunching and with analyzing libtorrent's source to update the previous plan to be more focused on seeding throughput to libtorrent peers. Fable also wrote the quick summary and agent-focused full plan below.

# Quick Summary

Move torrent disk IO off the session thread, prioritizing seeding throughput and concurrency safety. 

`tr::LocalData` (currently exists but is not used yet) becomes the gateway for all local data IO.
Its ordering contract is deliberately weaker than FIFO: data ops are unordered and may complete out of order; callbacks fire exactly once, This weaker-than-FIFO contract buys us optimizations focused on seeding, eg coalescing sequential requests. Callers guarantee ordering safety: we only write pieces we don't have; we only read pieces we have; rehashes are done by the last write completion. Workers operate on limited data: workers never see a `tr_torrent` or `tr_session` and are provided fds wrapped in refcounted guards instead of raw fds. `has_block` is set only when `pwrite()` completes, so resume files never claim unwritten data. Backpressure: outstanding requests + queued writes ≤ ~100 MiB, which is enforced where requests are issued.

Admin ops (move, rename, remove) run on workers too; their torrent-state updates run on the session thread from the op's completion. Shutdown finishes writes and admin ops and drops reads. Moves are crash-safe, so a kill mid-copy loses nothing.

Phases, in landing order (seeding is still the priority; the substrate landed with the writes):

0. route the call sites through LocalData, completion-driven accounting, shuffle backend in CI so FIFO assumptions fail before any thread exists. Landed.
1. gate + descriptor + pinned fds + async writes + budget; hash from still-buffered data (no read-back); open/prealloc move to workers. In review.
2. interrupt-safe move: copy to a temp name, rename into place, retry on startup.
3. async reads: RWF_NOWAIT inline for cache hits; cold misses coalesced to ≤ 1 MiB runs, elevator-sorted, per-peer in-flight cap.
4. admin ops (move, rename, remove) run on workers.
5. shutdown: drop reads, finish writes and admin ops, show what is holding quit.
6. verify folds into the scheduler (fixing its existing unsynchronized torrent-state writes); the bespoke verify thread retires.
7. (optional) — io_uring, copy-chain collapse, bigger fd pool.

# Full Plan (for agents)

[...]

PR #202, opened 2026-08-11
perf: route torrent disk IO through tr::LocalData

ckerr:
## Description

Implement P0 of #200.

Block reads and writes and completed-piece hashes now go through the `LocalData` facade instead of calling `tr_ioRead()` and `tr_ioWrite()` directly. The facade doesn't run anything asynchronously yet, so this phase is all setup and doesn't change behavior on its own.

Each of those operations now finishes in a completion handler. A caller can no longer assume the op is done when the call returns. Three of them used to assume synchronous writes:

  - has_block(B) now means B's write returned. Requests and duplicate rejection still happen when the block arrives from the peer, but completion_.add_block() and the piece hash wait for the write. A pending-write set covers the gap, so a block being written is neither re-requested nor written twice.

  - tr_ioWrite() no longer sets the torrent error and stops the torrent. It reports the failure and the session thread decides.

  - an upload read looks its request up by value when it completes. A request that was cancelled or rejected in the meantime drops its data, instead of answering whichever request now sits at the front of the queue.

The LT.ShuffledDiskIo test runs the disk-IO suites a second time, with completions parked and released out of order. tr::LocalData's ordering contract allows that, so whatever breaks under it is a caller assuming an ordering nobody promised. That is why this rework lands before any worker thread does. Only the test fixture turns the mode on, so it never reaches a production session.

Related fixes:

  - the 16 KiB per-block buffers were zero-filled and then immediately overwritten. LocalData::BlockData leaves them alone.

  - the full-preallocation fallback wrote 4 KiB at a time. It runs on FAT and exFAT, which are usually slow external drives, so it now writes 1 MiB at a time.

  - a full receive buffer left the read event enabled, so libevent spun on a socket peer-io was not going to read from.

P0 itself doesn't have any user-visible changes, but a couple of related fixes do:

Notes: Torrent disk reads and writes now run in the background, so a slow or busy drive no longer stalls the UI or other transfers, and seeding from disk is faster.

Notes:  Faster full file preallocation on FAT or exFAT drives.

## Checklist

### DIFF
--- a/libtransmission/peer-msgs.cc
+++ b/libtransmission/peer-msgs.cc
@@ -32,7 +32,7 @@
 #include "libtransmission/block-info.h"
 #include "libtransmission/clients.h"
 #include "libtransmission/crypto-utils.h"
-#include "libtransmission/inout.h"
+#include "libtransmission/local-data.h"
 #include "libtransmission/log.h"
 #include "libtransmission/macros.h"
 #include "libtransmission/peer-common.h"
@@ -226,9 +226,10 @@ struct tr_incoming {
 
     struct incoming_piece_data {
         explicit incoming_piece_data(uint32_t block_size)
-            : block_size_{ block_size }
+            : buf{ std::make_unique<tr::LocalData::BlockData>() }
+            , block_size_{ block_size }
         {
-            buf.resize(block_size);
+            buf->resize(block_size);
         }
 
         [[nodiscard]] bool add_span(size_t begin, size_t end)
@@ -249,7 +250,7 @@ struct tr_incoming {
             return have_.count() >= block_size_;
         }
 
-        std::vector<uint8_t> buf;
+        std::unique_ptr<tr::LocalData::BlockData> buf;
 
     private:
         std::bitset<tr_block_info::BlockSize> have_;
@@ -357,7 +358,7 @@ public:
             return active_requests.count();
 
         case tr_direction::PeerToClient: // requests they sent
-            return std::size(peer_requested_);
+            return std::size(peer_requested_) + std::size(reading_);
 
         default:
             TR_ASSERT(0);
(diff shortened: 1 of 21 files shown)


ckerr:
Changing to a draft. Sol and Claude think this PR is ready to go, but it has some code smells I want to change first.

libtransmission/peer-msgs.cc
```diff
@@ -1789,32 +1815,35 @@ ReadResult tr_peerMsgsImpl::read_piece_data(MessageReader& payload)
...
+    if (!tor_.on_block_received(block)) {
+        return false; // another peer beat us to it
     }
 
     active_requests.unset(block);
     publish(tr_peer_event::GotBlock(tor_.block_info(), block));
 
-    return 0;
+    // A failed write can pause the torrent, which destroys `this`.
+    // Don't touch any member after this call.
```
tearfur: I don't think this is accurate. `this` is kept alive here:

https://github.com/retransmission/retransmission/blob/60a250ab76d24cc3ff2472e8ce7bd5c26883580a/libtransmission/peer-msgs.cc#L1960-L1962

libtransmission/torrent.h
```diff
@@ -1296,6 +1329,15 @@ struct tr_torrent {
     // it means that piece needs to be checked before its data is used.
     tr_bitfield checked_pieces_ = tr_bitfield{ 0 };
 
+    // blocks we've received and are writing to disk.
+    // A block leaves this set when its write finishes.
+    tr_bitfield blocks_pending_write_ = tr_bitfield{ 0 };
+
+    // which version of a piece each in-flight hash is checking.
+    // An entry lives only as long as its hash.
+    std::unordered_map<tr_piece_index_t, uint64_t> hash_tokens_;
```
tearfur: TBD use small?

PR #205, opened 2026-08-11
fix: regression in zero-byte files support in `.resume` files

### LINKED ISSUE #203: Transmission 4.0.x resume incompatibility can remap and corrupt torrents containing zero-length files
## Note

This is the Retransmission counterpart of [transmission/transmission#9000](https://github.com/transmission/transmission/issues/9000).

That report follows an extended investigation of real data-integrity incidents. Multiple pre-patch `.resume` files were decoded and compared with their immutable metainfo, affected files were checked by size, inode, and content, and the relevant resume, file-I/O, and verification paths were traced in Transmission 4.1.3.

Retransmission currently inherits the affected resume filename loading behavior.

## Summary

Transmission 4.0.x omitted zero-length files while parsing metainfo. [transmission/transmission#6232](https://github.com/transmission/transmission/pull/6232), merged as [commit `ed4fad9b`](https://github.com/transmission/transmission/commit/ed4fad9b1868bbdddfe58d4aaf55032ba3eac70a), added zero-length file support by removing this early return from `torrent-metainfo.cc`:

```cpp
if (file_length_ == 0)
{
    return ok;
}
```

This correctly made zero-length files visible in later versions, but it also changed the file indexes of every following file. No migration was added for the per-file arrays already stored in Transmission 4.0.x `.resume` files.

For a torrent with `N` metainfo files and `K` zero-length files, Transmission 4.0.x saved the following arrays with `N-K` entries:

* `files`
* `dnd`
* `priority`
* `progress.mtimes`

The same metainfo has all `N` file entries in Retransmission.

The DND and priority loaders reject an incompatible list length, while the mtime loader explicitly handles the size mismatch. `load_filenames()`, however, accepts the shorter legacy list and applies every available pathname positionally. It is therefore the only one of these per-file loaders that applies an incompatible list positionally instead of rejecting or explicitly normalizing its length.

All paths after an omitted zero-length file are consequently assigned to the wrong metainfo indexes.

This is a deterministic state-compatibility regression, not a randomly malformed resume file.

## Reproduction

1. Add this multi-file torrent to Transmission 4.0.6:

   ```text
   0: root/a
   1: root/zero  (length 0)
   2: root/b
   3: root/c
   ```

2. Let Transmission 4.0.6 save its resume data. Because it omits the zero-length metainfo entry, the stored filename list is:

   ```text
   root/a
   root/b
   root/c
   ```

3. Start an unpatched Retransmission client with the same profile.

4. `load_filenames()` applies those entries to indexes `0`, `1`, and `2`. Index `3` keeps its canonical path, so indexes `2` and `3` both resolve to `root/c`:

   ```text
   0: root/a
   1: root/b
   2: root/c
   3: root/c
   ```

5. If the torrent runs, different logical files can read or write through the shifted mapping.

6. If Retransmission saves the resume, the filename list becomes full-length but internally inconsistent:

   ```text
   root/a
   root/b
   root/c
   root/c
   ```

The original short form proves that the resume was written using the Transmission 4.0.x layout. After an unpatched newer client expands and saves the shifted state, that cross-array signature is lost and the incorrect paths may no longer be distinguishable from legitimate user renames.

## Field evidence

Four independent `.torrent` / pre-patch `.resume` pairs from a Transmission 4.0.x installation had exactly the layout predicted by the regression:

| Case | Metainfo files | Zero-length indexes | `files` / `dnd` / `priority` / `mtimes` lengths |
| ---- | -------------: | ------------------- | ----------------------------------------------: |
| A    |             20 | 2                   |                               19 / 19 / 19 / 19 |
| B    |          2,549 | 0                   |                   2,548 / 2,548 / 2,548 / 2,548 |
| C    |            433 | 251                 |                           432 / 432 / 432 / 432 |
| D    |          4,532 | 3,034               |                   4,531 / 4,531 / 4,531 / 4,531 |

In every case, the stored path sequence matched the metainfo sequence with the zero-length entry omitted. Case B confirms the boundary case where index `0` is omitted and every stored pathname is shifted.

For case A, both generations were captured: a backup contained the original 19-entry form, while the live resume had already been rewritten by an unpatched newer client into a full-length, internally inconsistent form.

All paths after the zero-length placeholder were shifted and the final path was duplicated. Two file indexes addressed the same final file. Two adjacent files on disk were also confirmed byte-for-byte identical even though the metainfo described different files.

## Consequences and seedbox impact

Retransmission’s open-file cache is keyed by `(torrent id, file index)`, not by canonical path. Different file indexes resolving to the same pathname may therefore obtain separate file descriptors for the same inode and write unrelated torrent data at their respective local offsets.

Possible consequences include:

* data written into the wrong files;
* duplicated or corrupted files;
* invalid uploads;
* large progress drops during verification;
* completed filenames receiving `.part` suffixes again;
* unnecessary re-downloads;
* substantial manual recovery work.

This regression was discovered on a seedbox containing roughly 14,000 torrents, where many affected torrents were found. The same failure pattern was also observed on other servers.

Every multi-file torrent containing a zero-length file whose resume was written by Transmission 4.0.x carries the incompatible per-file layout. In a large, long-running session, the incompatible mapping may go unnoticed. Loading and saving a torrent can already persist the shifted mapping, while data corruption or invalid uploads occur only when the torrent later performs file I/O. Verification may expose the damage only afterward.

Inspecting or verifying an entire large session manually is impractical.

Verification by itself does not fix the problem. It cannot reconstruct the shifted filename mapping or restore physical files and data already written under the wrong indexes. If verification runs before the mapping is repaired, it can cause large progress drops and unnecessary re-downloads.

[transmission/transmission#8959](https://github.com/transmission/transmission/issues/8959) independently reports the same failure pattern in a separate FreeBSD/ZFS/NFS deployment containing approximately 40,000 torrents: an upgrade from Transmission 4.0.5 to 4.1.2 with existing resume files, torrents containing zero-length files, and Transmission treating those files as if they had been replaced by other files from the same torrent. Removing and re-adding the affected torrents without deleting their payload data resolved the problem, which is consistent with discarding the incompatible legacy resume mapping.

Possibly related, [transmission/transmission#8791](https://github.com/transmission/transmission/issues/8791) describes closely matching symptoms after an upgrade from Transmission 4.0.6 to 4.1.1: changed file structure and sizes followed by a 100%-to-0% drop during verification.

## Expected behavior and broader resume safety

A recognizable Transmission 4.0.x per-file layout should be handled before any stored filenames are applied or any file I/O is allowed.

More generally, an incompatible resume filename list should never be applied partially. Filename loading should fail safely and transactionally when the stored structure does not match the current metainfo.

This would protect not only the known zero-length-file migration case, but also against future file-index layout changes and malformed, truncated, or internally inconsistent filename arrays. Such resume data should not be allowed to silently remap only part of a torrent’s files.

If the mapping cannot be reconstructed safely, the torrent should remain stopped and clearly require filename recovery and local-data verification instead of operating through an ambiguous mapping.

A fix has already been prepared and adapted for Retransmission and will be submitted separately as a pull request.

None

_No response_

ckerr:
## Description

If a resume list is short by exactly the number of the torrent's zero-length files, realign the list while walking. This is to preserve any renames from 4.0.x. New regression tests by Claude.

Fixes #203.

Notes: Fixed a Transmission 4.1.0 regression that could cause torrents with zero-byte files to be handled incorrectly, sometimes losing downloaded progress.

## Checklist

### DIFF
--- a/libtransmission/resume.cc
+++ b/libtransmission/resume.cc
@@ -3,7 +3,7 @@
 // or any future license endorsed by Mnemosaic LLC.
 // License text can be found in the licenses/ folder.
 
-#include <algorithm> // std::min
+#include <algorithm> // std::min, std::ranges::adjacent_find, std::ranges::sort
 #include <cstdint>
 #include <cstring>
 #include <ctime>
@@ -431,13 +431,46 @@ tr_resume::fields_t load_filenames(tr_variant::Map const& map, tr_torrent* tor)
         return {};
     }
 
+    // The pathname each file would end up with. The mapping is built in full
+    // before any of it is applied, since a duplicate can turn up at any
+    // position and the entries before it would already be on their files.
+    auto subpaths = std::vector<std::string_view>{};
+    subpaths.reserve(n_files);
     for (tr_file_index_t i = 0, pos = 0; i < n_files; ++i) {
-        if (!file_has_list_entry(tor, alignment, i)) {
-            continue;
+        // A file with no entry, or whose entry isn't a usable pathname,
+        // keeps the pathname its metainfo gave it.
+        auto subpath = std::string_view{ tor->file_subpath(i) };
+
+        if (file_has_list_entry(tor, alignment, i)) {
+            if (auto const sv = (*list)[pos++].value_if<std::string_view>(); sv && !std::empty(*sv)) {
+                subpath = *sv;
+            }
         }
 
-        if (auto const sv = (*list)[pos++].value_if<std::string_view>(); sv && !std::empty(*sv)) {
-            tor->set_file_subpath(i, *sv);
+        subpaths.push_back(subpath);
+    }
+
+    // Two files sharing a pathname share one file on disk, but the open-file
+    // cache keys on file index, so each gets its own descriptor and they
+    // write over each other. A torrent's own file list can't name a file
+    // twice, so a duplicate is the saved list's, and the rest of that list is
+    // no more trustworthy than the part that collided.
+    auto sorted = subpaths;
+    std::ranges::sort(sorted);
+    if (auto const dupe = std::ranges::adjacent_find(sorted); dupe != std::end(sorted)) {
+        tr_logAddWarnTor(
+            tor,
+            fmt::format(
+                "Not using the filenames saved for this torrent: they give '{:s}' to more than one file. "
+                "Using the filenames from the torrent instead; the saved ones are in '{:s}' until it is written over.",
+                *dupe,
+                tor->resume_file()));
+        return {};
+    }
+
+    for (tr_file_index_t i = 0; i < n_files; ++i) {
+        if (auto const subpath = subpaths[i]; subpath != tor->file_subpath(i)) {
+            tor->set_file_subpath(i, subpath);
         }
     }
 
--- a/libtransmission/resume.cc
+++ b/libtransmission/resume.cc
@@ -43,6 +43,54 @@ constexpr auto MaxRememberedPeers = 200U;
 
 // ---
 
+// How a saved per-file list's entries line up with a torrent's files.
+enum class FileListAlignment : uint8_t {
+    ByIndex, // one entry per file, in file order
+    SkipEmptyFiles, // one entry per nonempty file, in file order
+    Unusable, // no reading pairs the entries up with the files
+};
+
+[[nodiscard]] FileListAlignment file_list_alignment(tr_torrent const* const tor, size_t const n_list)
+{
+    auto const n_files = tor->file_count();
+
+    if (n_list == n_files) {
+        return FileListAlignment::ByIndex;
+    }
+
+    // Resume files written before zero-length files were part of a torrent's
+    // file list are short by exactly those files, so each entry after an
+    // omitted one sits at a lower position in the list than its file index.
+    // Skipping the zero-length files while walking the list realigns them.
+    auto n_empty_files = size_t{};
+    for (tr_file_index_t i = 0; i < n_files; ++i) {
+        if (tor->file_size(i) == 0U) {
+            ++n_empty_files;
+        }
+    }
+
+    if (n_list + n_empty_files == n_files) {
+        return FileListAlignment::SkipEmptyFiles;
+    }
+
+    // A list of any other length can't be paired up with the files, so none
+    // of it can be applied: a partial mapping would give file indices other
+    // files' settings.
+    return FileListAlignment::Unusable;
+}
+
+// Whether a list with this alignment has an entry for this file. Zero-length
+// files have none in a list that was written without them.
+[[nodiscard]] bool file_has_list_entry(
+    tr_torrent const* const tor,
+    FileListAlignment const alignment,
+    tr_file_index_t const file_index)
+{
+    return alignment != FileListAlignment::SkipEmptyFiles || tor->file_size(file_index) != 0U;
+}
+
+// ---
+
 void save_peers(tr_variant::Map& map, tr_torrent const* tor)
 {
     if (auto const pex = tr_peerMgrGetPeers(tor, TR_AF_INET, TR_PEERS_INTERESTING, MaxRememberedPeers); !std::empty(pex)) {
@@ -157,14 +205,10 @@ tr_resume::fields_t load_dnd(tr_variant::Map const& map, tr_torrent* tor)
     }
 
     auto const n = tor->file_count();
-    if (std::size(*list) != n) {
-        tr_logAddDebugTor(
-            tor,
-            fmt::format(
-                "Couldn't load DND flags. DND list {} has {} children; torrent has {} files",
-                fmt::ptr(list),
-                std::size(*list),
-                n));
+    auto const n_list = std::size(*list);
+    auto const alignment = file_list_alignment(tor, n_list);
+
+    if (alignment == FileListAlignment::Unusable) {
         return {};
     }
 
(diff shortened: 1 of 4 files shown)


h1de0x:
I agree that a recognized `N-K` layout does not need to be quarantined. As long as the affected per-file lists are still short, #205 can identify the old 4.0.x layout and realign its entries safely.

However, #203 also documents a state that #205 cannot recognize. In case A, we have both versions of the resume file: the backup has the original 19-entry list, while the live resume had already been rewritten by an unpatched 4.1.x client into a full-length shifted list. In simplified form:

```text
metainfo: a, zero, b, c
resume:   a, b,    c, c
```

Since the rewritten list already has `N` entries, `file_list_alignment()` treats it as `ByIndex`. `load_filenames()` then applies it as-is, leaving two file indices pointing to `c`.

This is unsafe in practice. The open-file cache is keyed by torrent ID and file index, so two indices resolving to the same path may get separate file descriptors for the same underlying file. Reads or writes through that mapping can corrupt payload data and cause invalid uploads.

To catch this case, the effective filename mapping needs to be validated before it changes the torrent. A duplicate may appear near the end of the list, after the current loop has already applied earlier entries. The patch therefore builds the complete mapping first, checks it for duplicate paths, and only then applies it.

If a duplicate is found, none of the resume filenames are applied and the torrent keeps the canonical paths from its metainfo. This may require manual recovery of renames stored only in the resume file, but it is safer than allowing two file indices to alias the same path. If no payload data can be found under the canonical paths, Retransmission's existing missing-files check keeps the torrent stopped and reports `No data found`.

Everything else in #205 remains unchanged: valid full-length and legacy-length lists behave as before, entries that are not usable paths keep their canonical path without shifting later entries, and no quarantine or persistent state is added.

This covers the observed full-length rewrite by an unpatched 4.1.x client without changing #205's migration or the torrent start/verify lifecycle.

```diff

From 15c28744de463454c180cbf8bfdac03c12872594 Mon Sep 17 00:00:00 2001
From: Kirill Shtumf <shtumff@gmail.com>
Date: Wed, 12 Aug 2026 13:26:00 +0300
Subject: [PATCH] fix: validate resume filename mappings transactionally

Build the complete aligned filename mapping before applying it and reject duplicate effective paths, while preserving the existing lenient handling of unusable entries.
---
 libtransmission/resume.cc            | 33 ++++++++++++++++++++++++----
 tests/libtransmission/resume-test.cc | 12 ++++++++++
 2 files changed, 41 insertions(+), 4 deletions(-)

diff --git a/libtransmission/resume.cc b/libtransmission/resume.cc
index 14f4182e6..5cfd63ab4 100644
--- a/libtransmission/resume.cc
+++ b/libtransmission/resume.cc
@@ -8,7 +8,9 @@
 #include <cstring>
 #include <ctime>
 #include <limits>
+#include <string>
 #include <string_view>
+#include <unordered_set>
 #include <vector>
 
 #include <fmt/format.h>
@@ -432,16 +434,39 @@ tr_resume::fields_t load_filenames(tr_variant::Map const& map, tr_torrent* tor)
         return {};
     }
 
+    // Build and validate the complete mapping before changing the torrent.
+    // Applying only the usable entries can leave two file indices pointing at
+    // the same path when a resume file is malformed or was already rewritten
+    // after being loaded with a shifted legacy mapping.
+    auto filenames = std::vector<std::string>{};
+    filenames.reserve(n_files);
+    // The vector is fully reserved before views into its strings are stored,
+    // so those views remain valid throughout validation.
+    auto unique_filenames = std::unordered_set<std::string_view>{};
+    unique_filenames.reserve(n_files);
+
     for (tr_file_index_t i = 0, pos = 0; i < n_files; ++i) {
-        if (!file_has_list_entry(tor, alignment, i)) {
-            continue;
+        auto filename = std::string_view{ tor->file_subpath(i) };
+
+        if (file_has_list_entry(tor, alignment, i)) {
+            if (auto const sv = (*list)[pos++].value_if<std::string_view>(); sv && !std::empty(*sv)) {
+                filename = *sv;
+            }
         }
 
-        if (auto const sv = (*list)[pos++].value_if<std::string_view>(); sv && !std::empty(*sv)) {
-            tor->set_file_subpath(i, *sv);
+        filenames.emplace_back(filename);
+        if (!unique_filenames.emplace(std::string_view{ filenames.back() }).second) {
+            tr_logAddWarnTor(
+                tor,
+                fmt::format("Couldn't load filenames from resume file: duplicate path '{:s}'", filenames.back()));
+            return {};
         }
     }
 
+    for (tr_file_index_t i = 0; i < n_files; ++i) {
+        tor->set_file_subpath(i, filenames[i]);
+    }
+
     return tr_resume::Filenames;
 }
 
diff --git a/tests/libtransmission/resume-test.cc b/tests/libtransmission/resume-test.cc
index d9c2fea88..53a8df3d9 100644
--- a/tests/libtransmission/resume-test.cc
+++ b/tests/libtransmission/resume-test.cc
@@ -276,6 +276,18 @@ TEST_F(ResumeTest, savedFilenamesListLongerThanFileCount)
     expectSubpaths(torrentInit(ctor, file_sizes, savedFilenames(saved)), canonicalSubpaths(std::size(file_sizes)));
 }
 
+// An unpatched 4.1.x client rewrites a shifted legacy list at full length. The
+// duplicated tail path makes that mapping unsafe even though its length looks
+// ordinary, so none of it may be applied.
+TEST_F(ResumeTest, savedFilenamesFullLengthShiftedMappingIsRejected)
+{
+    auto const file_sizes = std::vector<uint64_t>{ 1U, 0U, 1U, 1U };
+    auto const saved = std::vector<std::string>{ "root/f0", "root/f2", "root/f3", "root/f3" };
+
+    auto ctor = tr_ctor{ session_ };
+    expectSubpaths(torrentInit(ctor, file_sizes, savedFilenames(saved)), canonicalSubpaths(std::size(file_sizes)));
+}
+
 // An entry that isn't a usable pathname leaves that one file alone. It still
 // holds a position, so the entries after it keep their alignment.
 TEST_F(ResumeTest, savedFilenamesUnusableEntry)
-- 
2.43.0

```

ckerr:
@h1de0x updated in 9440464fbc3ece6e4a16117a23e6d45015fcf796

h1de0x:
@ckerr Thanks for taking the time to work through this and turn the idea into a cleaner implementation. This looks great to me - nice work! :)

PR #206, opened 2026-08-11
fix: migrate pre-4.1 zero-length file resume layout safely

### LINKED ISSUE #203: Transmission 4.0.x resume incompatibility can remap and corrupt torrents containing zero-length files

h1de0x:
Fixes #203 

Upstream report: [transmission/transmission#9000](https://github.com/transmission/transmission/issues/9000)
Upstream implementation: [transmission/transmission#9002](https://github.com/transmission/transmission/pull/9002)

## Summary

Transmission 4.0.x omitted zero-length files from the metainfo file list. Support for zero-length files was added by [transmission/transmission#6232](https://github.com/transmission/transmission/pull/6232) / [commit `ed4fad9b`](https://github.com/transmission/transmission/commit/ed4fad9b1868bbdddfe58d4aaf55032ba3eac70a), so Retransmission now includes those files. However, no migration was added for the per-file arrays already stored in Transmission 4.0.x `.resume` files.

For a torrent with `N` metainfo files and `K` zero-length files, Transmission 4.0.x wrote `files`, `dnd`, `priority`, and `progress.mtimes` arrays with `N-K` entries. Retransmission sees all `N` files.

The existing filename loader accepts the shorter array and applies each available entry by index, shifting every stored path after an omitted zero-length file. Unlike the other affected per-file loaders, it applies an incompatible list positionally instead of rejecting or explicitly normalizing its length.

This is a data-integrity regression. Logical file indexes can be redirected to the wrong physical paths, and two indexes can eventually refer to the same path. Subsequent downloads, uploads, resume rewrites, and verification may then operate on the shifted mapping.

## What this PR does

This PR adds two layers of protection.

### Transactional filename validation

Resume filename loading now validates the complete mapping before applying any stored rename:

* the filename list must have exactly the current metainfo file count;
* every element must be a string;
* legacy empty entries continue to mean “use the canonical metainfo path”;
* effective paths must be unique by exact string comparison;
* no rename is applied until the entire list has passed validation.

A valid full-length list of stored renames continues to be accepted normally.

If an inconsistent mapping cannot be migrated safely, Retransmission rejects the complete stored filename list, falls back to canonical metainfo paths, and places the torrent in a persistent verification quarantine.

This prevents malformed, truncated, partially incompatible, or internally colliding filename arrays from silently remapping only part of a torrent’s files. It also provides a safety boundary for future changes that might make stored file indexes incompatible with the current metainfo layout.

### Deterministic migration of the 4.0.x layout

The primary resume layout written by Transmission 4.0.x is migrated only when all of the following are true:

* the torrent contains at least one zero-length file;
* `files`, `dnd`, `priority`, and `progress.mtimes` are all present;
* all four arrays have exactly `N-K` entries;
* inserting entries at the known zero-length file positions produces a complete, valid, unique filename mapping.

For a recognized layout, the migration inserts canonical/default entries for zero-length files and preserves the stored filename, selection, priority, and mtime state of every nonzero file.

The migrated torrent is still quarantined until its local data has been explicitly verified.

## Verification quarantine

The quarantine uses the existing torrent error and start paths, so no frontend-specific GTK, Qt, or Web UI changes are required.

It:

* blocks both automatic and manual starts;
* clears any queued state;
* persists across resume rewrites and process restarts;
* reports that invalid file paths were found in resume data and that local data must be verified;
* permits an explicit `Verify Local Data` operation;
* remains set if verification is aborted;
* is cleared only after successful verification;
* rewrites the corrected full-length resume state;
* leaves the torrent stopped after verification so that restarting remains an explicit user decision.

## Why validation alone is not sufficient

A minimal fix could reject every short filename array and use canonical metainfo paths. That prevents a new positional shift, but it discards legitimate renames stored by Transmission 4.0.x.

It also loses the correspondence needed to find existing data when files or directories were renamed and no longer match the metainfo layout.

That fallback is especially destructive when followed by verification. Verification checks bytes at the paths currently assigned to each file; it cannot infer the intended file-index-to-path mapping. If the mapping is wrong, correct data may be reported missing, progress can collapse, completed filenames can regain the `.part` suffix, and large amounts of data may be downloaded again.

The Transmission 4.0.x form is safely recognizable because all four per-file arrays were written from the same old file list. Requiring their complete, matching `N-K` signature allows the omitted positions to be restored without guessing.

Migration is therefore part of the safety fix, not merely a convenience. Without it, installing the fix on an untouched 4.0.x profile could itself turn a detectable compatibility problem into a destructive manual recovery procedure.

Already rewritten resumes are intentionally not reconstructed. Once an unpatched newer client has expanded and saved the shifted state, the original cross-array signature is gone, and stored renames cannot always be distinguished from corruption. Already rewritten resumes are intentionally not reconstructed. Once an unpatched newer client has expanded and saved the shifted state, the original cross-array signature is gone, and stored renames cannot always be distinguished from corruption. Full-length mappings that fail transactional validation are rejected and quarantined; otherwise, reconstructing the original layout would require guessing.

## Why verification remains mandatory

Restoring the mapping does not prove that the payload is intact. An unpatched client may already have downloaded or uploaded through the shifted mapping before Retransmission encounters the resume file.

For that reason, even a successfully migrated torrent is not started automatically.

Automatic verification would not solve the mapping problem: it validates the paths currently assigned to each file, so running it before the paths are reconstructed or manually restored can discard valid completion state.

The safe order is:

1. reconstruct the proven Transmission 4.0.x layout, or let the user restore an ambiguous mapping;
2. keep the torrent stopped;
3. verify the local data explicitly;
4. save the corrected full-length resume state;
5. let the user decide whether to start the torrent.

## Impact on large sessions

This regression was investigated on a long-running seedbox with approximately 14,000 torrents, where many affected torrents were found. The same failure pattern was observed on other servers.

Affected torrents may remain unnoticed until they next perform file I/O or are verified, so consequences can appear gradually after an upgrade.

The startup work added by this PR is metadata-only and linear in each torrent’s file count. It does not read or hash payload data and does not start automatic verification.

Valid resume files continue normally. Only torrents with a recognized legacy layout or an invalid filename mapping are stopped.

This distinction matters operationally: validation prevents further damage, while deterministic migration preserves paths and per-file state that might otherwise need to be reconstructed manually across a large session.

## Preventing invalid uploads

The Transmission 4.0.x `files` and `progress.mtimes` arrays are shifted together. The mtime stored at a logical file index may therefore match the wrong physical file selected by the shifted pathname, leaving the corresponding checked-piece state trusted.

A subsequent peer request can then read data through the shifted path without first calculating a new piece hash. The receiver rejects the data when its completed piece fails verification and may penalize or ban the sender after repeated corrupt blocks.

The quarantine therefore protects not only local payload data, but also other peers and the affected client’s reputation in the swarm.

## Tests

The new `ResumeTest` suite contains 10 tests covering:

* rejection of an incompatible filename count when the full migration signature is absent;
* migration with zero-length files at the beginning, middle, and end of the file list;
* preservation of nonzero-file names, selection state, priorities, and mtimes during migration;
* rejection when reconstructed paths collide;
* transactional rejection of a stored path colliding with a canonical metainfo path;
* acceptance of a complete valid list of stored renames;
* compatibility with legacy empty filename entries;
* transactional rejection when any filename element is not a string;
* quarantine persistence across resume rewrite and reload;
* successful verification clearing quarantine, rewriting the resume, and keeping the torrent stopped;
* aborted verification retaining quarantine.

### DIFF
--- a/libtransmission/quark.cc
+++ b/libtransmission/quark.cc
@@ -487,6 +487,7 @@ auto constexpr MyStatic = std::array<std::string_view, TR_N_KEYS>{
     "rename_partial_files"sv, // rpc, tr_session::Settings
     "reqq"sv, // BEP0010; BT protocol, rpc, tr_session::Settings
     "result"sv, // rpc
+    "resume_filenames_need_verification"sv, // .resume
     "rpc-authentication-required"sv, // daemon, rpc server settings
     "rpc-bind-address"sv, // daemon, rpc server settings
     "rpc-enabled"sv, // daemon, rpc server settings
--- a/libtransmission/quark.h
+++ b/libtransmission/quark.h
@@ -499,6 +499,7 @@ enum // NOLINT(performance-enum-size)
     TR_KEY_rename_partial_files,
     TR_KEY_reqq,
     TR_KEY_result,
+    TR_KEY_resume_filenames_need_verification,
     TR_KEY_rpc_authentication_required_kebab_APICOMPAT,
     TR_KEY_rpc_bind_address_kebab_APICOMPAT,
     TR_KEY_rpc_enabled_kebab_APICOMPAT,
--- a/libtransmission/resume.cc
+++ b/libtransmission/resume.cc
@@ -8,7 +8,10 @@
 #include <cstring>
 #include <ctime>
 #include <limits>
+#include <optional>
+#include <string>
 #include <string_view>
+#include <unordered_set>
 #include <vector>
 
 #include <fmt/format.h>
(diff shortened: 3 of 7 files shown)


ckerr:
Oop. Sorry @h1de0x I didn't notice you had a `transmission/transmission` PR for this already. I pushed up a PR for this at https://github.com/retransmission/retransmission/pull/205 before you pushed up 206 here.

I've picked some ideas out of this PR (I missed the mtime fix) and added you as co-author in 205.

One significant difference is that 205 doesn't quarantine. 206's `migrate_legacy_zero_file_layout()` only fires when all four arrays are still at length N-K. But `save_filenames()` unconditionally writes `file_count()` entries, so any 4.1.x client that loaded this torrent and saved its resume file would have left full-length arrays behind. The short-array signature is itself evidence that no unpatched 4.1 client ever acted on the shifted mapping. Quarantining the migrated torrent guards against a window that the detection condition has already excluded (barring a crash before the first resume save).

Closing this PR since there was already a fix in-flight but don't let that discourage you from future PRs.

PR #207, opened 2026-08-12
refactor: extract torrent cell from xib.

lolgear:
To increase flexibility when editing the torrent cell.

Extracting the cell from the XIB unlocks:
2. Implementing a proper `ProgressView` using `CALayer` to improve performance by offloading computations to the GPU.

### Results

This PR changes the behavior related to multiple invocations of `awakeFromNib` in `TorrentTableView` according to the [Apple documentation](https://apple.com).

> This method is usually called by the delegate in [tableView(_:viewFor:row:)](https://apple.com), but it can also be overridden to provide custom views for the identifier. Note that [awakeFromNib()](https://apple.com) is called each time this method is called, meaning that `awakeFromNib()` will trigger for the owner bundle repeatedly if the owner is passed down during instantiation.

### Before

Снимок экрана 2026-02-17 в 21 36 41

### After

Снимок экрана 2026-02-17 в 21 35 40

### DIFF
--- a/macosx/TorrentCell.mm
+++ b/macosx/TorrentCell.mm
@@ -16,6 +16,7 @@
 {
     if (self = [super initWithFrame:frameRect]) {
         [self configureViews];
+        [self configurePriorities];
         [self setupConstraints];
     }
     return self;
@@ -34,7 +35,6 @@
 
     auto torrentTitleField = [[NSTextField alloc] init];
     torrentTitleField.textColor = NSColor.labelColor;
-    torrentTitleField.backgroundColor = NSColor.textBackgroundColor;
     torrentTitleField.font = [NSFont systemFontOfSize:12.0 weight:NSFontWeightRegular];
 
     auto torrentPriorityView = [[NSImageView alloc] init];
@@ -44,12 +44,10 @@
 
     auto torrentProgressField = [[NSTextField alloc] init];
     torrentProgressField.textColor = NSColor.secondaryLabelColor;
-    torrentProgressField.backgroundColor = NSColor.textBackgroundColor;
     torrentProgressField.font = [NSFont systemFontOfSize:10.0 weight:NSFontWeightRegular];
 
     auto torrentStatusField = [[NSTextField alloc] init];
     torrentStatusField.textColor = NSColor.secondaryLabelColor;
-    torrentStatusField.backgroundColor = NSColor.textBackgroundColor;
     torrentStatusField.font = [NSFont systemFontOfSize:10.0 weight:NSFontWeightRegular];
 
     auto torrentProgressBarView = [[NSView alloc] init];
@@ -66,6 +64,7 @@
         textField.selectable = NO;
         textField.bordered = NO;
         textField.drawsBackground = NO;
+        textField.lineBreakMode = NSLineBreakByTruncatingMiddle;
     }
 
     for (NSButton* button in @[ actionButton, controlButton, revealButton ]) {
@@ -102,6 +101,46 @@
     self.fTorrentProgressBarView = torrentProgressBarView;
     self.fControlButton = controlButton;
     self.fRevealButton = revealButton;
+}
+
+- (void)configurePriorities
+{
+    auto groupIndicatorView = self.fGroupIndicatorView;
+    auto iconView = self.fIconView;
+    auto actionButton = self.fActionButton;
+    auto torrentTitleField = self.fTorrentTitleField;
+    auto torrentPriorityView = self.fTorrentPriorityView;
+    auto torrentProgressField = self.fTorrentProgressField;
+    auto torrentStatusField = self.fTorrentStatusField;
+    auto controlButton = self.fControlButton;
+    auto revealButton = self.fRevealButton;
+
+    // left items
+    for (NSView* leftView in @[ groupIndicatorView, iconView ]) {
+        [leftView setContentHuggingPriority:NSLayoutPriorityDefaultLow + 1 forOrientation:NSLayoutConstraintOrientationHorizontal];
+        [leftView setContentHuggingPriority:NSLayoutPriorityDefaultLow + 1 forOrientation:NSLayoutConstraintOrientationVertical];
+    }
+
+    [actionButton setContentHuggingPriority:NSLayoutPriorityDefaultHigh forOrientation:NSLayoutConstraintOrientationVertical];
+
+    /// middle items
+    [torrentTitleField setContentHuggingPriority:NSLayoutPriorityRequired forOrientation:NSLayoutConstraintOrientationHorizontal];
+    [torrentTitleField setContentHuggingPriority:NSLayoutPriorityDefaultHigh forOrientation:NSLayoutConstraintOrientationVertical];
+    [torrentPriorityView setContentHuggingPriority:NSLayoutPriorityRequired forOrientation:NSLayoutConstraintOrientationHorizontal];
+    [torrentPriorityView setContentHuggingPriority:NSLayoutPriorityDefaultLow + 1 forOrientation:NSLayoutConstraintOrientationVertical];
+
+    for (NSView* middleView in @[ torrentProgressField, torrentStatusField ]) {
+        [middleView setContentHuggingPriority:NSLayoutPriorityDefaultLow + 1 forOrientation:NSLayoutConstraintOrientationHorizontal];
+        [middleView setContentHuggingPriority:NSLayoutPriorityDefaultHigh forOrientation:NSLayoutConstraintOrientationVertical];
+    }
+
+    // right items
+    for (NSView* rightView in @[ controlButton, revealButton ]) {
+        [rightView setContentHuggingPriority:NSLayoutPriorityRequired forOrientation:NSLayoutConstraintOrientationHorizontal];
+        [rightView setContentHuggingPriority:NSLayoutPriorityDefaultHigh forOrientation:NSLayoutConstraintOrientationVertical];
+        [rightView setContentCompressionResistancePriority:NSLayoutPriorityRequired
+                                            forOrientation:NSLayoutConstraintOrientationHorizontal];
+    }
 }
 
 - (void)setupConstraints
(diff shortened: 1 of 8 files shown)


macosx/TorrentTableView.mm
```diff
@@ -256,7 +256,17 @@ - (NSView*)outlineView:(NSOutlineView*)outlineView viewForTableColumn:(NSTableCo
...
         } else {
-            torrentCell = [outlineView makeViewWithIdentifier:@"TorrentCell" owner:self];
+            // TODO(lolgear): Remove torrent cell from xib.
+            torrentCell = [outlineView makeViewWithIdentifier:@"NewTorrentCell" owner:self];
+            if (!torrentCell) {
+                torrentCell = [[TorrentCell alloc] initWithFrame:NSZeroRect];
+                torrentCell.identifier = @"NewTorrentCell";
+            }
+
+            ((TorrentCellControlButton*)torrentCell.fControlButton).torrentCell = torrentCell;
```
lolgear, line 241: It's ok, this clumsy code will also be removed later.

ckerr: Yep. claude flagged this too.

> Three casts, re-assigned on every viewForTableColumn: call for every recycled cell. configureViews creates these buttons and already knows their concrete types — actionButton.torrentCell = self; there does it once at construction. That also makes the three header changes unnecessary: the properties can stay in the class extensions where IB still reaches them (proof: the XIB currently connects them from there, MainMenu.xib:206,220,230). This is a ~10-line net deletion.

macosx/TorrentCell.mm
```diff
@@ -7,9 +7,171 @@
...
+    auto torrentProgressBarView = [[NSView alloc] init];
+
+    auto controlButton = [[TorrentCellControlButton alloc] init];
+    auto revealButton = [[TorrentCellRevealButton alloc] init];
+
+    for (NSImageView* imageView in @[ groupIndicatorView, iconView, torrentPriorityView ]) {
+        imageView.imageScaling = NSImageScaleProportionallyDown;
+    }
+
+    for (NSTextField* textField in @[ torrentTitleField, torrentProgressField, torrentStatusField ]) {
```
ckerr, line 62: The XIB set truncatingMiddle on title, progress, and status (lines 159, 185, 193); the new code sets none. Long torrent names lose the middle ellipsis.

GroupCell sets lineBreakMode explicitly on every field (GroupCell.mm:27), worth matching.

The XIB's hugging/compression priorities are also dropped (title and priority view at 1000, progress/status at 251, buttons 1000/1000)

ckerr (changes requested):
> `fTorrentPriorityViewWidthConstraint` (`TorrentCell.h`:20) is now never assigned on the programmatic path. It's unused everywhere in the codebase — delete it. Relatedly, the priority view's explicit 12×12 from the XIB is gone; harmless only because `PriorityHighTemplate.png` is 12×12.

macosx/TorrentCell.mm
lolgear: Oh no...
I forgot about hugging/compression :(

macosx/TorrentCell.h
```diff
@@ -17,7 +17,6 @@
 @property(nonatomic) IBOutlet NSStackView* fStackView;
 @property(nonatomic) IBOutlet NSTextField* fTorrentTitleField;
 @property(nonatomic) IBOutlet NSImageView* fTorrentPriorityView;
-@property(nonatomic) IBOutlet NSLayoutConstraint* fTorrentPriorityViewWidthConstraint;
```
ckerr, line 20: My first review was wrong!! This _is_ still used by SmallTorrentCell. I was wrong to recommend removing this line

macosx/TorrentCell.mm
```diff
@@ -7,9 +7,211 @@
...
+- (void)configurePriorities
+{
+    auto groupIndicatorView = self.fGroupIndicatorView;
+    auto iconView = self.fIconView;
+    auto actionButton = self.fActionButton;
+    auto torrentTitleField = self.fTorrentTitleField;
+    auto torrentPriorityView = self.fTorrentPriorityView;
+    auto torrentProgressField = self.fTorrentProgressField;
+    auto torrentStatusField = self.fTorrentStatusField;
+    auto torrentProgressBarView = self.fTorrentProgressBarView;
```
ckerr: the `torrentProgressBarView` local appears to be unused here; can we remove it?

macosx/TorrentCell.mm
```diff
@@ -7,9 +7,211 @@
...
+        [middleView setContentHuggingPriority:NSLayoutPriorityDefaultLow + 1 forOrientation:NSLayoutConstraintOrientationHorizontal];
+        [middleView setContentHuggingPriority:NSLayoutPriorityDefaultHigh forOrientation:NSLayoutConstraintOrientationVertical];
+    }
+
+    // right items
+    for (NSView* rightView in @[ controlButton, revealButton ]) {
+        [rightView setContentHuggingPriority:NSLayoutPriorityRequired forOrientation:NSLayoutConstraintOrientationHorizontal];
+        [rightView setContentHuggingPriority:NSLayoutPriorityDefaultHigh forOrientation:NSLayoutConstraintOrientationVertical];
+        [rightView setContentCompressionResistancePriority:NSLayoutPriorityRequired
+                                            forOrientation:NSLayoutConstraintOrientationVertical];
```
ckerr: Did you mean this to be `forOrientation:NSLayoutConstraintOrientationHorizontal`?

lolgear: Yes. Fixed it.

macosx/TorrentCell.mm
lolgear: Yes

macosx/TorrentTableView.mm
```diff
@@ -256,7 +256,17 @@ - (NSView*)outlineView:(NSOutlineView*)outlineView viewForTableColumn:(NSTableCo
                 torrentCell.fRevealButton.hidden = YES;
             }
         } else {
-            torrentCell = [outlineView makeViewWithIdentifier:@"TorrentCell" owner:self];
+            // TODO(lolgear): Remove torrent cell from xib.
+            torrentCell = [outlineView makeViewWithIdentifier:@"NewTorrentCell" owner:self];
+            if (!torrentCell) {
+                torrentCell = [[TorrentCell alloc] initWithFrame:NSZeroRect];
```
ckerr: > High: TorrentTableView.mm:262-276 constructs the cell with NSZeroRect, then immediately uses fIconView.frame to size an error-badge image. Before Auto Layout runs, that frame is zero, so newly created error rows can receive an empty image. Use the fixed icon size or lay out the cell before reading frames.

macosx/TorrentCell.mm
```diff
@@ -7,9 +7,210 @@
...
+        [torrentProgressField.topAnchor constraintEqualToAnchor:stackView.bottomAnchor constant:1],
+
+        // torrentStatusField
+        [torrentStatusField.leadingAnchor constraintEqualToAnchor:torrentProgressBarView.leadingAnchor],
+        [torrentStatusField.trailingAnchor constraintEqualToAnchor:torrentProgressBarView.trailingAnchor],
+        [torrentStatusField.topAnchor constraintEqualToAnchor:torrentProgressBarView.bottomAnchor constant:1],
+        [torrentStatusField.bottomAnchor constraintEqualToAnchor:self.bottomAnchor constant:-1],
+
+        // torrentProgressBarView
+        [torrentProgressBarView.centerYAnchor constraintEqualToAnchor:torrentProgressBarView.centerYAnchor],
```
ckerr: > TorrentCell.mm:184-197 constrains torrentProgressBarView.centerYAnchor to itself. The XIB instead had progressBar.top = progressField.bottom + 1. The current layout relies indirectly on the status field’s intrinsic height and may shift with different metrics. Restore the explicit top constraint.

lolgear: WHAT?

macosx/TorrentTableView.mm
lolgear: Oh..
Let's move it to `TorrentCell` later.
I left a TODO.

macosx/TorrentCell.mm
ckerr: >> WHAT?

> [torrentProgressBarView.centerYAnchor constraintEqualToAnchor:torrentProgressBarView.centerYAnchor],

The same `torrentProgresBarView.centerYAnchor` is on both the left-hand-side and right-hand-side of this statement. Appears to be a copy-paste error?

lolgear: Yes, I guess.
Not enough vibe-coding, only copy-pasting.

macosx/TorrentTableView.mm
ckerr: The fix is a one-liner. There's no benefit to waiting until later to fix it.

```ObjC++
if (!torrentCell) {
    torrentCell = [[TorrentCell alloc] initWithFrame:NSZeroRect];
    torrentCell.identifier = @"NewTorrentCell";
    [torrentCell layoutSubtreeIfNeeded];
}
```

ckerr (changes requested):

macosx/TorrentCell.mm
```diff
@@ -7,9 +7,209 @@
...
+        [torrentProgressField.topAnchor constraintEqualToAnchor:stackView.bottomAnchor constant:1],
+
+        // torrentStatusField
+        [torrentStatusField.leadingAnchor constraintEqualToAnchor:torrentProgressBarView.leadingAnchor],
+        [torrentStatusField.trailingAnchor constraintEqualToAnchor:torrentProgressBarView.trailingAnchor],
+        [torrentStatusField.topAnchor constraintEqualToAnchor:torrentProgressBarView.bottomAnchor constant:1],
+        [torrentStatusField.bottomAnchor constraintEqualToAnchor:self.bottomAnchor constant:-1],
+
+        // torrentProgressBarView
+        [torrentProgressBarView.heightAnchor constraintEqualToConstant:14],
```
ckerr, line 196:
```suggestion
        // torrentProgressBarView
        [torrentProgressBarView.topAnchor constraintEqualToAnchor:torrentProgressField.bottomAnchor constant:1],
        [torrentProgressBarView.heightAnchor constraintEqualToConstant:14],
```

macosx/TorrentTableView.mm
```diff
@@ -256,7 +256,18 @@ - (NSView*)outlineView:(NSOutlineView*)outlineView viewForTableColumn:(NSTableCo
...
```
ckerr:
```suggestion
                torrentCell = [[TorrentCell alloc] initWithFrame:NSMakeRect(0, 0, NSWidth(outlineView.bounds), self.rowHeight)];
}
```

ckerr (changes requested):

ckerr (approved):

PR #208, opened 2026-08-12
perf: remove reloadVisibleRows on selection did change in torrents and files table views

lolgear:

These methods don't do anything except "reload visible cells" one more time.
But Controller already did it every second.

I checked with this snippet

```objective-c++
// Inside -(void)outlineViewSelectionDidChange {
static uint64_t totalNanoseconds = 0;
static uint64_t callCount = 0;

uint64_t start = mach_absolute_time();

// --- body of outlineViewSelectionDidChange... ---

[self reloadVisibleRows]; // comment out this line for "native" approach.

uint64_t end = mach_absolute_time();

mach_timebase_info_data_t info;
mach_timebase_info(&info);
uint64_t elapsedNano = (end - start) * info.numer / info.denom;

totalNanoseconds += elapsedNano;
callCount++;

// Average time every 10 calls (select and press and hold down arrow).
if (callCount % 10 == 0) {
    NSLog(@"[PERF_TEST] Calls: %llu | Avg time per cell: %.3f ms",
          callCount, (double)totalNanoseconds / callCount / 1000000.0);
}
```

### Results

| Selection Ticks / Calls | Before Optimization (main) | After Optimization (This PR) | Speedup / Improvement |
|---|---|---|---|
| 130 ticks | 15.826 ms | 0.000 ms | Instant (O(1)) |
| 140 ticks | 15.788 ms | 0.000 ms | Instant (O(1)) |
| 150 ticks | 15.836 ms | 0.000 ms | Instant (O(1)) |
| 160 ticks | 15.839 ms | 0.000 ms | Instant (O(1)) |
| 170 ticks | 15.886 ms | 0.000 ms | Instant (O(1)) |
| 180 ticks | 15.860 ms | 0.000 ms | Instant (O(1)) |
| 190 ticks | 15.813 ms | 0.000 ms | Instant (O(1)) |
| 200 ticks | 15.806 ms | 0.000 ms | Instant (O(1)) |
| 210 ticks | 15.832 ms | 0.000 ms | Instant (O(1)) |
| Average | 15.832 ms | 0.000 ms | Pure Native Speed |

### Key Takeaway:
Spending ~15.8 ms of main thread time on a single click or keyboard tick is dangerously high (nearly dropping a frame at 60 FPS). When a user holds down the arrow keys to scroll through torrents, this creates an enormous rendering bottleneck, explaining the high CPU reports on macOS.
With this PR, the overhead is reduced to 0.000 ms because AppKit natively handles row highlighting states via CoreAnimation layers without forcing full cell view modernizations. No data is lost or delayed since the 1-second background timer still ensures text/progress values are perfectly accurate.

### DIFF
--- a/macosx/TorrentTableView.mm
+++ b/macosx/TorrentTableView.mm
@@ -71,8 +71,6 @@ static NSTimeInterval const kToggleProgressSeconds = 0.175;
 
 @property(nonatomic) NSDictionary* fHoverEventDict;
 
-@property(nonatomic) NSMutableIndexSet* fPendingSelectionReloadRows;
-
 @end
 
 @implementation TorrentTableView
@@ -397,35 +395,6 @@ static NSTimeInterval const kToggleProgressSeconds = 0.175;
     } else {
         return [self.dataSource outlineView:outlineView objectValueForTableColumn:[self tableColumnWithIdentifier:@"Group"]
                                      byItem:item];
-    }
-}
-
-- (void)outlineViewSelectionDidChange:(NSNotification*)notification
-{
-    NSIndexSet* oldSelection = self.fSelectedRowIndexes ?: [NSIndexSet indexSet];
-    NSIndexSet* newSelection = self.selectedRowIndexes;
-    self.fSelectedRowIndexes = newSelection;
-
-    NSIndexSet* changedRows = [oldSelection symmetricDifference:newSelection];
-    if (changedRows.count > 0) {
-        if (!self.fPendingSelectionReloadRows) {
-            self.fPendingSelectionReloadRows = [[NSMutableIndexSet alloc] init];
-            [self performSelector:@selector(flushSelectionReload) withObject:nil afterDelay:0 inModes:@[ NSRunLoopCommonModes ]];
-        }
-
-        [self.fPendingSelectionReloadRows addIndexes:changedRows];
-    }
-}
-
-- (void)flushSelectionReload
-{
-    NSMutableIndexSet* rows = self.fPendingSelectionReloadRows;
-    self.fPendingSelectionReloadRows = nil;
-
-    NSInteger const numberOfRows = self.numberOfRows;
-    [rows removeIndexesInRange:NSMakeRange(numberOfRows, NSIntegerMax - numberOfRows)];
-    if (rows.count > 0) {
-        [self reloadDataForRowIndexes:rows columnIndexes:[NSIndexSet indexSetWithIndex:0]];
     }
 }
 
--- a/macosx/TorrentTableView.mm
+++ b/macosx/TorrentTableView.mm
@@ -28,28 +28,6 @@ static CGFloat const kErrorImageSize = 20.0;
 
 static NSTimeInterval const kToggleProgressSeconds = 0.175;
 
-@interface NSIndexSet (Transmission)
-- (NSIndexSet*)symmetricDifference:(NSIndexSet*)otherSet;
-@end
-
-@implementation NSIndexSet (Transmission)
-
-- (NSIndexSet*)symmetricDifference:(NSIndexSet*)otherSet
-{
-    NSMutableIndexSet* result = [self mutableCopy];
-    [result addIndexes:otherSet];
-
-    [self enumerateIndexesUsingBlock:^(NSUInteger idx, BOOL*) {
-        if ([otherSet containsIndex:idx]) {
-            [result removeIndex:idx];
-        }
-    }];
-
-    return [result copy];
-}
-
-@end
-
 @interface TorrentTableView ()
 
 @property(nonatomic) IBOutlet Controller* fController;
@@ -60,8 +38,6 @@ static NSTimeInterval const kToggleProgressSeconds = 0.175;
 
 @property(nonatomic) IBOutlet NSMenu* fContextRow;
 @property(nonatomic) IBOutlet NSMenu* fContextNoRow;
-
-@property(nonatomic) NSIndexSet* fSelectedRowIndexes;
 
 @property(nonatomic) CGFloat piecesBarPercent;
 @property(nonatomic) NSAnimation* fPiecesBarAnimation;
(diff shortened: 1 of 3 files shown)


macosx/TorrentTableView.mm
```diff
@@ -400,35 +398,6 @@ - (NSString*)outlineView:(NSOutlineView*)outlineView typeSelectStringForTableCol
     }
 }
 
-- (void)outlineViewSelectionDidChange:(NSNotification*)notification
-{
-    NSIndexSet* oldSelection = self.fSelectedRowIndexes ?: [NSIndexSet indexSet];
-    NSIndexSet* newSelection = self.selectedRowIndexes;
-    self.fSelectedRowIndexes = newSelection;
```
ckerr, line 407: `fSelectedRowIndexes` is now unused, its `@property` declaration at the top of the file should be removed

macosx/TorrentTableView.mm
```diff
@@ -400,35 +398,6 @@ - (NSString*)outlineView:(NSOutlineView*)outlineView typeSelectStringForTableCol
...
-
-    NSIndexSet* changedRows = [oldSelection symmetricDifference:newSelection];
```
ckerr, line 409: `symmetricDifference` is now unused; its `@interface` declaration at the top of the file should be removed

ckerr (changes requested):
`TorrentCell.mm:24` picks the priority arrow's color from `self.backgroundStyle` inside `drawRect:,` and `TorrentCell` does not override `setBackgroundStyle:`.  AppKit propagates `backgroundStyle` to subviews but is not documented to invalidate the cell view's own drawing, so a high/low-priority torrent likely keeps the stale white/dark arrow until the next  `updateUI` call. 

Every other selection-aware cell here (eg `GroupCell.mm:109`, `BaseFileNameCellView.mm:126`, `FilePriorityCellView.mm:206`) reacts in a `setBackgroundStyle:` override, which is why those are safe under this change. `TorrentCell` is the one that needed the reload.

```diff
--- a/macosx/TorrentCell.mm
+++ b/macosx/TorrentCell.mm
@@ -34,6 +34,15 @@
 
     [super drawRect:dirtyRect];
 }
 
+// -drawRect: tints the priority icon for the current backgroundStyle,
+// so a change of style needs a redraw
+- (void)setBackgroundStyle:(NSBackgroundStyle)backgroundStyle
+{
+    BOOL const changed = backgroundStyle != self.backgroundStyle;
+    [super setBackgroundStyle:backgroundStyle];
+    if (changed) {
+        self.needsDisplay = YES;
+    }
+}
+
 // otherwise progress bar is inverted
 - (BOOL)isFlipped
 {
     return YES;
 }
```

lolgear:
@ckerr 
Good point for TorrentCell `backgroundStyle`.

I think that the change should be a little different.

Since `imageWithColor` doesn't work as expect ( I revealed it in previous PR with FilePriorityCell ), it seems that the change is more elegant.

```objective-c++
- (void)setBackgroundStyle:(NSBackgroundStyle)backgroundStyle
{
    [super setBackgroundStyle:backgroundStyle];

    NSColor* priorityColor = backgroundStyle == NSBackgroundStyleEmphasized ? NSColor.whiteColor : NSColor.labelColor;
    self.fTorrentPriorityView.contentTintColor = priorityColor;
}
```

macosx/TorrentTableView.mm
lolgear: Done

macosx/TorrentTableView.mm
lolgear: Done

ckerr:
> @ckerr Good point for TorrentCell `backgroundStyle`.
> 
> I think that the change should be a little different.
> 
> Since `imageWithColor` doesn't work as expect ( I revealed it in previous PR with FilePriorityCell ), it seems that the change is more elegant.
> 
> ```objc++
> - (void)setBackgroundStyle:(NSBackgroundStyle)backgroundStyle
> {
>     [super setBackgroundStyle:backgroundStyle];
> 
>     NSColor* priorityColor = backgroundStyle == NSBackgroundStyleEmphasized ? NSColor.whiteColor : NSColor.labelColor;
>     self.fTorrentPriorityView.contentTintColor = priorityColor;
> }
> ```

(Claude)

Agreed — `contentTintColor` is the right mechanism, and it's the pattern we already settled on elsewhere. One addition, though: as written the snippet is only half the change, and the missing half is exactly why `imageWithColor` didn't behave for you before.

`contentTintColor` only applies to a template image. `-drawRect:` currently assigns `[[NSImage imageNamed:@"…Template"] imageWithColor:priorityColor]`, and `imageWithColor:` (macosx/NSImageAdditions.mm:37) bakes the color into a copy with `lockFocus` + a source-atop fill. Adding the setter without touching that leaves two tinting mechanisms fighting over the same image view, and which one wins depends on whether the baked copy still carries the template flag. f4a49ede7 (#8324) handled this as a pair for `FilePriorityCellView`: it deleted the `imageWithColor:` calls, assigned the plain `…Template` images, and moved the color to `contentTintColor`.

So alongside your `setBackgroundStyle:` override:

```diff
--- a/macosx/TorrentCell.mm
+++ b/macosx/TorrentCell.mm
@@
-#import "NSImageAdditions.h"
@@
         // set priority icon
         if (torrent.priority != TR_PRI_NORMAL) {
-            NSColor* priorityColor = self.backgroundStyle == NSBackgroundStyleEmphasized ? NSColor.whiteColor : NSColor.labelColor;
-            NSImage* priorityImage = [[NSImage imageNamed:(torrent.priority == TR_PRI_HIGH ? @"PriorityHighTemplate" : @"PriorityLowTemplate")]
-                imageWithColor:priorityColor];
-
-            self.fTorrentPriorityView.image = priorityImage;
+            self.fTorrentPriorityView.image = [NSImage
+                imageNamed:(torrent.priority == TR_PRI_HIGH ? @"PriorityHighTemplate" : @"PriorityLowTemplate")];
 
             [self.fStackView setVisibilityPriority:NSStackViewVisibilityPriorityMustHold forView:self.fTorrentPriorityView];
```

`NSImageAdditions.h` was imported solely for `imageWithColor:`, so it can be removed too.

This is strictly better than the `needsDisplay` invalidation I suggested, so please ignore that one: with the tint owned by the image view, `-drawRect:` stops reading `backgroundStyle` altogether, which means no override is needed and selection changes no longer force a full-cell redraw.

ckerr (approved):

PR #209, opened 2026-08-12
perf: file name cell icon caching added.

lolgear:
In previous https://github.com/transmission/transmission/pull/8497 PR file name cell was slightly optimized.

This PR continues optimization by adding icon caching.

I checked with this snippet.

```objective-c++
#import <mach/mach_time.h>

// Inside -(void)updateDisplay {
static uint64_t totalNanoseconds = 0;
static uint64_t callCount = 0;

uint64_t start = mach_absolute_time();

// --- body of updateDisplay... ---

uint64_t end = mach_absolute_time();

mach_timebase_info_data_t info;
mach_timebase_info(&info);
uint64_t elapsedNano = (end - start) * info.numer / info.denom;

totalNanoseconds += elapsedNano;
callCount++;

// Average time every 500 calls (scrolling).
if (callCount % 500 == 0) {
    NSLog(@"[PERF_TEST] Calls: %llu | Avg time per cell: %.3f ms", 
          callCount, (double)totalNanoseconds / callCount / 1000000.0);
}
// }
// [self updateColors];
// ...
// } // end of -(void)updateDisplay
```

#### Before cells separation

| Total Rendered Cells | Before Optimization (main) | After Optimization (This PR) | Speedup / Improvement |
|---|---|---|---|
| 1,000 cells | 0.534 ms | 0.409 ms | -23.4% time spent |
| 1,500 cells | 0.539 ms | 0.357 ms | -33.8% time spent |
| 2,000 cells | 0.540 ms | 0.377 ms | -30.2% time spent |
| 2,500 cells | 0.543 ms | 0.389 ms | -28.4% time spent |
| 3,000 cells | 0.533 ms | 0.397 ms | -25.5% time spent |
| Average | 0.538 ms | 0.386 ms | ~1.39x Faster |

#### After cells separation with icon caching

| Total Rendered Cells | Before Optimization (main) | After Optimization (This PR) | Speedup / Improvement |
|---|---|---|---|
| 1,000 cells | 0.409 ms | 0.138 ms | -66.3% time spent |
| 1,500 cells | 0.357 ms | 0.134 ms | -62.5% time spent |
| 2,000 cells | 0.377 ms | 0.131 ms | -65.3% time spent |
| 2,500 cells | 0.389 ms | 0.129 ms | -66.8% time spent |
| 3,000 cells | 0.397 ms | 0.128 ms | -67.8% time spent |
| Average | 0.386 ms | 0.132 ms | ~2.9x Faster |

#### Before Optimization

https://github.com/user-attachments/assets/8d425ffb-f043-4402-923d-fd27f7947ca7

#### After Optimization

https://github.com/user-attachments/assets/118224dc-eacb-4402-ad95-54f1da35c59f

### DIFF
--- a/macosx/BaseFileNameCellView.mm
+++ b/macosx/BaseFileNameCellView.mm
@@ -2,6 +2,9 @@
 // It may be used under the MIT (SPDX: MIT) license.
 // License text can be found in the licenses/ folder.
 
+#include <libtransmission/transmission.h>
+#import <UniformTypeIdentifiers/UniformTypeIdentifiers.h>
+
 #import "BaseFileNameCellView.h"
 #import "FileListNode.h"
 #import "Torrent.h"
@@ -22,7 +25,27 @@ static CGFloat const kPaddingBetweenNameAndFolderStatus = 4.0;
 @property(nonatomic, strong) NSArray<NSLayoutConstraint*>* dynamicConstraints;
 @end
 
+static NSCache<UTType*, NSImage*>* iconCache;
+
 @implementation BaseFileNameCellView
+
++ (void)initialize
+{
+    iconCache = [[NSCache alloc] init];
+}
+
++ (NSImage*)cachedIconForUTType:(UTType*)uttype
+{
+    auto cachedIcon = [iconCache objectForKey:uttype];
+
+    if (cachedIcon == nil) {
+        auto icon = [NSWorkspace.sharedWorkspace iconForContentType:uttype];
+        [iconCache setObject:icon forKey:uttype];
+        return icon;
+    }
+
+    return cachedIcon;
+}
 
 - (instancetype)initWithFrame:(NSRect)frameRect
 {
@@ -70,6 +93,10 @@ static CGFloat const kPaddingBetweenNameAndFolderStatus = 4.0;
 {
 }
 
+- (void)updateImageIfNeeded
+{
+}
+
 - (void)setNode:(FileListNode*)node
 {
     _node = node;
@@ -85,7 +112,7 @@ static CGFloat const kPaddingBetweenNameAndFolderStatus = 4.0;
     FileListNode* node = self.node;
 
     // Update icon
-    self.iconView.image = node.icon;
+    [self updateImageIfNeeded];
 
     // Update name
     self.nameField.stringValue = node.name;
@@ -154,6 +181,18 @@ static CGFloat const kPaddingBetweenNameAndFolderStatus = 4.0;
 
 @implementation FileNameCellView
 
+- (void)updateImageIfNeeded
+{
+    if (self.node.name == nil) {
+        return;
+    }
+
+    auto uttype = [UTType typeWithFilenameExtension:self.node.name.pathExtension];
+    auto image = [self.class cachedIconForUTType:uttype];
+
+    self.iconView.image = image;
+}
+
 - (void)setupConstraints
 {
     NSImageView* iconView = self.iconView;
@@ -184,6 +223,14 @@ static CGFloat const kPaddingBetweenNameAndFolderStatus = 4.0;
 @end
 
 @implementation FolderNameCellView
+
+- (instancetype)initWithFrame:(NSRect)frameRect
+{
+    if (self = [super initWithFrame:frameRect]) {
+        self.iconView.image = [self.class cachedIconForUTType:UTTypeFolder];
+    }
+    return self;
+}
 
 - (void)setupConstraints
 {
--- a/macosx/BaseFileNameCellView.mm
+++ b/macosx/BaseFileNameCellView.mm
@@ -2,7 +2,6 @@
 // It may be used under the MIT (SPDX: MIT) license.
 // License text can be found in the licenses/ folder.
 
-#include <libtransmission/transmission.h>
 #import <UniformTypeIdentifiers/UniformTypeIdentifiers.h>
 
 #import "BaseFileNameCellView.h"
@@ -22,7 +21,6 @@ static CGFloat const kPaddingBetweenNameAndFolderStatus = 4.0;
 @property(nonatomic, weak) NSImageView* iconView;
 @property(nonatomic, weak) NSTextField* nameField;
 @property(nonatomic, weak) NSTextField* statusField;
-@property(nonatomic, strong) NSArray<NSLayoutConstraint*>* dynamicConstraints;
 @end
 
 static NSCache<UTType*, NSImage*>* iconCache;
@@ -31,7 +29,9 @@ static NSCache<UTType*, NSImage*>* iconCache;
 
 + (void)initialize
 {
-    iconCache = [[NSCache alloc] init];
+    if (self == [BaseFileNameCellView class]) {
+        iconCache = [[NSCache alloc] init];
+    }
 }
 
 + (NSImage*)cachedIconForUTType:(UTType*)uttype
@@ -93,7 +93,7 @@ static NSCache<UTType*, NSImage*>* iconCache;
 {
 }
 
-- (void)updateImageIfNeeded
+- (void)updateImage
 {
 }
 
@@ -112,7 +112,7 @@ static NSCache<UTType*, NSImage*>* iconCache;
     FileListNode* node = self.node;
 
     // Update icon
-    [self updateImageIfNeeded];
+    [self updateImage];
 
     // Update name
     self.nameField.stringValue = node.name;
@@ -181,16 +181,24 @@ static NSCache<UTType*, NSImage*>* iconCache;
 
 @implementation FileNameCellView
 
-- (void)updateImageIfNeeded
+- (void)updateImage
 {
-    if (self.node.name == nil) {
+    NSString* const name = self.node.name;
+
+    if (name == nil) {
         return;
     }
 
-    auto uttype = [UTType typeWithFilenameExtension:self.node.name.pathExtension];
-    auto image = [self.class cachedIconForUTType:uttype];
+    // typeWithFilenameExtension: returns nil when the name carries no extension
+    // ("README", ".gitignore"), so fall back to the generic item icon.
+    auto uttype = [UTType typeWithFilenameExtension:name.pathExtension] ?: UTTypeItem;
 
-    self.iconView.image = image;
+    // cachedIconForUTType: returns one shared instance per type, so this
+    // compares identity rather than contents.
+    auto image = [self.class cachedIconForUTType:uttype];
+    if (self.iconView.image != image) {
+        self.iconView.image = image;
+    }
 }
 
 - (void)setupConstraints


macosx/BaseFileNameCellView.mm
```diff
@@ -154,6 +181,18 @@ - (void)updateColors
...
 @implementation FileNameCellView
 
+- (void)updateImageIfNeeded
+{
+    if (self.node.name == nil) {
+        return;
+    }
+
+    auto uttype = [UTType typeWithFilenameExtension:self.node.name.pathExtension];
+    auto image = [self.class cachedIconForUTType:uttype];
```
ckerr: `+[UTType typeWithFilenameExtension:]` is `nullable`, and the case that returns nil is an empty extension: unknown extensions get a dynamic `dyn.*` type instead. `pathExtension` returns `@""` for `README`, `LICENSE`, `data`, and for dotfiles like `.gitignore`.

The nil then flows unguarded into `cachedIconForUTType:`, where every consumer rejects it: `iconForContentType:` is nonnull-annotated, and `-[NSCache setObject:forKey:]` raises `NSInvalidArgumentException` on a nil key or object. Best case the user sees a blank icon where `iconForFileType:@""` previously returned the generic document icon; worst case it throws.

Note the existing guard checks `self.node.name == nil`, which is the case that essentially never happens, while the common one is unguarded. Falling back to `UTTypeItem` (or `UTTypeData`) covers both:

```suggestion
    auto uttype = [UTType typeWithFilenameExtension:self.node.name.pathExtension] ?: UTTypeItem;
    auto image = [self.class cachedIconForUTType:uttype];
```

macosx/BaseFileNameCellView.mm
```diff
@@ -22,8 +25,28 @@ @interface BaseFileNameCellView ()
 @property(nonatomic, strong) NSArray<NSLayoutConstraint*>* dynamicConstraints;
 @end
 
+static NSCache<UTType*, NSImage*>* iconCache;
+
 @implementation BaseFileNameCellView
 
++ (void)initialize
+{
```
ckerr, line 31: `+initialize` is inherited, so it runs for `BaseFileNameCellView`, `FileNameCellView`, and `FolderNameCellView` and each run assigns a fresh `NSCache`, discarding what the previous one held. The folder icon cached during the first `FolderNameCellView` init is thrown away when the first `FileNameCellView` is created.

Impact is a couple of redundant `LaunchServices` calls at startup. Not a blocker, but the code reads as if the cache is shared and stable and it isn't. `TrackerCell.mm:37-44` already has the guard this needs:

```diff
+ (void)initialize
{
    if (self != [BaseFileNameCellView self])
        return;
    ...
}
```

Or even better, drop `+initialize` and use a `dispatch_once` static local inside `cachedIconForUTType:`, matching the approach in `GetTorrentFileType()` at `macos/DefaultAppHelper.mm:16`.

macosx/BaseFileNameCellView.mm
```diff
@@ -2,6 +2,9 @@
 // It may be used under the MIT (SPDX: MIT) license.
 // License text can be found in the licenses/ folder.
 
+#include <libtransmission/transmission.h>
```
ckerr: It doesn't look like `transmission.h` is needed here?

macosx/BaseFileNameCellView.mm
```diff
@@ -154,6 +181,18 @@ - (void)updateColors
 
 @implementation FileNameCellView
 
+- (void)updateImageIfNeeded
```
ckerr: `updateImageIfNeeded` never checks whether it's needed.

It re-derives the `UTType` and re-assigns `iconView.image` on every `updateDisplay`, which is the hot path this PR is supposed to be optimizing for. `NSImageView` doesn't short-circuit an identical image, so each reuse dirties the view.

Either compare against the current image first, or rename it `updateImage` so the name doesn't imply a check it doesn't perform.

macosx/BaseFileNameCellView.mm
```diff
@@ -85,7 +112,7 @@ - (void)updateDisplay
     FileListNode* node = self.node;
 
     // Update icon
-    self.iconView.image = node.icon;
```
ckerr, line 88: Looks like this was the last use of `-[FileListNode icon]`, so we can remove it:

| Location | Item |
|--|--|
| FileListNode.h:17 | `@property(nonatomic, readonly) NSImage* icon;` |
| FileListNode.mm:16	| `@property(nonatomic) NSImage* iconInternal;` |
| FileListNode.mm:76-83 | the icon getter |
| FileListNode.mm:103 | `_iconInternal = nil;` in `updateFromOldName:` |
| FileListNode.mm:5-9 | `@import AppKit` / `#import <AppKit/AppKit.h>` |

ckerr (changes requested):
I requested changes, but this looks like a nice perf win. I'm looking forward to merging it

macosx/BaseFileNameCellView.mm
lolgear: Yes, but I would like to not to touch `FileListNode` in this PR.
I have another PR with some performance tricks for `FileListNode`.

ckerr: Unless `-[FileListNode icon]` is used in the other PR, then it's still dead code that we should remove?

lolgear: Yes..

lolgear: Can we move forward and remove later this unused property?

macosx/BaseFileNameCellView.mm
```diff
@@ -70,13 +93,32 @@ - (void)setupConstraints
...
+- (void)updateImageForOldNode:(FileListNode*)oldNode newNode:(FileListNode*)newNode
+{
+    if (newNode.name == nil) {
+        return;
+    }
+
+    if ([newNode.name isEqualToString:oldNode.name]) {
+        // same images
+        return;
+    }
```
ckerr: > renames mutate a FileListNode in place: `Torrent.mm`:1974 calls `-[FileListNode updateFromOldName:toNewName:inPath:]`, which assigns _name on the existing node, and `renameFinished:` only re-sorts afterward. New nodes are built solely in `createFileList`, which a rename never calls. So the cell gets back the same pointer it already held, `oldNode == newNode`, and the check compares the name to itself. Rename `foo.txt` to `foo.mp4` and the `.txt` icon stays.
>
> I recommended reverting the old/new plumbing entirely and comparing `self.iconView.image != image` instead. Immune to the aliasing, and a pointer compare, since the cache returns the same NSImage instance per UTType.

```diff
--- a/macosx/BaseFileNameCellView.mm
+++ b/macosx/BaseFileNameCellView.mm
@@ -97,28 +97,13 @@ static NSCache<UTType*, NSImage*>* iconCache;
 {
 }
 
--- (void)updateImageForOldNode:(FileListNode*)oldNode newNode:(FileListNode*)newNode
--{
--    if (newNode.name == nil) {
--        return;
--    }
--
--    if ([newNode.name isEqualToString:oldNode.name]) {
--        // same images
--        return;
--    }
--
--    [self updateImage];
--}
--
 - (void)setNode:(FileListNode*)node
 {
-    auto oldNode = _node;
     _node = node;
-    [self updateDisplayForOldNode:oldNode newNode:node];
+    [self updateDisplay];
 }
 
-- (void)updateDisplayForOldNode:(FileListNode*)oldNode newNode:(FileListNode*)newNode
+- (void)updateDisplay
 {
     if (!self.node) {
         return;
@@ -127,7 +112,7 @@ static NSCache<UTType*, NSImage*>* iconCache;
     FileListNode* node = self.node;
 
     // Update icon
-    [self updateImageForOldNode:oldNode newNode:newNode];
+    [self updateImage];
 
     // Update name
     self.nameField.stringValue = node.name;
@@ -198,10 +183,21 @@ static NSCache<UTType*, NSImage*>* iconCache;
 
 - (void)updateImage
 {
-    auto uttype = [UTType typeWithFilenameExtension:self.node.name.pathExtension] ?: UTTypeItem;
-    auto image = [self.class cachedIconForUTType:uttype];
+    NSString* const name = self.node.name;
+    if (name == nil) {
+        return;
+    }
+
+    // typeWithFilenameExtension: returns nil when the name carries no extension
+    // ("README", ".gitignore"), so fall back to the generic item icon.
+    auto uttype = [UTType typeWithFilenameExtension:name.pathExtension] ?: UTTypeItem;
 
-    self.iconView.image = image;
+    // cachedIconForUTType: returns one shared instance per type, so this
+    // compares identity rather than contents.
+    auto image = [BaseFileNameCellView cachedIconForUTType:uttype];
+    if (self.iconView.image != image) {
+        self.iconView.image = image;
+    }
 }
```

ckerr (changes requested):
(Claude)

Fixed: the transmission.h include, the +initialize guard, and the nil UTType fallback (?: UTTypeItem). updateImageIfNeeded was renamed to updateImage and given a real guard.

That guard is the new problem — and it's blocking. The author added updateImageForOldNode:newNode:, which skips the icon update when [newNode.name isEqualToString:oldNode.name]. But renames mutate a FileListNode in place: Torrent.mm:1974 calls -[FileListNode updateFromOldName:toNewName:inPath:], which assigns _name on the existing node, and renameFinished: only re-sorts afterward — new nodes are built solely in createFileList, which a rename never calls. So the cell gets back the same pointer it already held, oldNode == newNode, and the check compares the name to itself. Rename foo.txt → foo.mp4 and the .txt icon stays.

It's intermittent rather than always-wrong, which is worse: if makeViewWithIdentifier: happens to return a different recycled cell for that row, that cell's _node differs, the names differ, and the icon updates fine. Reproduces on one machine, not another.

I recommended reverting the old/new plumbing entirely and comparing self.iconView.image != image instead — immune to the aliasing, and a pointer compare, since the cache returns the same NSImage instance per UTType. Worth saying out loud that the guard buys very little either way: the PR's own numbers show the win came from killing the LaunchServices call, not from skipping an NSCache lookup.

Two smaller things. updateDisplayForOldNode:newNode: takes a newNode that is always _node, assigned one line earlier and re-read from self.node in the body — only one of the two new parameters carries information. And the dynamicConstraints property removal is a genuine dead-code cleanup (it was already unused before this PR), but it isn't icon caching, so it wants its own commit.

Carried over unaddressed: self.class at lines 202/241, the parens at 240, and the -[FileListNode icon] dead code. I switched the file to -[Class method] notation throughout, per your earlier question.

macosx/BaseFileNameCellView.mm
lolgear: Ok, let's compare pointers

macosx/BaseFileNameCellView.mm
```diff
@@ -70,8 +93,13 @@ - (void)setupConstraints
 {
 }
 
+- (void)updateImage
+{
+}
+
 - (void)setNode:(FileListNode*)node
 {
+    auto oldNode = _node;
```
ckerr: `oldNode` is unused

ckerr (changes requested):

ckerr (approved):

PR #213, opened 2026-08-12
fix: discard old value in `tr_bitfield::set_from_bools`

tearfur:
## Description

`tr_bitfield::set_from_bools` should discard any old values before setting values from the Boolean array.

Production code does not trigger this, so I'm marking as `notes:none`.

## Checklist

### DIFF
--- a/libtransmission/bitfield.cc
+++ b/libtransmission/bitfield.cc
@@ -271,7 +271,7 @@ void tr_bitfield::set_from_bools(std::span<bool const> const flags)
     size_t true_count = 0;
 
     free_array();
-    ensure_bits_alloced(flags.size());
+    flags_.resize(getBytesNeeded(flags.size()));
 
     for (size_t i = 0; i < flags.size(); ++i) {
         if (flags[i]) {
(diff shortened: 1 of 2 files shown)


libtransmission/bitfield.cc
```diff
@@ -271,7 +271,7 @@ void tr_bitfield::set_from_bools(std::span<bool const> const flags)
     size_t true_count = 0;
 
     free_array();
-    ensure_bits_alloced(flags.size());
+    flags_.resize(getBytesNeeded(flags.size()));
```
ckerr: We can try to reuse the existing capacity by using `assign()` here instead of free + resize:

```diff
-  free_array();
-  flags_.resize(getBytesNeeded(flags.size()));
+  flags_.assign(getBytesNeeded(flags.size()), std::byte{}) ;
```

ckerr (changes requested):
Good catch!

libtransmission/bitfield.cc
tearfur: cd4a2a7b62c9e72b804e0c228384fb167d28fc0c

ckerr (approved):

PR #215, opened 2026-08-13
refactor: rework `tr_bitfield` to have coherent internal states

tearfur:
## Description

Reworked `tr_bitfield` so that its internal logic is consistent and safe. This indirectly fixes an over-allocation attack vector from `torrent_add.files_[un]wanted` that's possible because `tr_bitfield` did not cap the bit position to set.

1. Enforce "have all" and "have none" being the only valid states when `bit_count_ == 0`
2. All `set*()` methods will now reject changes that tries to write beyond `bit_count`
3. As a result of (2), `true_count_ > bit_count_` will no longer be a possible state, and the size of `flags_` will never be bigger than the number of bytes needed to store `bit_count_` bits.
4. `is_valid()` now reflects (2) and (3), as well as enforcing the fact that excess bits should be set to 0. More assertions are added.
5. `bit_count_ == 0` is now a well-defined concept (a.k.a. size is not yet known), and `empty()` has been rewritten to `is_size_known()` to reflect this.
6. ~To accommodate peer `HAVE` bitfields before we got the metainfo for a magnet link, a new method `init_size()` was added to allow setting the bitfield size after constructing the bitfield with size 0.~
7. `raw()` now always return the correct size.

Notes: Hardened the `torrent-add` RPC method against malformed requests.

### Attack vector

1. Client adds a torrent using `torrent_add` with argument `"files_unwanted": [SIZE_MAX - 1]`.
 https://github.com/retransmission/retransmission/blob/541ea54ae12c6f2d278a6f91bc9c837288199447/libtransmission/rpcimpl.cc#L1593-L1594

2. Value of `unwanted_` is `{ SIZE_MAX - 1 }`.
  https://github.com/retransmission/retransmission/blob/541ea54ae12c6f2d278a6f91bc9c837288199447/libtransmission/torrent-builder.cc#L110-L112

3. Value of `files` is `{ SIZE_MAX - 1 }`.
  https://github.com/retransmission/retransmission/blob/541ea54ae12c6f2d278a6f91bc9c837288199447/libtransmission/torrent.h#L1210

4. Value of `file` is `SIZE_MAX - 1`.
  https://github.com/retransmission/retransmission/blob/541ea54ae12c6f2d278a6f91bc9c837288199447/libtransmission/file-piece-map.cc#L213-L214

5. Value of `file` is `SIZE_MAX - 1`.
  https://github.com/retransmission/retransmission/blob/541ea54ae12c6f2d278a6f91bc9c837288199447/libtransmission/file-piece-map.cc#L208

6. Value of `nth` is `SIZE_MAX - 1`.
  https://github.com/retransmission/retransmission/blob/541ea54ae12c6f2d278a6f91bc9c837288199447/libtransmission/bitfield.cc#L291

8. Value of `file` is `SIZE_MAX - 1`.
  https://github.com/retransmission/retransmission/blob/541ea54ae12c6f2d278a6f91bc9c837288199447/libtransmission/bitfield.cc#L188

9. Value of `bytes_needed` is `(SIZE_MAX + 7) >> 3` (assume no integer overflow) -> `std::length_error` or `std::bad_alloc`.
  https://github.com/retransmission/retransmission/blob/541ea54ae12c6f2d278a6f91bc9c837288199447/libtransmission/bitfield.cc#L174

## Checklist


tearfur:
This is a big change to a class that's essential to BitTorrent operations, so I'd suggest more careful review than usual.

ckerr (commented):
> To accommodate peer HAVE bitfields before we got the metainfo for a magnet link, a new method `init_size()` was added to allow setting the bitfield size after constructing the bitfield with size 0.

`init_size()` doesn't appear to be used. Is this intentional?

tearfur:
> > To accommodate peer HAVE bitfields before we got the metainfo for a magnet link, a new method `init_size()` was added to allow setting the bitfield size after constructing the bitfield with size 0.
> 
> `init_size()` doesn't appear to be used. Is this intentional?

It's meant for #218. I'll remove it in this PR and put it in #218.

tearfur:
@ckerr Ready for review again. I've made a couple more changes since you last review as I've noticed more things.

libtransmission/bitfield.cc
```diff
@@ -17,61 +17,66 @@
...
     }
 
-    if (first_byte >= std::size(flags_)) {
+    if (has_all()) {
+        return end - begin;
+    }
+
+    if (!is_size_known()) [[unlikely]] {
         return 0;
     }
```
ckerr: - Clamping `end` before the `has_all()` check changes what an unknown-size "have all" bitfield reports. A magnet peer that sends `FextHaveAll` has `have_all_hint_` set with `bit_count_ == 0`, so `end` clamps to 0, `begin >= end` returns at the line below, and `count(0, 10)` is 0 — it returned `end - begin` before. `percent()` still returns 1.0F for that same object, so the two accessors now disagree.

- That also makes the `if (!is_size_known()) [[unlikely]]` guard on line 69 dead: `bit_count_ == 0` always exits at the `begin >= end` check above it. Hoisting the unknown-size and `has_all()` cases above the clamp would fix both.

```suggestion
    if (is_size_known()) {
        end = std::min(end, bit_count_);
    }

    if (begin >= end) {
        return 0;
    }

    if (has_all()) {
        return end - begin;
    }
}
```

libtransmission/bitfield.cc
```diff
@@ -109,89 +114,82 @@ size_t tr_bitfield::count_flags(size_t begin, size_t end) const noexcept
...
-
+    auto const n = getBytesNeeded(bit_count_);
     auto raw = std::vector<std::byte>(n);
 
     if (has_all()) {
         setAllTrue(raw, bit_count_);
+    } else if (!std::empty(flags_)) {
+        // flags_ does not always contain all bit_count_ bits.
+        // E.g. After `set_raw()` was called with a smaller vector.
+        std::ranges::copy(flags_, raw.begin());
```
ckerr: Worth making this copy fail-safe rather than assertion-safe. `raw` holds exactly `getBytesNeeded(bit_count_)` bytes, so this overflows if `flags_` is ever longer. Every current writer upholds that invariant and `is_valid()` checks it, but only inside `TR_ASSERT`, which is compiled out under NDEBUG — so in release builds nothing stands between a future invariant break and a heap overflow.

`std::ranges::copy_n(std::begin(flags_), std::min(std::size(flags_), n), std::begin(raw))` costs nothing and can't overflow however `flags_` got its size.


```suggestion
        std::ranges::copy_n(flags_.begin(), std::min(std::size(flags_), n), raw.begin());```

File: libtransmission/bitfield.cc [hunk 3]
```diff
@@ -109,89 +114,82 @@ size_t tr_bitfield::count_flags(size_t begin, size_t end) const noexcept
...
+bool tr_bitfield::ensure_bits_alloced(size_t const n)
 {
-    bool const has_all = this->has_all();
+    if (!is_size_known() || n > size()) [[unlikely]] {
+        return false;
+    }
+
+    auto const has_all = this->has_all();
 
     /* Can't use getBytesNeededSafe as n can be > SIZE_MAX - 8. */
```
ckerr: No `getBytesNeededSafe()` anymore, this comment can be removed

```suggestion
```

libtransmission/bitfield.cc
```diff
@@ -109,89 +114,82 @@ size_t tr_bitfield::count_flags(size_t begin, size_t end) const noexcept
...
 {
-    // count is zero-based, so we need to allocate nth+1 bits before setting the nth */
-    if (nth == SIZE_MAX) {
-        return false;
-    }
-
-    ensure_bits_alloced(nth + 1);
-    return true;
+    // count is zero-based, so we need to allocate nth+1 bits before setting the nth
+    return ensure_bits_alloced(nth + 1U);
```
ckerr: This drops the `if (nth == SIZE_MAX) return false;` guard, and the replacement returns success in that case. `nth + 1U` wraps to 0; in `ensure_bits_alloced`, `0 > size()` is false so the new guard passes, `getBytesNeeded(0)` is 0 so no resize happens, and the function returns `true`. A caller trusting that `true` indexes `flags_[SIZE_MAX >> 3]`.

Both current callers filter it out first, so this is latent — but the deleted guard was what made the wraparound fail closed.


```suggestion
    return nth != SIZE_MAX && ensure_bits_alloced(nth + 1U);```

File: libtransmission/bitfield.cc [hunk 5]
```diff
@@ -264,14 +270,19 @@ void tr_bitfield::set_raw(std::span<std::byte const> const raw)
     }
 
     rebuild_true_count();
+    TR_ASSERT(is_valid());
```
ckerr: Redundant — `rebuild_true_count()` calls `set_true_count()`, whose last statement is `TR_ASSERT(is_valid())`, and nothing changes state in between. Same pattern at lines 294, 419, and 441. `is_valid()` walks all of `flags_` via `count_flags()`, so each duplicate is a second full O(n) scan on assert-enabled builds, and the repetition implies something in between could invalidate the object.

```suggestion
```

libtransmission/bitfield.cc
```diff
@@ -280,16 +291,22 @@ void tr_bitfield::set_from_bools(std::span<bool const> const flags)
...
 
     set_true_count(true_count);
+    TR_ASSERT(is_valid());
+    return true;
 }
 
-void tr_bitfield::set(size_t nth, bool value)
+bool tr_bitfield::set(size_t const nth, bool const value)
 {
+    if (!is_size_known() || nth >= size()) {
```
ckerr, line 300: *(blocker)*

This silently breaks peer HAVE tracking for magnet links, and it doesn't recover once the metainfo arrives.

`tr_peerMsgsImpl::have_` is built as `tr_bitfield{torrent_in.piece_count()}`
(`peer-msgs.cc:301`), which is 0 for a magnet link, and nothing ever resizes it. `on_torrent_got_metainfo()` (`peer-msgs.cc:382`) only calls `update_active()`. So for a peer that sends `FextHaveNone` (or no bitfield at all) and then HAVEs, every `have_.set(ui32)` at `peer-msgs.cc:1412` is rejected here, forever.

Compiling `bitfield.cc` from this branch and from `main` against the same probe:

```
    main:   test(5) == true,  count() == 1
    branch: test(5) == false, count() == 0
```

The `have_none_hint_ = true` default compounds it: `has_none()` is true, so `test()` short-circuits to false regardless of `flags_`. And because the dedup guard `if (!have_.test(ui32))` never trips, `GotHave` is re-published to `swarm::got_have()` on every duplicate HAVE, inflating piece-replication counts. Those peers also show 0% via `peer->percent_done()` and are never seen as seeds.

This looks like the case the `init_size()` from description item 6 covered, removed in c5b4b23d / 620183c3f. Some way to set the size once it's known — or resizing `have_` from `on_torrent_got_metainfo()` — seems needed.

libtransmission/bitfield.cc
```diff
@@ -280,16 +291,22 @@ void tr_bitfield::set_from_bools(std::span<bool const> const flags)
...
+        return false;
+    }
+
     if (test(nth) == value) {
-        return;
+        return false;
```
ckerr, line 305: `false` here means "the bit already had that value", while `false` three lines up means "the write was rejected" — callers can't tell them apart. `set_span` has the same overlap at lines 341 and 350.

That matters for the attack this PR closes: out-of-range indices from RPC `torrent_add.files_[un]wanted` reach `tr_files_wanted::set` (`file-piece-map.cc:209`), which can't distinguish "index 2^64-2 is bogus" from "that file was already unwanted", so the RPC still reports success. The new test asserts both meanings against the same value — `EXPECT_FALSE(bf.set(0U, false))` and `EXPECT_FALSE(bf.set(bf.size()))`.

A tri-state result, or keeping `void` and exposing a separate range check, would let callers act on the rejection.


```suggestion
        return true;
```

libtransmission/bitfield.cc
```diff
@@ -311,15 +328,18 @@ void tr_bitfield::set(size_t nth, bool value)
...
 
 /* Sets bit range [begin, end) to 1 */
-void tr_bitfield::set_span(size_t begin, size_t end, bool value)
+bool tr_bitfield::set_span(size_t const begin, size_t end, bool const value)
 {
     // bounds check
     end = std::min(end, bit_count_);
-    if (end == 0 || begin >= end) {
-        return;
+    if (!is_size_known() || end > size() || begin >= end) {
```
ckerr: `end > size()` can't be true — `end` was clamped to `bit_count_` on the line above. It reads as a live bounds check, so a future reader has to re-derive that the clamp covers it. `if (!is_size_known() || begin >= end)` says the same thing.

```suggestion
    if (!is_size_known() || begin >= end) {
```

libtransmission/bitfield.cc
```diff
@@ -280,16 +291,22 @@ void tr_bitfield::set_from_bools(std::span<bool const> const flags)
     }
 
     set_true_count(true_count);
+    TR_ASSERT(is_valid());
```
ckerr:
```suggestion
```

libtransmission/bitfield.cc
```diff
@@ -372,6 +392,9 @@ void tr_bitfield::set_span(size_t begin, size_t end, bool value)
 
         decrement_true_count(old_count);
     }
+
+    TR_ASSERT(is_valid());
```
ckerr:
```suggestion
```

libtransmission/bitfield.cc
```diff
@@ -385,13 +408,15 @@ tr_bitfield& tr_bitfield::operator|=(tr_bitfield const& that)
...
 
+    bit_count_ = std::max(bit_count_, that.bit_count_);
     flags_.resize(std::max(std::size(flags_), std::size(that.flags_)));
 
     for (size_t i = 0, n = std::size(that.flags_); i < n; ++i) {
         flags_[i] |= that.flags_[i];
     }
 
     rebuild_true_count();
+    TR_ASSERT(is_valid());
```
ckerr:
```suggestion
```

libtransmission/bitfield.cc
```diff
@@ -413,6 +438,7 @@ tr_bitfield& tr_bitfield::operator&=(tr_bitfield const& that)
...
```
ckerr:
```suggestion
```

libtransmission/bitfield.cc
```diff
@@ -385,13 +408,15 @@ tr_bitfield& tr_bitfield::operator|=(tr_bitfield const& that)
         return *this;
     }
 
+    bit_count_ = std::max(bit_count_, that.bit_count_);
```
ckerr, line 411: _(blocker)_

`operator&=` never assigns `bit_count_` at all, so the `*this = that` shortcut at the top shrinks it: a 64-bit field with `set_has_all()` `&=` an 8-bit field comes out `size() == 8`.

Pre-existing, but on-topic here: this PR gives `operator|=` a size invariant a few lines up, and `bitwiseOperationsWithDifferentSizes` asserts
`left.size() == 12` after `left &= right` — a contract the general path meets only by accident, since it never assigns `bit_count_`, and that both shortcuts break. Worth settling what width each operator produces while the invariant is being established.

Whatever `|=` settles on, `&=` should mirror.

libtransmission/bitfield.h
```diff
@@ -57,19 +57,19 @@ class tr_bitfield
     {
         set_span(begin, end, false);
     }
-    void set_from_bools(std::span<bool const> flags);
+    bool set_from_bools(std::span<bool const> flags);
```
ckerr:
```suggestion
    [[nodisard]] bool set_from_bools(std::span<bool const> flags);
```

libtransmission/bitfield.h
```diff
@@ -57,19 +57,19 @@ class tr_bitfield
...
 
     // "raw" here is in BEP0003 format: "The first byte of the bitfield
     // corresponds to indices 0 - 7 from high bit to low bit, respectively.
     // The next one 8-15, etc. Spare bits at the end are set to zero."
-    void set_raw(std::span<std::byte const> raw);
+    bool set_raw(std::span<std::byte const> raw);
```
ckerr, line 65:
```suggestion
    [[nodiscard]] bool set_raw(std::span<std::byte const> raw);
```

libtransmission/bitfield.h
```diff
@@ -57,19 +57,19 @@ class tr_bitfield
...
     [[nodiscard]] std::vector<std::byte> raw() const;
 
     template<typename R>
         requires(!requires(R const& range) { std::span<std::byte const>{ range }; })
-    void set_raw(R const& raw)
+    bool set_raw(R const& raw)
```
ckerr, line 70:
```suggestion
    [[nodiscard]] bool set_raw(R const& raw)
```

tests/libtransmission/bitfield-test.cc
```diff
@@ -144,6 +165,104 @@ TEST(Bitfield, setRaw)
...
+    EXPECT_TRUE(bf.set(0U, false));
+    EXPECT_TRUE(bf.has_none());
+
+    EXPECT_FALSE(bf.set_span(0U, 0U));
+    EXPECT_FALSE(bf.set_span(5U, std::numeric_limits<size_t>::max()));
+    EXPECT_FALSE(bf.set_span(4U, 2U));
+    EXPECT_TRUE(bf.is_valid());
+}
+
+TEST(Bitfield, supportsLargestSize)
```
ckerr:
```suggestion
TEST(Bitfield, rejectsOutOfBoundsAtLargestSize)
```

libtransmission/resume.cc
```diff
@@ -481,14 +481,14 @@ tr_resume::fields_t load_filenames(tr_variant::Map const& map, tr_torrent* tor)
```
ckerr, line 481: `set_raw()`'s new `false` return is ignored when loading resume data, so a size-mismatched bitfield silently becomes "have none" without setting `err`, leaving the torrent at 0% and not queued for verification.

`load_progress` builds `blocks = tr_bitfield{tor->block_count()}` and calls `blocks.set_raw(*raw)` at line 654 (and via `raw_to_bitfield` at line 502 for both the `checked` pieces and `blocks`). `set_raw` now returns false whenever `std::size(raw) > getBytesNeeded(bit_count_)` (bitfield.cc:255) and leaves the field untouched — verified: a 3-byte raw into a 10-bit field returns false and leaves `count() == 0`. Because the result is discarded, `err` stays `nullptr`, so the `"Torrent needs to be verified"` branch at line 661 is skipped and `helper.load_blocks(blocks)` is called with an all-none bitfield.

A resume file whose stored bitfield is wider than the current block/piece count (stale file, changed block size, truncated metainfo) therefore reports 0% complete and re-downloads without ever verifying the data on disk; on origin/main the bits were loaded.

ckerr (changes requested):
This is a good start. This PR exposes some issues in bitfield's handling of different widths that I think we need to resolve.

I don't have any strong opinions on how to fix, ie if you have spent a lot of time thinking about it and know what you want, go for it.

libtransmission/bitfield.cc
tearfur: This is what #218 is for. Do you want me to subsume #218 into this PR?

libtransmission/bitfield.cc
tearfur: I'm confused by this comment, mainly by the fact that your words are suggesting a tri-state or a separate method for checking, but the code suggestion simply changes from `false` to `true`.

I'll respond to what I can understand.

So you're saying you want RPC to return error codes for invalid file indices? I agree that would be nice, but it's a big-enough change to warrant a separate PR.

OT1H, being given invalid file indices doesn't stop the torrent itself from being added to the session, so returning an error code here seems overkill. We either remove the torrent from the session and return an RPC error code if any of the file indices turned out to be invalid; or we implement something like RPC "warning codes" that are sometimes returned with an RPC success, so that we can indicate "the torrent was added, but we could not use some of the file indices".

OT2H, `tr_files_wanted::set` is called in `tr_torrent::init`, which is within `tr_torrentNew`. Whether we want error codes or warning codes, we need to refactor `tr_torrentNew` to support that.

---

I'll explain why this line is returning `false` now.

Later in the PR series, #222 uses the return value of this method to decide whether to run code that follows a value change in `files_wanted`, like emitting a signal to the wishlist and whatnot. So if the bit remains unchanged, this method returns `false` so the caller knows not to emit the signal.

libtransmission/bitfield.h
tearfur: #223

libtransmission/bitfield.h
tearfur: #223

libtransmission/resume.cc
tearfur: #223

libtransmission/bitfield.cc
tearfur: We never intersect or union 2 bitfields of different sizes in real use. I did not put much thought into this because anything I decide on will be arbitrary. If a real use-case ever comes up, we'll likely have to change the design of these bitwise operators anyway.

In fact, `operator&=` is not used anywhere. Let's just remove it. a85417e83cde11e1b7129831359121faa54fbdcc

tests/libtransmission/bitfield-test.cc
tearfur: ded2855b68fda67709121357693d6380a453dec5

libtransmission/bitfield.h
tearfur: 2241955be49225b3205af4c3283a2e44d90bb993

libtransmission/bitfield.cc
tearfur: 63560b3f3a4186eb956917dc59d7cac317bc02d8

libtransmission/bitfield.cc
tearfur: 63560b3f3a4186eb956917dc59d7cac317bc02d8

libtransmission/bitfield.cc
tearfur: 63560b3f3a4186eb956917dc59d7cac317bc02d8

libtransmission/bitfield.cc
tearfur: 63560b3f3a4186eb956917dc59d7cac317bc02d8

libtransmission/bitfield.cc
tearfur: 63560b3f3a4186eb956917dc59d7cac317bc02d8

libtransmission/bitfield.cc
tearfur: 6f80eefb98a1391028d99723bb3deefb6e2973f9

libtransmission/bitfield.cc
tearfur: 0a40aa013606f04268755e6c232983e311ee499a

libtransmission/bitfield.cc
tearfur: ce04369e84217d5a9e94a2c89cc38b2453f85817

libtransmission/bitfield.cc
tearfur: 87eee0d8d9779f4ced2f4eb54a25d3c943fd71b7

libtransmission/bitfield.cc
tearfur: 799bfd3d70dc81d012dc15e1c0069c5aa007985f

libtransmission/bitfield.cc
tearfur: I say we keep this part as-is for now, and if you feel strongly about adding RPC errors/warnings, create an issue or PR for it.

libtransmission/bitfield.cc
ckerr: No, I just hadn't read ahead. My time is limited for a couple days and so I just read the first PR in the stack :roll_eyes: 

215 -> 218 makes perfect sense and takes care of this.

libtransmission/bitfield.cc
ckerr: Oh nice catch on `operator&=`, no code --> no bugs :+1: 

Agree that this doesn't come up in practice. In my first read I thought peer-mgr.cc:1525 had it but I was wrong.

ckerr (approved):

libtransmission/bitfield.cc
ckerr: OH I'm an idiot, I did read and review 218, I'm just distracted. To rephrase, I didn't see the big picture of how they fit together.

ckerr: Actually that's a slight dodge. I was letting Claude do too much of the work.

I'll try to refrain from Vibe Reviewing. :roll_eyes:

tearfur: I'm witnessing a self-reflection/loathing episode here 😆

ckerr: Y'know that's only half wrong. I've been a professional dev for like ~30 years and adjusting to AI in 2026 is by far the largest shift I've seen in that time.

AI-assisted review is *super* helpful when it's right. But it's wrong a lot and still needs a human. So I like reviews like "I read it and the AI read it and then I compared notes", but not "I told the agent /review PR 9091 and then pasted the output".

The other thing is that so many PRs seem to be AI-only, Happily we haven't seen that here so much but it's a daily occurrence in the other two projects I spend the most time on, Electron and Transmission. If I devote review time as a human to review to every AI-only PR, I'll never get anything else done. So AI PRs *do* get a "/review PR 9091" + paste and I don't give human review time until AI is happy. :sweat_smile:

tearfur: AI has been great for helping me proofread my hand-written code. Not so much at work though, higher ups have been pushing for higher output volume using AI and it's suffocating.

ckerr: For my own work, it's been a mixed bag, but mostly positive. Some of my tasks were mostly mechanical but  too complicated to automate. It's perfect for AI since the tasks aren't creative or interesting.

:+1: on the expectations of higher output. That seems to be the new norm everywhere

PR #216, opened 2026-08-13
fix: disconnect peer if we got oversized bitfield

tearfur:
## Description

Notes: Reject peer `bitfield` messages that are too big and disconnect the peer.

## Checklist

### DIFF
--- a/libtransmission/peer-msgs.cc
+++ b/libtransmission/peer-msgs.cc
@@ -1422,7 +1422,7 @@ ReadResult tr_peerMsgsImpl::process_peer_message(uint8_t id, MessageReader& payl
 
     case BtPeerMsgs::Bitfield:
         logtrace(this, "got a bitfield");
-        have_ = tr_bitfield{ tor_.has_metainfo() ? tor_.piece_count() : std::size(payload) * 8 };
+        have_.init_size(std::size(payload) * 8);
         have_.set_raw(payload);
         peer_info->set_seed(is_seed());
         publish(tr_peer_event::GotBitfield(&have_));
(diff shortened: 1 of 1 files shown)


ckerr (commented):
Main finding: the new guard can't fire for the case the title describes

Trace the two branches at `peer-msgs.cc:1423`, given that `is_message_length_correct()` has already run at line 1354:

* Metainfo known — `peer-msgs.cc:750` requires `len == 1 + ceil(piece_count/8)` exactly, so an oversized bitfield is already rejected with `EMSGSIZE` before the `switch. have_.bit_count_ is piece_count`, `init_size` no-ops, and `set_raw`'s `std::size(raw) > getBytesNeeded(bit_count_)` test compares equal values. Cannot fail.

* Metainfo unknown — the length check waves everything through, and `init_size(std::size(payload) * 8)` then sizes `have_` to fit whatever arrived, so `getBytesNeeded()` equals the payload size by construction. Cannot fail either.

That second branch is the one that matters: `can_read_impl()` puts no cap on the 4-byte length prefix, so on a magnet link a peer can still declare a multi-gigabyte "bitfield", have us buffer it, and have us copy it into `flags_` — this PR sizes the bitfield to the attack rather than rejecting it. If bounding the magnet case is the goal, the check has to be against an independent plausibility bound (a max piece count, or the metadata size once known), not against the bitfield's own just-assigned size.

The one path that does fire is mislabeled.

An empty payload (`len == 1`) on a magnet: `is_message_length_correct` passes (`!has_metainfo()`), `init_size(0)` no-ops on the `bit_count == 0U` guard, and `set_raw` returns `false` via `!is_size_known()`. We then log "bitfield too big" for a bitfield that is empty. The only other reachable case is a second bitfield larger than the first — also not really "too big" in the sense meant.

tearfur:
Yeah, I forgot about `is_message_length_correct()` completely...

I'm going to close this PR as it does not make sense.

PR #218, opened 2026-08-13
fix: process `HAVE` messages after receiving metainfo for magnet link

tearfur:
## Description

This PR does 2 things:
- Set the size of the `tr_peerMsgImpl::have_` bitfields after receiving metainfo for a magnet link.
- If the peer sent a `BITFIELD` message that is not the right size before the client received metainfo, disconnect the peer as soon as the client received the metainfo.

No release notes needed because the loose nature of `tr_bitfield` before #215 allowed the code to work anyway (intentional? fluke? I don't know).

---

`tr_peerMsgsImpl::have_` of peers that were connected before the client received the metainfo for a magnet link was constructed with size 0 `tr_bitfield{ 0 }`. In this state, `have_.set(piece, true)` invocations is a no-op.

This is fine before the client receives the metainfo, because the client had no way to determine if the `HAVE` message carried a valid piece index, so ignoring the operations was the only thing we could do.

However, after receiving the metainfo, the current code does not set the bitfield size accordingly, so `HAVE` messages are now silently ignored.

https://github.com/retransmission/retransmission/blob/541ea54ae12c6f2d278a6f91bc9c837288199447/libtransmission/peer-msgs.cc#L1412

## Checklist

### DIFF
--- a/libtransmission/bitfield.cc
+++ b/libtransmission/bitfield.cc
@@ -214,6 +214,57 @@ void tr_bitfield::decrement_true_count(size_t dec) noexcept
     set_true_count(true_count_ - dec);
 }
 
+void tr_bitfield::unset_excess_bits() noexcept
+{
+    if (std::empty(flags_) || std::size(flags_) != tr_bytes_needed(bit_count_)) {
+        return;
+    }
+
+    auto const excess_bit_count = calc_excess_bits(bit_count_);
+    TR_ASSERT(excess_bit_count <= 7U);
+    flags_.back() &= std::byte{ 0xff } << excess_bit_count;
+}
+
+bool tr_bitfield::init_size(size_t const bit_count) noexcept
+{
+    if (bit_count == 0U || bit_count_ != 0U) {
+        return false;
+    }
+
+    bit_count_ = bit_count;
+
+    if (have_all_hint_) {
+        set_has_all(); // update true_count_
+    }
+
+    TR_ASSERT(is_valid());
+    return true;
+}
+
+bool tr_bitfield::shrink_to(size_t const bit_count) noexcept
+{
+    if (bit_count == 0U || !is_size_known() || bit_count > size()) {
+        return false;
+    }
+
+    if (bit_count == size()) {
+        return true;
+    }
+
+    bit_count_ = bit_count;
+
+    if (have_all_hint_) {
+        set_has_all(); // update true_count_
+    } else if (!has_none()) {
+        flags_.resize(std::min(std::size(flags_), tr_bytes_needed(bit_count_)));
+        unset_excess_bits();
+        rebuild_true_count();
+    }
+
+    TR_ASSERT(is_valid());
+    return true;
+}
+
 // ---
 
 tr_bitfield::tr_bitfield(size_t bit_count)
@@ -255,15 +306,7 @@ bool tr_bitfield::set_raw(std::span<std::byte const> const raw)
     flags_.assign(raw.begin(), raw.end());
 
     // ensure any excess bits at the end of the array are set to '0'.
-    if (raw.size() == tr_bytes_needed(bit_count_)) {
-        auto const excess_bit_count = (raw.size() * 8) - bit_count_;
-
-        TR_ASSERT(excess_bit_count <= 7);
-
-        if (excess_bit_count != 0) {
-            flags_.back() &= std::byte{ 0xff } << excess_bit_count;
-        }
-    }
+    unset_excess_bits();
 
     rebuild_true_count();
     // rebuild_true_count() already asserts is_valid(), so we don't need it here
@@ -316,16 +359,14 @@ bool tr_bitfield::set(size_t const nth, bool const value)
 #endif
 
     if (value) {
-        ++true_count_;
+        increment_true_count(1);
         TR_ASSERT(old_byte_pop + 1 == new_byte_pop);
     } else {
-        --true_count_;
+        decrement_true_count(1);
         TR_ASSERT(new_byte_pop + 1 == old_byte_pop);
     }
-    have_all_hint_ = true_count_ == bit_count_;
-    have_none_hint_ = true_count_ == 0;
 
-    TR_ASSERT(is_valid());
+    // (de|in)crement_true_count() already asserts is_valid(), so we don't need it here
     return true;
 }
 
--- a/libtransmission/bitfield.cc
+++ b/libtransmission/bitfield.cc
@@ -246,6 +246,8 @@ bool tr_bitfield::set_size(size_t const bit_count) noexcept
         flags_.resize(std::min(std::size(flags_), tr_bytes_needed(bit_count_)));
         unset_excess_bits();
         rebuild_true_count();
+    } else {
+        TR_ASSERT(std::empty(flags_));
     }
 
     TR_ASSERT(is_valid());
--- a/libtransmission/bitfield.cc
+++ b/libtransmission/bitfield.cc
@@ -11,16 +11,12 @@
 
 #include "libtransmission/bitfield.h"
 #include "libtransmission/tr-assert.h" // TR_ASSERT, TR_ENABLE_ASSERTS
+#include "libtransmission/utils.h"
 
 // ---
 
 namespace
 {
-
-[[nodiscard]] constexpr size_t getBytesNeeded(size_t const bit_count) noexcept
-{
-    return (bit_count >> 3) + ((bit_count & 7) != 0 ? 1 : 0);
-}
 
 [[nodiscard]] constexpr uint8_t calc_excess_bits(size_t const bit_count) noexcept
 {
@@ -34,7 +30,7 @@ namespace
 void setAllTrue(std::span<std::byte> bytes, size_t const bit_count)
 {
     static auto constexpr Val = std::byte{ 0xFF };
-    auto const n = getBytesNeeded(bit_count);
+    auto const n = tr_bytes_needed(bit_count);
 
     TR_ASSERT(bytes.size() >= n);
     if (n <= 0U || bytes.size() < n) {
@@ -134,7 +130,7 @@ bool tr_bitfield::is_valid() const
         return std::empty(flags_) && true_count_ == 0U && have_all_hint_ != have_none_hint_;
     }
 
-    auto const bytes_needed = getBytesNeeded(bit_count_);
+    auto const bytes_needed = tr_bytes_needed(bit_count_);
     auto const flags_size = std::size(flags_);
     auto const excess_bits_mask = ~(std::byte{ 0xFFU } << calc_excess_bits(bit_count_));
     return true_count_ <= bit_count_ &&
@@ -144,7 +140,7 @@ bool tr_bitfield::is_valid() const
 
 std::vector<std::byte> tr_bitfield::raw() const
 {
-    auto const n = getBytesNeeded(bit_count_);
+    auto const n = tr_bytes_needed(bit_count_);
     auto raw = std::vector<std::byte>(n);
 
     if (has_all()) {
@@ -168,7 +164,7 @@ bool tr_bitfield::ensure_bits_alloced(size_t const n)
     }
 
     auto const has_all = this->has_all();
-    auto const bytes_needed = getBytesNeeded(has_all ? std::max(n, true_count_) : n);
+    auto const bytes_needed = tr_bytes_needed(has_all ? std::max(n, true_count_) : n);
 
     if (std::size(flags_) < bytes_needed) {
         flags_.resize(bytes_needed);
@@ -252,14 +248,14 @@ bool tr_bitfield::set_raw(std::span<std::byte const> const raw)
         return false;
     }
 
-    if (auto const bytes_needed = getBytesNeeded(bit_count_); std::size(raw) > bytes_needed) {
+    if (auto const bytes_needed = tr_bytes_needed(bit_count_); std::size(raw) > bytes_needed) {
         return false;
     }
 
     flags_.assign(raw.begin(), raw.end());
 
     // ensure any excess bits at the end of the array are set to '0'.
-    if (raw.size() == getBytesNeeded(bit_count_)) {
+    if (raw.size() == tr_bytes_needed(bit_count_)) {
         auto const excess_bit_count = (raw.size() * 8) - bit_count_;
 
         TR_ASSERT(excess_bit_count <= 7);
@@ -280,7 +276,7 @@ bool tr_bitfield::set_from_bools(std::span<bool const> const flags)
         return false;
     }
 
-    flags_.assign(getBytesNeeded(flags.size()), {});
+    flags_.assign(tr_bytes_needed(flags.size()), {});
 
     size_t true_count = 0;
     for (size_t i = 0; i < flags.size(); ++i) {
--- a/libtransmission/bitfield.cc
+++ b/libtransmission/bitfield.cc
@@ -225,29 +225,15 @@ void tr_bitfield::unset_excess_bits() noexcept
     flags_.back() &= std::byte{ 0xff } << excess_bit_count;
 }
 
-bool tr_bitfield::init_size(size_t const bit_count) noexcept
+bool tr_bitfield::set_size(size_t const bit_count) noexcept
 {
-    if (bit_count == 0U || bit_count_ != 0U) {
+    // 0 is reserved for "unknown size", so it is never a valid destination.
+    // Once the size is known it can only narrow.
+    if (bit_count == 0U || (is_size_known() && bit_count > bit_count_)) {
         return false;
     }
 
-    bit_count_ = bit_count;
-
-    if (have_all_hint_) {
-        set_has_all(); // update true_count_
-    }
-
-    TR_ASSERT(is_valid());
-    return true;
-}
-
-bool tr_bitfield::shrink_to(size_t const bit_count) noexcept
-{
-    if (bit_count == 0U || !is_size_known() || bit_count > size()) {
-        return false;
-    }
-
-    if (bit_count == size()) {
+    if (bit_count == bit_count_) {
         return true;
     }
 
--- a/libtransmission/bitfield.cc
+++ b/libtransmission/bitfield.cc
@@ -133,7 +133,8 @@ bool tr_bitfield::is_valid() const
     auto const bytes_needed = tr_bytes_needed(bit_count_);
     auto const flags_size = std::size(flags_);
     auto const excess_bits_mask = ~(std::byte{ 0xFFU } << calc_excess_bits(bit_count_));
-    return true_count_ <= bit_count_ &&
+    return true_count_ <= bit_count_ && have_all_hint_ == (true_count_ == bit_count_) &&
+        have_none_hint_ == (true_count_ == 0U) &&
         (flags_size < bytes_needed || (flags_size == bytes_needed && (flags_.back() & excess_bits_mask) == std::byte{})) &&
         (std::empty(flags_) || true_count_ == count_flags());
 }
--- a/libtransmission/peer-msgs.cc
+++ b/libtransmission/peer-msgs.cc
@@ -414,6 +414,8 @@ public:
             [[maybe_unused]] auto const shrink_to_res = have_.shrink_to(tor_.piece_count());
             TR_ASSERT(shrink_to_res);
         }
+
+        peer_info->set_seed(is_seed());
     }
 
     void cancel_block_request(tr_block_index_t block)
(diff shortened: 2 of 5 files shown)


ckerr (changes requested):
The comment at peer-msgs.cc:411 states the no-op as intended, but the consequence is that have_.size() stays at payload_bytes * 8 — never piece_count. init_size() bails on bit_count_ != 0, and nothing else shrinks it.

Concretely, `piece_count == 10`, seed sends `FF C0` before metainfo:

* `init_size(16)` -> `bit_count_ = 16;` `set_raw` computes `excess_bit_count = 16 - 16 = 0`, so spare bits are never masked (BEP-3 says drop peers that set them; the metainfo-known path masks them).
* `true_count_ = 10`, so `has_all()` = `16 == 10` = `false` -> `is_seed()` is permanently `false` for any torrent whose piece count isn't a multiple of 8, i.e. 7 in 8 torrents.
* `percent()` = `10/16` -> the peer shows 62.5% forever in the peer list.

That's a pre-existing bug, but it's the same kind of bug this PR fixes and this PR already computes everything needed to fix the other bug.

tearfur:
> The comment at peer-msgs.cc:411 states the no-op as intended, but the consequence is that have_.size() stays at payload_bytes * 8 — never piece_count. init_size() bails on bit_count_ != 0, and nothing else shrinks it.
> 
> Concretely, `piece_count == 10`, seed sends `FF C0` before metainfo:
> 
> * `init_size(16)` -> `bit_count_ = 16;` `set_raw` computes `excess_bit_count = 16 - 16 = 0`, so spare bits are never masked (BEP-3 says drop peers that set them; the metainfo-known path masks them).
> * `true_count_ = 10`, so `has_all()` = `16 == 10` = `false` -> `is_seed()` is permanently `false` for any torrent whose piece count isn't a multiple of 8, i.e. 7 in 8 torrents.
> * `percent()` = `10/16` -> the peer shows 62.5% forever in the peer list.

This should be fixed now.

> That's a pre-existing bug, but it's the same kind of bug this PR fixes and this PR already computes everything needed to fix the other bug.

Not sure what you mean here. I hope the new changes are what you are looking for.

libtransmission/peer-msgs.cc
```diff
@@ -384,6 +384,35 @@ class tr_peerMsgsImpl final
...
+        // Torrent piece count was not known before this point,
+        // and this sets the bitfield size to the piece count.
+        // - If the peer sent "have all", then this line will preserve
+        //   the "have all".
+        // - If the peer sent "bitfield", then this line is a no-op,
+        //   since the bitfield size was already initialized.
+        if (!have_.init_size(tor_.piece_count())) {
+            // If the peer sent us a bitfield msg, then have_'s size is already initialized.
+            // Check if the bitfield's size is valid.
+            auto const peer_bitfield_bytes = tr_bytes_needed(have_.size());
```
ckerr, line 394: needs to include libtransmission/utils.h at top

libtransmission/peer-msgs.cc
```diff
@@ -384,6 +384,35 @@ class tr_peerMsgsImpl final
...
+                        "peer sent bitfield with {} bytes, but a valid bitfield should take {} bytes, disconnecting",
+                        peer_bitfield_bytes,
+                        actual_bitfield_bytes));
+                disconnect_soon();
+                return;
+            }
+
+            // Now that we know the exact piece count, we need to shrink the bitfield to the correct size.
+            [[maybe_unused]] auto const shrink_to_res = have_.shrink_to(tor_.piece_count());
+            TR_ASSERT(shrink_to_res);
```
ckerr, line 408: Do we need to upate tr_peer_info.is_seed_ after calling shrink_to()?

libtransmission/peer-msgs.cc
```diff
@@ -384,6 +384,35 @@ class tr_peerMsgsImpl final
...
         update_active();
+
+        TR_ASSERT(tor_.has_metainfo());
+
+        // Torrent piece count was not known before this point,
+        // and this sets the bitfield size to the piece count.
+        // - If the peer sent "have all", then this line will preserve
+        //   the "have all".
+        // - If the peer sent "bitfield", then this line is a no-op,
+        //   since the bitfield size was already initialized.
```
ckerr: Seems like a peer sending "bitfield" can initiate a `disconnect_soon()` so, not a no-op

libtransmission/utils.h
```diff
@@ -76,6 +76,16 @@ template<typename T>
 
 // ---
 
+/**
+ * Calculates the number of bytes needed to store `bit_count` bits.
+ */
+[[nodiscard]] constexpr size_t tr_bytes_needed(size_t const bit_count) noexcept
```
ckerr, line 82: peer-msgs.cc:750 should use this too. I think.

libtransmission/peer-msgs.cc
```diff
@@ -384,6 +384,35 @@ class tr_peerMsgsImpl final
...
+            auto const peer_bitfield_bytes = tr_bytes_needed(have_.size());
+            auto const actual_bitfield_bytes = tr_bytes_needed(tor_.piece_count());
+            if (peer_bitfield_bytes != actual_bitfield_bytes) {
+                logdbg(
+                    this,
+                    fmt::format(
+                        "peer sent bitfield with {} bytes, but a valid bitfield should take {} bytes, disconnecting",
+                        peer_bitfield_bytes,
+                        actual_bitfield_bytes));
+                disconnect_soon();
```
ckerr, line 403: Calling `disconnect_soon();` here isn't consistent with how peer-mgr and peer-msgs currently work together.

Usually peer-msgs publishes an event and peer-mgr makes the executive decisions.

There is a `peer_msgs::ban()` method that also calls `disconnect_soon()` but peer-mgr is the one who calls it.

tests/libtransmission/bitfield-test.cc
```diff
@@ -449,6 +449,133 @@ TEST(Bitfield, deferredSize)
     EXPECT_FALSE(have_none.set_span(0U, 1U));
     EXPECT_FALSE(have_none.set_raw(Raw));
     EXPECT_FALSE(have_none.set_from_bools(Flags));
+
```
ckerr, line 452: (claude)

The new tests miss the PR's central transition (a partial bitfield promoted to have-all by shrink_to) and the unknown-size guard case they appear to cover.

libtransmission/bitfield.cc
```diff
@@ -218,6 +214,57 @@ void tr_bitfield::decrement_true_count(size_t dec) noexcept
...
+    } else if (!has_none()) {
+        flags_.resize(std::min(std::size(flags_), tr_bytes_needed(bit_count_)));
+        unset_excess_bits();
+        rebuild_true_count();
+    }
+
+    TR_ASSERT(is_valid());
+    return true;
+}
+
```
ckerr: These are kind of the same method, no? Could delegate to a private helper


```suggestion
// 0 is reserved for "unknown size", so it is never a valid destination.
// Once the size is known it can only narrow.
bool tr_bitfield::set_size(size_t const bit_count) noexcept   // private
{
    if (bit_count == 0U || (is_size_known() && bit_count > bit_count_)) {
        return false;
    }

    if (bit_count == bit_count_) {
        return true;
    }

    bit_count_ = bit_count;

    if (have_all_hint_) {
        set_has_all(); // update true_count_
    } else if (!has_none()) {
        flags_.resize(std::min(std::size(flags_), tr_bytes_needed(bit_count_)));
        unset_excess_bits();
        rebuild_true_count();
    }

    TR_ASSERT(is_valid());
    return true;
}

[[nodiscard]] bool tr_bitfield::init_size(size_t const bit_count) noexcept
{
    return !is_size_known() && set_size(bit_count);
}

[[nodiscard]] bool tr_bitfield::shrink_to(size_t const bit_count) noexcept
{
    return is_size_known() && set_size(bit_count);
}
```

libtransmission/peer-msgs.cc
tearfur: Funny you pointed this out. I don't like how the peer mgr error handler obfuscates error details using generic errno values, so I raised #238.

Calling `disconnect_soon()` here is consistent with #238's direction.

libtransmission/bitfield.cc
tearfur: bcd4f6e1d7829ba5002f18760b2b6e184122ed68

tests/libtransmission/bitfield-test.cc
tearfur: Ooh nice catch. I think this does it?

6f62e45c61b32ef2452a055db0cb10c4dcb9cc6a
42eeef6ff278487798f90d57f9c0c91ae4712e37

libtransmission/utils.h
tearfur: `peer-msgs.cc:750` is an empty line though..?

If you meant `is_message_length_correct`, then fc878e0663658669c7fb3ce186d43d999bbf4d62.

libtransmission/peer-msgs.cc
tearfur: ee71e5a654371ebf64b37b65c059d47c81424156

libtransmission/peer-msgs.cc
tearfur: Ooh yeah, nice catch.

330a189b6ed5a9f6bb943dc6bc9604bd8f0edd4e

libtransmission/peer-msgs.cc
tearfur: 22004ebb965dff5b6cbfa342be83833717746c24

libtransmission/bitfield.cc
ckerr: `is_valid()` doesn't look at the hints at all, which means the `TR_ASSERT(is_valid())` guarding `set_size()` can't catch the one kind of corruption `set_size()` is able to cause. For example `bit_count_=16, true_count_=4, flags_={0xF0,0x00}, have_none_hint_=true` passes `is_valid()` today, while `has_none()` returns true and `count()` returns 4.

```suggestion
    return true_count_ <= bit_count_ &&
        have_all_hint_ == (true_count_ == bit_count_) &&
        have_none_hint_ == (true_count_ == 0U) &&
        (flags_size < bytes_needed || (flags_size == bytes_needed && (flags_.back() & excess_bits_mask) == std::byte{})) &&
        (std::empty(flags_) || true_count_ == count_flags());
```

libtransmission/bitfield.cc
```diff
@@ -218,6 +214,43 @@ void tr_bitfield::decrement_true_count(size_t dec) noexcept
...
+    bit_count_ = bit_count;
+
+    if (have_all_hint_) {
+        set_has_all(); // update true_count_
+    } else if (!has_none()) {
+        flags_.resize(std::min(std::size(flags_), tr_bytes_needed(bit_count_)));
+        unset_excess_bits();
+        rebuild_true_count();
+    }
```
ckerr, line 251:
```suggestion
    } else {
        TR_ASSERT(std::empty(flags_));
    }
```

libtransmission/peer-msgs.cc
```diff
@@ -384,6 +385,37 @@ class tr_peerMsgsImpl final
...
         update_active();
+
+        TR_ASSERT(tor_.has_metainfo());
+
+        // Torrent piece count was not known before this point,
+        // and this sets the bitfield size to the piece count.
+        // - If the peer sent "have all", then this line will preserve
+        //   the "have all".
+        // - If the peer sent "bitfield", then `have_.init_size()` is no-op,
+        //   and `have_.shrink_to()` will adjust the bitfield size instead.
```
ckerr: 1. I'm not sure how much of this we actually need to document in this code comment? I'd vote to keep it only iff it's needed to understand what's going on in this function? If we keep it, we probably ought to document that it can also result in the peer being dropped.

2. Comment duplication between here and a few lines below in 398, and also between lines 391 and 413. Not a blocker.

libtransmission/peer-msgs.cc
```diff
@@ -384,6 +385,37 @@ class tr_peerMsgsImpl final
...
+        if (!have_.init_size(tor_.piece_count())) {
```
ckerr, line 393: (claude)

Worth confirming this is observable at all, because I don't think it is on the path
the commit message describes.

`tr_torrent::set_metainfo()` emits `got_metainfo_(this)` (torrent.cc:1021) — which
is what reaches this code — and then calls `on_metainfo_completed()` five statements
later (torrent.cc:1026). With the default `TR_VERIFY_ADDED_FAST`, torrent.cc:811
reduces to `!is_new_torrent_a_seed()`, which is true for a magnet with no local data,
so `tr_torrentVerify()` runs. `tr_session_thread::run()` executes inline when already
on the session thread, so `stop_now()` → `stopped_(this)` → `tr_swarm::stop()` →
`remove_all_peers()` all happen synchronously: every peer whose `have_` was just
sized is destroyed, and they reconnect after verification with the piece count
already known.

The one path where peers survive is the `on_metainfo_completed()` else-branch — a
magnet whose data is already on disk — and there the client is instantly a seed, so
"couldn't process HAVE messages ... peers ... downloading the torrent themselves"
doesn't apply.

Related: `tests/libtransmission/peer-msgs-test.cc` is still a lone
`TEST(PeerMsgs, placeholder)`, so none of the three branches here is exercised
anywhere. The two new bitfield tests cover the container, not this caller.

libtransmission/peer-msgs.cc
```diff
@@ -384,6 +385,37 @@ class tr_peerMsgsImpl final
...
+            auto const peer_bitfield_bytes = tr_bytes_needed(have_.size());
+            auto const actual_bitfield_bytes = tr_bytes_needed(tor_.piece_count());
+            if (peer_bitfield_bytes != actual_bitfield_bytes) {
+                logdbg(
+                    this,
+                    fmt::format(
+                        "peer sent bitfield with {} bytes, but a valid bitfield should take {} bytes, disconnecting",
+                        peer_bitfield_bytes,
+                        actual_bitfield_bytes));
+                disconnect_soon();
```
ckerr, line 403: Hmm not sure if this would matter in practice but we probably ought to call `set_seed()` before we return after calling `disconnect_soon();` for consistency with the call at the end of this function, eg something like

```c++
                peer_info->set_seed(false);
                disconnect_soon();
                return;
```

libtransmission/peer-msgs.cc
```diff
@@ -384,6 +385,37 @@ class tr_peerMsgsImpl final
...
+                return;
+            }
+
+            // Now that we know the exact piece count, we need to shrink the bitfield to the correct size.
+            [[maybe_unused]] auto const shrink_to_res = have_.shrink_to(tor_.piece_count());
+            TR_ASSERT(shrink_to_res);
+        }
+
+        peer_info->set_seed(is_seed());
```
ckerr: (claude)

(IDK about this one, the use case seems tortured IMO, but passing it along to see what you think)

`init_size()` preserves `have_all_hint_`, and `shrink_to()` can only make
`count == size` *more* likely — so `is_seed()` after this block is never lower than
it was before, and a `false` written here is either redundant or destructive.
`tr_peer_info` persists in `connectable_pool` across connections and the codebase
treats `is_seed_` as sticky (`merge()`: `/* keep is_seed_ as-is */`). Concretely:
peer sends FEXT_HAVE_ALL on one connection, drops, reconnects, and metainfo arrives
from a different peer before the new connection's HAVE_ALL is parsed — `have_` is
still have-none, so this line erases what we already knew.

```suggestion
        // Neither init_size() nor shrink_to() can turn "have all" off, so writing
        // `false` here could only discard what an earlier connection told us.
        if (is_seed()) {
            peer_info->set_seed();
        }
```

libtransmission/peer-msgs.cc
ckerr, line 1443:
```suggestion
        /* a peer can send the same HAVE twice, and before the metainfo arrives we
           have no way to tell whether the piece index is even in range */
        if (have_.set(ui32)) {
            peer_info->set_seed(is_seed());
            publish(tr_peer_event::GotHave(ui32));
```

ckerr, line 1456: (claude)

`set_raw()`'s return value is discarded but sometimes we need it:

A BITFIELD with a zero-length payload reaches this handler — a length prefix of 1 is
not a keepalive, and `is_message_length_correct()` short-circuits on
`!tor.has_metainfo()`. Then `std::size(payload) * 8` is 0, so `have_` is constructed
size-unknown and `set_raw()` fails silently. When the metainfo lands, `init_size()`
*succeeds* and the new byte-count check never runs — so a 1-byte bitfield for that
torrent is dropped while a 0-byte one is kept.

The resulting state is harmless (a correctly sized have-none bitfield), so this is
about the validation being consistent:

```suggestion
        have_ = tr_bitfield{ tor_.has_metainfo() ? tor_.piece_count() : std::size(payload) * 8 };
        if (!have_.set_raw(payload)) {
            publish(tr_peer_event::GotError(EMSGSIZE));
            return { ReadState::Err, {} };
        }
        peer_info->set_seed(is_seed());
```

tests/libtransmission/bitfield-test.cc
```diff
@@ -449,6 +449,144 @@ TEST(Bitfield, deferredSize)
...
+    // shrinking a partially set bitfield to a size that includes all the true bits makes it have-all
+    bf = tr_bitfield{ 20 };
+    ASSERT_TRUE(bf.set_span(0U, 15U));
+    EXPECT_EQ(15U, bf.count());
+    EXPECT_FALSE(bf.has_all());
+    EXPECT_TRUE(bf.shrink_to(15U));
+    EXPECT_EQ(15U, bf.size());
+    EXPECT_EQ(15U, bf.count());
+    EXPECT_TRUE(bf.has_all());
```
ckerr: (claude)

Two small things about the new have-all block. It's the only one in this test
without an `is_valid()` check — the siblings all end with one — so internal state
the getters can't see (stale excess bits, `flags_`/`true_count_` disagreement) goes
unchecked. And `EXPECT_TRUE(bf.has_all())` can't actually distinguish the promotion
it's named for: `has_all()` falls back to `bit_count_ > 0 && bit_count_ ==
true_count_`, which holds whether or not `set_size()` sets `have_all_hint_` and frees
the array.

Also, reusing `bf` here is a bit surprising since it's 10 bits wide at this point,
while the two blocks above each introduce a fresh local.

```suggestion
    auto promoted = tr_bitfield{ 20 };
    ASSERT_TRUE(promoted.set_span(0U, 15U));
    EXPECT_EQ(15U, promoted.count());
    EXPECT_FALSE(promoted.has_all());
    EXPECT_TRUE(promoted.shrink_to(15U));
    EXPECT_EQ(15U, promoted.size());
    EXPECT_EQ(15U, promoted.count());
    EXPECT_TRUE(promoted.has_all());
    EXPECT_EQ((std::vector<std::byte>{ std::byte{ 0xff }, std::byte{ 0xfe } }), promoted.raw());
    EXPECT_TRUE(promoted.is_valid());
```

libtransmission/peer-msgs.cc
tearfur: The reason why we disconnect the peer here is because the bitfield it sent does not make sense for this torrent. I think we shouldn't do anything based on such a bitfield, including `set_seed()`.

tests/libtransmission/bitfield-test.cc
tearfur: 354dbd98256b21b0b99afda325e81cc516cf233a

libtransmission/peer-msgs.cc
tearfur: 6887115e452d5c9c709a62772901964c529c2772

libtransmission/peer-msgs.cc
tearfur: 1e623ee18a7dd7898edd7b1ba3032c444513562c

libtransmission/bitfield.cc
tearfur: Noting that while this is nice to enforce for as low a memory profile as possible, this is not needed for correctness.

`is_valid()` enforces the `flags_.size() <= bytes_needed` and this is enough for the methods to work correctly.

74db0cf1bc5e34f6138134ec6fd3807431bdb434

libtransmission/bitfield.cc
tearfur: c3ed793bcd31121b7785d0c0ec48b909d74d2c4c

libtransmission/peer-msgs.cc
tearfur: If the resulting state is harmless, then what benefit is there to disconnecting the peer? It only serves to reduce our connectivity.

ckerr (approved):

PR #219, opened 2026-08-13
fix: OOB read/write in `tr_file_priorities::priorities_`

tearfur:
## Description

Fix a potential buffer overflow from RPC `torrent_add.priority_*`.

The way to trigger the buffer overflow is simple: Add a torrent with `N` pieces using `torrent_add` with argument `"priority_high": [N]`, and the client will try to write one byte past the end of `tr_file_priorities::priorities_`.

Notes: Hardened the `torrent-add` RPC method against malformed requests.

## Checklist

### DIFF
--- a/libtransmission/file-piece-map.cc
+++ b/libtransmission/file-piece-map.cc
@@ -140,32 +140,40 @@ tr_file_piece_map::file_offset_t tr_file_piece_map::file_offset(uint64_t const o
 
 // ---
 
-void tr_file_priorities::set(tr_file_index_t const file, tr_priority_t const new_priority)
+bool tr_file_priorities::set(tr_file_index_t const file, tr_priority_t const new_priority)
 {
+    if (file >= fpm_->file_count()) {
+        return false;
+    }
+
     if (std::empty(priorities_)) {
         if (new_priority == TR_PRI_NORMAL) {
-            return;
+            return false;
         }
 
         priorities_.assign(fpm_->file_count(), TR_PRI_NORMAL);
         priorities_.shrink_to_fit();
     }
 
-    priorities_[file] = new_priority;
+    return std::exchange(priorities_[file], new_priority) != new_priority;
 }
 
-void tr_file_priorities::set(std::span<tr_file_index_t const> const files, tr_priority_t const new_priority)
+bool tr_file_priorities::set(std::span<tr_file_index_t const> const files, tr_priority_t const new_priority)
 {
-    for (auto const file : files) {
-        set(file, new_priority);
+    if (std::ranges::any_of(files, [n_files = fpm_->file_count()](tr_file_index_t file) { return file >= n_files; })) {
+        return false;
     }
+
+    auto ret = false;
+    for (auto const file : files) {
+        ret |= set(file, new_priority);
+    }
+    return ret;
 }
 
 tr_priority_t tr_file_priorities::file_priority(tr_file_index_t const file) const
 {
-    TR_ASSERT(file < fpm_->file_count());
-
-    if (std::empty(priorities_)) {
+    if (file >= priorities_.size()) {
         return TR_PRI_NORMAL;
     }
 
--- a/libtransmission/file-piece-map.cc
+++ b/libtransmission/file-piece-map.cc
@@ -79,7 +79,7 @@ void tr_file_piece_map::reset(tr_torrent_metainfo const& tm)
     for (tr_file_index_t i = 0U; i < n; ++i) {
         file_sizes[i] = tm.file_size(i);
     }
-    reset({ tm.total_size(), tm.piece_size() }, file_sizes);
+    reset(tm.block_info(), file_sizes);
 }
 
 namespace
--- a/libtransmission/file-piece-map.cc
+++ b/libtransmission/file-piece-map.cc
@@ -6,6 +6,7 @@
 #include <algorithm>
 #include <cstdint>
 #include <iterator>
+#include <utility> // std::exchange()
 #include <vector>
 
 #include <small/set.hpp>
--- a/libtransmission/file-piece-map.h
+++ b/libtransmission/file-piece-map.h
@@ -87,8 +87,8 @@ public:
     {
     }
 
-    void set(tr_file_index_t file, tr_priority_t priority);
-    void set(std::span<tr_file_index_t const> files, tr_priority_t priority);
+    [[nodiscard]] bool set(tr_file_index_t file, tr_priority_t priority);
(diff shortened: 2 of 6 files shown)


libtransmission/file-piece-map.cc
ckerr, line 10:
```suggestion
#include <iterator>
#include <utility> // std::exchange()
#include <vector>
```

libtransmission/file-piece-map.h
```diff
@@ -87,8 +87,8 @@ class tr_file_priorities
     {
     }
 
-    void set(tr_file_index_t file, tr_priority_t priority);
-    void set(std::span<tr_file_index_t const> files, tr_priority_t priority);
+    [[nodiscard]] bool set(tr_file_index_t file, tr_priority_t priority);
+    [[nodiscard]] bool set(std::span<tr_file_index_t const> files, tr_priority_t priority);
```
ckerr, line 92:
```suggestion
    // returns true if any file's priority changed.
    [[nodiscard]] bool set(tr_file_index_t file, tr_priority_t priority);
    [[nodiscard]] bool set(std::span<tr_file_index_t const> files, tr_priority_t priority);
```

libtransmission/torrent.cc
```diff
@@ -1907,10 +1907,7 @@ tr_block_span_t tr_torrent::block_span_for_file(tr_file_index_t const file) cons
 
 void tr_torrent::set_file_priorities(std::span<tr_file_index_t const> const files, tr_priority_t priority)
 {
-    if (std::ranges::any_of(files, [this, priority](tr_file_index_t file) {
-            return priority != file_priorities_.file_priority(file);
-        })) {
-        file_priorities_.set(files, priority);
+    if (file_priorities_.set(files, priority)) {
```
ckerr, line 1912: not new to this PR, but do we need a unique_lock here? I notice `set_files_wanted()` has one 🤔

libtransmission/torrent.h
ckerr: This works the same as the span variant now, so we can replace all five lines of the impl with

```c++
    set_file_priorities(std::span{ &file, 1U }, priority);
```

tests/libtransmission/file-piece-map-test.cc
```diff
@@ -238,7 +238,7 @@ TEST_F(FilePieceMapTest, priorities)
     // Prep for the next test: set all files to normal priority
     pri = TR_PRI_NORMAL;
     for (tr_file_index_t i = 0U; i < n_files; ++i) {
-        file_priorities.set(i, pri);
+        EXPECT_EQ(i != 8U, file_priorities.set(i, pri));
```
ckerr: (claude)

8U is correct only because of the set(8U, TR_PRI_NORMAL) ~18 lines earlier. Inserting or reordering a step between them fails here with no hint that the index is the stale part. expected_file_priorities already holds the right values at this point (LOW everywhere except [8]=NORMAL, [9]=HIGH), so the derived form is equivalent today and self-correcting afterwards.


```suggestion
        EXPECT_EQ(expected_file_priorities[i] != pri, file_priorities.set(i, pri));
```

tests/libtransmission/file-piece-map-test.cc
```diff
@@ -271,19 +271,44 @@ TEST_F(FilePieceMapTest, priorities)
...
+    auto const invalid_files = std::to_array<tr_file_index_t>({ 1U, n_files });
+    EXPECT_FALSE(file_priorities.set(invalid_files, TR_PRI_LOW));
+    EXPECT_EQ(TR_PRI_NORMAL, file_priorities.file_priority(1U));
+
+    constexpr auto Files = std::to_array<tr_file_index_t>({ 1U, 2U });
+    EXPECT_TRUE(file_priorities.set(Files, TR_PRI_LOW));
+    EXPECT_FALSE(file_priorities.set(Files, TR_PRI_LOW));
+    EXPECT_EQ(TR_PRI_LOW, file_priorities.file_priority(1U));
+    EXPECT_EQ(TR_PRI_LOW, file_priorities.file_priority(2U));
+}
```
ckerr, line 318: (claude)

The test builds a 17-file map and probes n_files and {1U, n_files}. Its set(n_files, TR_PRI_LOW) at line 297 runs only after line 295 allocated 17 entries, so it exercises a one-past-the-end write into a non-empty buffer that ASan may or may not catch.

It never covers file_count() == 0 — where pre-PR assign(0, TR_PRI_NORMAL) left the vector empty and priorities_[file] wrote into a zero-length allocation. That's the magnet shape, i.e. the variant that reached users. Reverting just the scalar guard at line 145 leaves the whole suite green.

Sketch to anchor on line 289 (I haven't compiled this — a default-constructed tr_block_info may need adjusting, and reset() should be checked against an empty file-size span):

```c++
    // A torrent with no files yet -- a magnet before its metadata arrives --
    // must reject every index rather than index an empty vector.
    auto const empty_fpm = tr_file_piece_map{ tr_block_info{}, std::span<uint64_t const>{} };
    auto empty_priorities = tr_file_priorities(&empty_fpm);
    EXPECT_FALSE(empty_priorities.set(0U, TR_PRI_HIGH));
    EXPECT_EQ(TR_PRI_NORMAL, empty_priorities.file_priority(0U));

    auto const fpm = tr_file_piece_map{ block_info_, FileSizes };
```

tearfur: 57a347818dd7fc73747c71de3cf280ed969997b8

tests/libtransmission/file-piece-map-test.cc
tearfur: 3d5eceb002b52e2e317b38cdc800b7e54415cf20

libtransmission/torrent.h
tearfur: 550fbb94b43dda297e374156f41f71bf5e52620f

libtransmission/torrent.cc
tearfur: I think so. cbf73e5abdfaab9f7ea64e90eb9458768911f280

The reason why `set_files_wanted()` has one is because the method is exposed in the C API. The same goes for `set_file_priorities()`.

libtransmission/file-piece-map.h
tearfur: d987f1527f96e9804176f396188627d5d2fee240

libtransmission/file-piece-map.cc
tearfur: f5f99dd833e0881858e139c401043297323c69fd

ckerr (approved):

PR #220, opened 2026-08-13
Extract small torrent cell from xib.

lolgear:
To increase flexibility when editing the small torrent cell.
Based on #8547

Extracting the cell from the XIB unlocks:
2. Implementing a proper `ProgressView` using `CALayer` to improve performance by offloading computations to the GPU.

### Results

This PR changes the behavior related to multiple invocations of `awakeFromNib` in `TorrentTableView` according to the [Apple documentation](https://apple.com).

> This method is usually called by the delegate in [tableView(_:viewFor:row:)](https://apple.com), but it can also be overridden to provide custom views for the identifier. Note that [awakeFromNib()](https://apple.com) is called each time this method is called, meaning that `awakeFromNib()` will trigger for the owner bundle repeatedly if the owner is passed down during instantiation.

### DIFF
--- a/macosx/SmallTorrentCell.mm
+++ b/macosx/SmallTorrentCell.mm
@@ -14,6 +14,106 @@
 
 @implementation SmallTorrentCell
 
+// Layout
+- (void)configurePriorities
+{
+    auto groupIndicatorView = self.fGroupIndicatorView;
+    auto iconView = self.fIconView;
+    auto torrentTitleField = self.fTorrentTitleField;
+    auto torrentPriorityView = self.fTorrentPriorityView;
+    auto torrentStatusField = self.fTorrentStatusField;
+
+    // left items
+    for (NSView* leftView in @[ groupIndicatorView, iconView ]) {
+        [leftView setContentHuggingPriority:NSLayoutPriorityDefaultLow + 1 forOrientation:NSLayoutConstraintOrientationHorizontal];
+        [leftView setContentHuggingPriority:NSLayoutPriorityDefaultLow + 1 forOrientation:NSLayoutConstraintOrientationVertical];
+    }
+
+    /// middle items
+    [torrentTitleField setContentHuggingPriority:NSLayoutPriorityRequired forOrientation:NSLayoutConstraintOrientationHorizontal];
+    [torrentTitleField setContentHuggingPriority:NSLayoutPriorityDefaultHigh forOrientation:NSLayoutConstraintOrientationVertical];
+    [torrentPriorityView setContentHuggingPriority:NSLayoutPriorityRequired forOrientation:NSLayoutConstraintOrientationHorizontal];
+    [torrentPriorityView setContentHuggingPriority:NSLayoutPriorityDefaultLow + 1 forOrientation:NSLayoutConstraintOrientationVertical];
+
+    [torrentStatusField setContentHuggingPriority:NSLayoutPriorityDefaultLow + 1 forOrientation:NSLayoutConstraintOrientationHorizontal];
+    [torrentStatusField setContentHuggingPriority:NSLayoutPriorityDefaultHigh forOrientation:NSLayoutConstraintOrientationVertical];
+    [torrentStatusField setContentCompressionResistancePriority:NSLayoutPriorityRequired
+                                                 forOrientation:NSLayoutConstraintOrientationHorizontal];
+}
+
+- (void)setupConstraints
+{
+    auto groupIndicatorView = self.fGroupIndicatorView;
+    auto iconView = self.fIconView;
+    auto actionButton = self.fActionButton;
+    auto stackView = self.fStackView;
+    auto torrentStatusField = self.fTorrentStatusField;
+    auto torrentProgressBarView = self.fTorrentProgressBarView;
+    auto controlButton = self.fControlButton;
+    auto revealButton = self.fRevealButton;
+
+    for (NSView* view in @[
+             groupIndicatorView,
+             iconView,
+             actionButton,
+             torrentProgressBarView,
+             stackView,
+             torrentStatusField,
+             controlButton,
+             revealButton,
+         ]) {
+        [self addSubview:view];
+    }
+
+    [NSLayoutConstraint activateConstraints:@[
+        // groupIndicatorView
+        [groupIndicatorView.leadingAnchor constraintEqualToAnchor:self.leadingAnchor],
+        [groupIndicatorView.centerYAnchor constraintEqualToAnchor:self.centerYAnchor],
+        [groupIndicatorView.widthAnchor constraintEqualToConstant:6],
+        [groupIndicatorView.heightAnchor constraintEqualToConstant:6],
+
+        // iconView
+        [iconView.leadingAnchor constraintEqualToAnchor:groupIndicatorView.trailingAnchor constant:8],
+        [iconView.centerYAnchor constraintEqualToAnchor:self.centerYAnchor],
+        [iconView.widthAnchor constraintEqualToConstant:16],
+        [iconView.heightAnchor constraintEqualToConstant:16],
+
+        // actionButton
+        [actionButton.centerXAnchor constraintEqualToAnchor:iconView.centerXAnchor],
+        [actionButton.centerYAnchor constraintEqualToAnchor:iconView.centerYAnchor],
+        [actionButton.widthAnchor constraintEqualToConstant:16],
+        [actionButton.heightAnchor constraintEqualToConstant:16],
+
+        // torrentProgressBarView
+        [torrentProgressBarView.leadingAnchor constraintEqualToAnchor:iconView.trailingAnchor constant:15],
+        [torrentProgressBarView.trailingAnchor constraintEqualToAnchor:self.trailingAnchor constant:-5],
+        [torrentProgressBarView.centerYAnchor constraintEqualToAnchor:self.centerYAnchor],
+        [torrentProgressBarView.heightAnchor constraintEqualToConstant:18],
+
+        // stackView
+        [stackView.leadingAnchor constraintEqualToAnchor:torrentProgressBarView.leadingAnchor],
+        [stackView.topAnchor constraintEqualToAnchor:torrentProgressBarView.topAnchor],
+        [stackView.bottomAnchor constraintEqualToAnchor:torrentProgressBarView.bottomAnchor],
+
+        // torrentStatusField
+        [torrentStatusField.leadingAnchor constraintEqualToAnchor:stackView.trailingAnchor],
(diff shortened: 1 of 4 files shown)


macosx/TorrentTableView.mm
```diff
@@ -205,7 +205,16 @@ - (NSView*)outlineView:(NSOutlineView*)outlineView viewForTableColumn:(NSTableCo
 
         TorrentCell* torrentCell;
         if (minimal) {
-            torrentCell = [outlineView makeViewWithIdentifier:@"SmallTorrentCell" owner:self];
+            // TODO(lolgear): Remove small torrent cell from xib.
+            torrentCell = [outlineView makeViewWithIdentifier:@"NewSmallTorrentCell" owner:self];
```
ckerr, line 209: According to Claude, 

@"NewSmallTorrentCell" isn't in the XIB (the nib's identifier is SmallTorrentCell), so makeViewWithIdentifier: returns nil and this fallback runs for every compact row. Four consequences:

- The new SmallTorrentCell -setupConstraints is dead code. initWithFrame: dispatches to TorrentCell's version — the entire point of the PR never executes.
- Compact rows get the regular layout. TorrentCell -setupConstraints chains title-stack → progress field → 14pt bar → status field with bottom == self.bottom - 1, requiring ~62pt (kRowHeightRegular), plus a 36pt icon. Compact rows are kRowHeightSmall = 22.0. Auto Layout will break constraints and log on every row.
- Hover breaks in compact view. mouseEntered:/mouseExited:/updateTrackingAreas live only on SmallTorrentCell; a TorrentCell installs no tracking area, so hoverEventBeganForView: is never called for the row and the control/reveal buttons never appear.
- Even if it were called, macosx/TorrentTableView.mm:597 and macosx/TorrentTableView.mm:638 gate on isKindOfClass:[SmallTorrentCell class]; the latter would take the update = NO path and leave hover state stuck.


Fix is one word ([[SmallTorrentCell alloc] initWithFrame:…]), but it means the compact view was not exercised — please re-verify it by hand before merge.

ckerr (changes requested):

macosx/TorrentTableView.mm
lolgear: Claude is joking.

Everything is good ( OOP resolves any issue Claude mentioned ).

According to documentation of [makeViewWithIdentifier](https://developer.apple.com/documentation/appkit/nstableview/makeview(withidentifier:owner:)).

> This method may also return a reused view with the same identifier that is no longer available on screen. If a view with the specified identifier can’t be instantiated from the nib file **or found in the reuse queue**, this method returns nil.

When you set identifier, view will be assigned to specific reuse queue ( pool ).
It is done across project in different places as `TorrentCell`, `GroupCell`, `FileNameCell`, `FilePriorityCell` etc.

lolgear:
@ckerr 
Could you check it?
It blocks fast gpu-friendly progress bar in torrent cell.
And cleanup of whole xib legacy in torrent table view. ( not whole, torrent table view is still in xib, however, a lot of it).

ckerr (commented):
- `configurePriorities` should call `super`. SmallTorrentCell.mm:18 re-types the base's `left-item` loop and `middle-item` block verbatim, then diverges: it silently drops the base's right-item block (required horizontal hugging + required compression resistance on control/reveal) and the action button's vertical hugging. That's inert today — all three have fixed 14×14/16×16 constraints in the small layout — so this isn't a bug. It is a maintenance trap: a future adjustment to the base priorities won't reach the small cell, and nothing signals that. `[super configurePriorities]` followed by the three status-field lines is ~18 lines shorter and can't drift.

- `TorrentCell.h` now publishes two subclass hooks (`configurePriorities`, `setupConstraints`) while `configureViews` stays private, so the trio is split across two visibilities for no stated reason. A `TorrentCell+Subclass.h` carrying all three would keep the public surface to outlets.

- `stackView.top/bottom` pinned to the 18pt progress bar, where the XIB used `centerY == cell.centerY` with the stack's intrinsic 15pt height; `stackView.alignment` is still never set (XIB: `centerY`). Also unported: the priority view's 12×12 pin and the stack's 4pt trailing `edgeInset`.

- `addSubview`: still lives inside `setupConstraints`, and the subview list is now duplicated across both classes — a separate overridable `-addSubviews` would fix both.

- `fTorrentPriorityViewWidthConstraint` is still read nowhere — deletable with the XIB cells.

lolgear:
> * `configurePriorities` should call `super`. SmallTorrentCell.mm:18 re-types the base's `left-item` loop and `middle-item` block verbatim, then diverges: it silently drops the base's right-item block (required horizontal hugging + required compression resistance on control/reveal) and the action button's vertical hugging. That's inert today — all three have fixed 14×14/16×16 constraints in the small layout — so this isn't a bug. It is a maintenance trap: a future adjustment to the base priorities won't reach the small cell, and nothing signals that. `[super configurePriorities]` followed by the three status-field lines is ~18 lines shorter and can't drift.

Actually, yes and no.
Keep them separated as much as possible is good.
You don't need to copy "constraints" (layout) because they are actually "NOT" the same.
They share similar views but their layout **seems** similar.

> * `TorrentCell.h` now publishes two subclass hooks (`configurePriorities`, `setupConstraints`) while `configureViews` stays private, so the trio is split across two visibilities for no stated reason. A `TorrentCell+Subclass.h` carrying all three would keep the public surface to outlets.

It's ok cause you only need to configure only priorities and layout.
Also adding new file is a lot of pain.

> * `stackView.top/bottom` pinned to the 18pt progress bar, where the XIB used `centerY == cell.centerY` with the stack's intrinsic 15pt height; `stackView.alignment` is still never set (XIB: `centerY`). Also unported: the priority view's 12×12 pin and the stack's 4pt trailing `edgeInset`.

stackView.top/bottom is nice pinned to 18pt progress bar.
Also stackView is horizontal by default. And for horizontal stackView alignment is `centerY` by default.
priorityView 12x12 constraints - yes, lost.
not sure about 4pt trailing edgeInset but lets add it.

> * `addSubview`: still lives inside `setupConstraints`, and the subview list is now duplicated across both classes — a separate overridable `-addSubviews` would fix both.

A list of views is similar but not the same.

> * `fTorrentPriorityViewWidthConstraint` is still read nowhere — deletable with the XIB cells.

Later, of course.

macosx/SmallTorrentCell.mm
```diff
@@ -14,6 +14,122 @@ @interface SmallTorrentCell ()
 
 @implementation SmallTorrentCell
 
+- (void)configureViews
+{
+    [super configureViews];
+    // also set alignment as part of priorities ( text inside a textLabel ).
```
ckerr: This comment got copied out of configurePriorities but doesn't quite make sense here, eg no priorities in the code here? Probably would remove this comment


```suggestion
```

macosx/SmallTorrentCell.mm
```diff
@@ -14,6 +14,122 @@ @interface SmallTorrentCell ()
...
+    self.fTorrentStatusField.alignment = NSTextAlignmentRight;
+
+    // just ported from XIB.
+    // not sure if required.
+    self.fStackView.edgeInsets = NSEdgeInsetsMake(0, 0, 0, 4);
+}
```
ckerr, line 22:
```suggestion
- (void)configureViews
{
    [super configureViews];
    self.fTorrentStatusField.alignment = NSTextAlignmentRight;
}
```

macosx/SmallTorrentCell.mm
```diff
@@ -14,6 +14,122 @@ @interface SmallTorrentCell ()
...
+        [torrentPriorityView.heightAnchor constraintEqualToConstant:12],
+        [torrentPriorityView.widthAnchor constraintEqualToConstant:12],
+
+        // stackView
+        [stackView.leadingAnchor constraintEqualToAnchor:torrentProgressBarView.leadingAnchor],
+        [stackView.topAnchor constraintEqualToAnchor:torrentProgressBarView.topAnchor],
+        [stackView.bottomAnchor constraintEqualToAnchor:torrentProgressBarView.bottomAnchor],
+
+        // torrentStatusField
+        [torrentStatusField.leadingAnchor constraintEqualToAnchor:stackView.trailingAnchor],
```
ckerr:
```suggestion
        [torrentStatusField.leadingAnchor constraintEqualToAnchor:stackView.trailingAnchor constant:4],```

File: macosx/SmallTorrentCell.mm [hunk 5]
```diff
@@ -14,6 +14,122 @@ @interface SmallTorrentCell ()
...
+    [torrentPriorityView setContentHuggingPriority:NSLayoutPriorityDefaultLow + 1 forOrientation:NSLayoutConstraintOrientationVertical];
+
+    [torrentStatusField setContentHuggingPriority:NSLayoutPriorityDefaultLow + 1 forOrientation:NSLayoutConstraintOrientationHorizontal];
+    [torrentStatusField setContentHuggingPriority:NSLayoutPriorityDefaultHigh forOrientation:NSLayoutConstraintOrientationVertical];
+    [torrentStatusField setContentCompressionResistancePriority:NSLayoutPriorityRequired
+                                                 forOrientation:NSLayoutConstraintOrientationHorizontal];
+}
+
+- (void)setupConstraints
+{
```
ckerr, line 52: Can we make some symbolic constants here for use below? eg

```objc++
// Leading edge inward.
static CGFloat const kGroupIndicatorSize = 6.0;
static CGFloat const kGroupIndicatorToIconSpacing = 8.0;
static CGFloat const kIconSize = 16.0;
static CGFloat const kActionButtonSize = 16.0;
static CGFloat const kIconToBarSpacing = 15.0;

// Trailing edge inward. The progress bar stops short of the cell edge; the status
// text and reveal button stop short of the bar, landing them 8pt in from the cell —
// the same offset TorrentCell gives its reveal button.
static CGFloat const kCellTrailingInset = 5.0;
static CGFloat const kContentTrailingInset = 8.0;
static CGFloat const kBarTrailingInset = kContentTrailingInset - kCellTrailingInset;

// Within the bar.
static CGFloat const kProgressBarHeight = 18.0;
static CGFloat const kPriorityIconSize = 12.0;
static CGFloat const kTitleToStatusSpacing = 4.0;
static CGFloat const kHoverButtonSize = 14.0;
static CGFloat const kHoverButtonSpacing = 3.0;
```

lolgear: Yes, but they are used exactly once, so some of them should be reviewed. Also, some are unnecessary because those views can be grouped into containers. These containers could then handle the insets/offsets instead of duplicating them across multiple views. Personally, I prefer to keep them close to the original XIB.

lolgear:
@ckerr 
Could we merge it?

ckerr (approved):

PR #222, opened 2026-08-13
perf: execute follow-ups iff files_wanted changed

tearfur:
## Description

Execute the follow-ups, callbacks and whatnot iff `files_wanted` changed, instead of executing them unconditionally as long as `tr_torrent::set_files_wanted` is called.

## Checklist

### DIFF
--- a/libtransmission/file-piece-map.cc
+++ b/libtransmission/file-piece-map.cc
@@ -212,16 +212,22 @@ tr_files_wanted::tr_files_wanted(tr_file_piece_map const* const fpm)
     wanted_.set_has_all(); // by default we want all files
 }
 
-void tr_files_wanted::set(tr_file_index_t const file, bool const wanted)
+bool tr_files_wanted::set(tr_file_index_t const file, bool const wanted)
 {
-    wanted_.set(file, wanted);
+    return wanted_.set(file, wanted);
 }
 
-void tr_files_wanted::set(std::span<tr_file_index_t const> const files, bool const wanted)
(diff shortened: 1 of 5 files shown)


libtransmission/file-piece-map.cc
```diff
@@ -212,16 +212,22 @@ tr_files_wanted::tr_files_wanted(tr_file_piece_map const* const fpm)
...
 
-void tr_files_wanted::set(tr_file_index_t const file, bool const wanted)
+bool tr_files_wanted::set(tr_file_index_t const file, bool const wanted)
 {
-    wanted_.set(file, wanted);
+    return wanted_.set(file, wanted);
 }
 
-void tr_files_wanted::set(std::span<tr_file_index_t const> const files, bool const wanted)
+bool tr_files_wanted::set(std::span<tr_file_index_t const> const files, bool const wanted)
```
ckerr, line 220: Only one of these two public functions safeguards against `file >= fpm_->file_count()`.

Maybe this function should actually be the one calling `wanted_.set()` and the function at line 215 should be a passthrough to the function at line 220. That way both would get the file-index-too-large safety check.

ckerr (approved):
Mostly LGTM. One optional suggestion on fpm.set()

libtransmission/file-piece-map.cc
tearfur: The variant at line 215 get's its safeguards from `tr_bitfield`, since `wanted_.size() == fpm_->file_count()`.

PR #227, opened 2026-08-14
refactor: add upper limit for incoming BT payload size

tearfur:
## Description

Incorporate an upper size limit for incoming BT messages into `is_message_length_correct`, currently arbitrarily chosen to be 4MiB. This is meant to prevent malicious peers from launching DoS attacks by exhausting the client's memory.

I've considered adding similar checks in peerIo code, but figured this isn't needed because the read buffer in peerIo never grows large:
1. Data in the peerIo read buffer is always moved into `tr_incoming::payload` soon after it was buffered.
2. The most number bytes buffered into peerIo at a time is limited by the connection's MTU.

I've also stopped using `tr_peer_event::GotError(EMSGSIZE)` because all the event handler code does is log and disconnecting the peer. Nothing concerns the swarm, yet the error details are obfuscated by the errno. I plan to remove the `GotError` event altogether in a future PR.

More details in the commit messages.

Notes: Implemented size limits on peer messages to guard against memory exhaustion.

## Checklist

### DIFF
--- a/libtransmission/peer-msgs.cc
+++ b/libtransmission/peer-msgs.cc
@@ -762,7 +762,7 @@ private:
 
 // ---
 
-[[nodiscard]] constexpr bool is_message_length_correct(tr_torrent const& tor, uint8_t id, uint32_t len)
+[[nodiscard]] constexpr bool is_message_length_correct(tr_torrent const& tor, uint8_t const id, uint32_t const len)
 {
     switch (id) {
     case BtPeerMsgs::Choke:
@@ -771,12 +771,12 @@ private:
     case BtPeerMsgs::NotInterested:
     case BtPeerMsgs::FextHaveAll:
     case BtPeerMsgs::FextHaveNone:
-        return len == 1U;
+        return len == sizeof(id);
 
     case BtPeerMsgs::Have:
     case BtPeerMsgs::FextSuggest:
     case BtPeerMsgs::FextAllowedFast:
-        return len == 5U;
+        return len == sizeof(id) + sizeof(uint32_t /*piece*/);
 
     case BtPeerMsgs::Bitfield:
         if (tor.has_metainfo()) {
@@ -787,16 +787,16 @@ private:
     case BtPeerMsgs::Request:
     case BtPeerMsgs::Cancel:
     case BtPeerMsgs::FextReject:
-        return len == 13U;
+        return len == sizeof(id) + sizeof(uint32_t /*piece*/) + sizeof(uint32_t /*offset*/) + sizeof(uint32_t /*length*/);
 
     case BtPeerMsgs::Piece:
         return len <= sizeof(id) + sizeof(uint32_t /*piece*/) + sizeof(uint32_t /*offset*/) + tr_block_info::BlockSize;
 
     case BtPeerMsgs::DhtPort:
-        return len == 3U;
+        return len == sizeof(id) + sizeof(uint16_t /*port*/);
 
     case BtPeerMsgs::Ltep:
-        return len >= 2U;
+        return len >= sizeof(id) + sizeof(uint8_t /*ltep_msgid*/);
 
     default: // unrecognized message
         return false;
--- a/libtransmission/peer-msgs.cc
+++ b/libtransmission/peer-msgs.cc
@@ -1809,9 +1809,15 @@ ReadResult tr_peerMsgsImpl::can_read_impl(tr_peerIo* io)
 
 ReadState tr_peerMsgsImpl::can_read(tr_peerIo* io, void* vmsgs, size_t* piece)
 {
+    auto const msgs = static_cast<tr_peerMsgsImpl*>(vmsgs);
+    if (msgs->is_disconnecting()) {
+        io->read_buffer_discard();
+        return ReadState::Err;
+    }
+
     // Some errors (e.g. disk IO error) can immediately remove this peer from the peer mgr,
     // which will destroy this object if we don't keep it alive
-    auto const msgs = static_cast<tr_peerMsgsImpl*>(vmsgs)->shared_from_this();
+    auto const keep_alive = msgs->shared_from_this();
 
     auto ret = ReadState::Now;
     *piece = 0U;
--- a/libtransmission/peer-msgs.cc
+++ b/libtransmission/peer-msgs.cc
@@ -808,7 +808,11 @@ private:
         return false;
     }
 
-    return len >= sizeof(id) && len <= MaxIncomingMsgBytes;
+    // IMPORTANT: Any branches with no upper bound checks MUST NOT
+    // return early in order for program execution to reach this line.
+    // N.B. Every caller ensures len >= sizeof(id), so we don't need
+    // to check that here.
+    return len <= MaxIncomingMsgBytes;
 }
 
 namespace protocol_send_message_helpers
--- a/libtransmission/peer-msgs.cc
+++ b/libtransmission/peer-msgs.cc
@@ -1388,17 +1388,7 @@ ReadResult tr_peerMsgsImpl::process_peer_message(uint8_t id, MessageReader& payl
             static_cast<int>(id),
             std::size(payload)));
 
-    if (!is_message_length_correct(tor_, id, sizeof(id) + std::size(payload))) {
-        logdbg(
-            this,
-            fmt::format(
-                "bad msg: '{:s}' ({:d}) with payload len {:d}",
-                BtPeerMsgs::debug_name(id),
-                static_cast<int>(id),
-                std::size(payload)));
-        publish(tr_peer_event::GotError(EMSGSIZE));
-        return { ReadState::Err, {} };
-    }
+    TR_ASSERT(is_message_length_correct(tor_, id, sizeof(id) + std::size(payload)));
 
     switch (id) {
     case BtPeerMsgs::Choke:
@@ -1764,6 +1754,18 @@ ReadResult tr_peerMsgsImpl::can_read_impl(tr_peerIo* io)
 
         io->read_uint8(&message_type);
         current_message_type = message_type;
+    }
+
+    if (!is_message_length_correct(tor_, *current_message_type, *current_message_len)) {
+        logdbg(
+            this,
+            fmt::format(
+                "bad msg: '{:s}' ({:d}) with len {:d}, disconnecting",
+                BtPeerMsgs::debug_name(*current_message_type),
+                static_cast<int>(*current_message_type),
+                *current_message_len));
+        disconnect_soon();
+        return { ReadState::Err, {} };
     }
 
     // read <payload>
--- a/libtransmission/peer-msgs.cc
+++ b/libtransmission/peer-msgs.cc
@@ -170,6 +170,9 @@ auto constexpr Reject = 2;
 
 } // namespace MetadataMsgType
 
+// Arbitrary upper size limit to avoid memory exhaustion attacks.
+auto constexpr MaxIncomingMsgBytes = size_t{ 4U * 1024U * 1024U };
+
 auto constexpr MinChokePeriodSec = time_t{ 10 };
 
 // idle seconds before we send a keepalive
@@ -776,7 +779,10 @@ private:
         return len == 5U;
 
     case BtPeerMsgs::Bitfield:
-        return !tor.has_metainfo() || len == 1 + tr_bytes_needed(tor.piece_count());
+        if (tor.has_metainfo()) {
+            return len == sizeof(id) + tr_bytes_needed(tor.piece_count());
+        }
+        break;
 
     case BtPeerMsgs::Request:
     case BtPeerMsgs::Cancel:
@@ -796,6 +802,8 @@ private:
     default: // unrecognized message
         return false;
     }
+
+    return len <= MaxIncomingMsgBytes;
 }
 
 namespace protocol_send_message_helpers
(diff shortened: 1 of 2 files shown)


libtransmission/peer-msgs.cc
```diff
@@ -759,7 +762,7 @@ class tr_peerMsgsImpl final
 
 // ---
 
-[[nodiscard]] constexpr bool is_message_length_correct(tr_torrent const& tor, uint8_t id, uint32_t len)
+[[nodiscard]] constexpr bool is_message_length_correct(tr_torrent const& tor, uint8_t const id, uint32_t const len)
```
ckerr: +`noexcept`

libtransmission/peer-msgs.cc
```diff
@@ -768,34 +771,38 @@ class tr_peerMsgsImpl final
     case BtPeerMsgs::NotInterested:
     case BtPeerMsgs::FextHaveAll:
     case BtPeerMsgs::FextHaveNone:
-        return len == 1U;
+        return len == sizeof(id);
```
ckerr, line 777: There are other places eg on 799 where we write this as `sizeof(uint8_t /*letp_msgid*/)`. Let's add an `auto constexpr IdLen = sizeof(uint8_t)`  in the `LtepMessages` namespace and use it everywhere?

libtransmission/peer-msgs.cc
```diff
@@ -768,34 +771,38 @@ class tr_peerMsgsImpl final
...
     case BtPeerMsgs::Request:
     case BtPeerMsgs::Cancel:
     case BtPeerMsgs::FextReject:
-        return len == 13U;
+        return len == sizeof(id) + sizeof(uint32_t /*piece*/) + sizeof(uint32_t /*offset*/) + sizeof(uint32_t /*length*/);
 
     case BtPeerMsgs::Piece:
-        len -= sizeof(id) + sizeof(uint32_t /*piece*/) + sizeof(uint32_t /*offset*/);
-        return len <= tr_block_info::BlockSize;
+        return len <= sizeof(id) + sizeof(uint32_t /*piece*/) + sizeof(uint32_t /*offset*/) + tr_block_info::BlockSize;
```
ckerr: This used to drop if len < 9, but now they're OK. I *think* this was unintentional?

```suggestion
    case BtPeerMsgs::Piece:
        {
            auto constexpr HeaderLen = sizeof(id) + sizeof(uint32_t /*piece*/) + sizeof(uint32_t /*offset*/);
            return len >= HeaderLen && len <= HeaderLen + tr_block_info::BlockSize;
        }
```

libtransmission/peer-msgs.cc
```diff
@@ -768,34 +771,38 @@ class tr_peerMsgsImpl final
...
 
     case BtPeerMsgs::DhtPort:
-        return len == 3U;
+        return len == sizeof(id) + sizeof(uint16_t /*port*/);
 
     case BtPeerMsgs::Ltep:
-        return len >= 2U;
+        return len >= sizeof(id) + sizeof(uint8_t /*ltep_msgid*/);
```
ckerr: (claude)

This `return` is the reason the PR does not accomplish what it sets out to. Only the `break` in the `Bitfield` arm (:785) falls through to `MaxIncomingMsgBytes` at :805 — every other case returns, so the cap gates exactly one branch in one torrent state, and the message type with a genuinely open-ended length is not it. This arm accepts `len` ∈ [2, 0xFFFFFFFF].

The exposure itself is not new (`main` is unbounded here too), but it is the specific hole this PR exists to close.

Concretely: a peer sends `<len=0xFFFFFFFF><id=20><ltep_msgid=0xFF>` — an *unknown* sub-id — and trickles bytes. `can_read_impl` accepts at :1758, `full_payload_len` becomes 0xFFFFFFFE, and :1774-1776 grow `incoming_.payload` with no ceiling. Nothing downstream compensates: `parse_ltep` is not called until the whole message is resident, and an unknown sub-id is silently dropped at :967 — so the bencode parser's `MaxBencStrLength` never runs at all. `MessageBuffer`'s `std::ratio<5, 1>` growth factor then overshoots whatever was received.

```suggestion
    case BtPeerMsgs::Ltep:
        return len >= sizeof(id) + sizeof(uint8_t /*ltep_msgid*/) && len <= MaxIncomingMsgBytes;
```

The deeper fix is to hoist `if (len > MaxIncomingMsgBytes) { return false; }` above the switch, or to bound `full_payload_len` at :1774 where the buffer actually grows. Either makes the cap type-independent instead of opt-in per case; see the :801-805 comment.

libtransmission/peer-msgs.cc
```diff
@@ -768,34 +771,38 @@ class tr_peerMsgsImpl final
...
 
     default: // unrecognized message
         return false;
     }
+
+    return len >= sizeof(id) && len <= MaxIncomingMsgBytes;
```
ckerr: (claude)

the shared tail reads as a backstop but governs one arm

Two problems in three lines.

The `len >= sizeof(id)` conjunct can never be false. `can_read_impl` returns early on `message_len == 0` (:1743) before the id is read, and `build_peer_message` starts `msg_len` at `sizeof(type)` (:887), so every one of the three call sites is already `>= 1`.

More importantly, a tail reached by a single `break` reads like a default that applies to the whole switch. It does not — and the failure mode is silent: any message type added here inherits no cap unless its author remembers to `break` rather than `return`. That is an opt-in default for a security invariant, with nothing in the switch signalling which way is safe.

```suggestion
    default: // unrecognized message
        return false;
    }

    // Only a metainfo-less Bitfield reaches here: its length cannot be
    // checked exactly until the piece count is known, so bound it instead.
    return len <= MaxIncomingMsgBytes;
}
```

libtransmission/peer-msgs.cc
```diff
@@ -1758,6 +1755,18 @@ ReadResult tr_peerMsgsImpl::can_read_impl(tr_peerIo* io)
...
+        logdbg(
+            this,
+            fmt::format(
+                "bad msg: '{:s}' ({:d}) with len {:d}, disconnecting",
+                BtPeerMsgs::debug_name(*current_message_type),
+                static_cast<int>(*current_message_type),
+                *current_message_len));
+        disconnect_soon();
+        return { ReadState::Err, {} };
+    }
```
ckerr, line 1784: (claude)

the check re-runs on every partial read, against state that can change

This sits outside the `if (!current_message_type)` block above it, so it is re-evaluated once per readable socket event for the whole life of a message, not once when the header is parsed.

That is mostly waste — a 16 KiB Piece arriving in ~1448-byte segments re-answers the same question ~12 times on the receive hot path — but for `Bitfield` it is a behavior change. The verdict depends on `tor_.has_metainfo()`, which can flip between passes: on a magnet download a peer's bitfield is accepted on pass 1 via the metainfo-less `break`, returns `Later` with `incoming_` retained, metadata completes from another peer between event-loop turns, and pass 2 takes the `if (tor.has_metainfo())` branch at :782, demands the exact `sizeof(id) + tr_bytes_needed(tor.piece_count())`, and drops a peer that did nothing wrong. `on_torrent_got_metainfo()` (:386-408) already applies that rule at the right moment, after the bitfield has been recorded. The pre-PR code evaluated the predicate exactly once, after the payload was complete.

GitHub is glitching on me making a suggestion since part of the change isn't in the changed code, but this is a possible replacement for 1748-1768:

```c++
    if (!current_message_type) {
        auto message_type = uint8_t{};
        if (io->read_buffer_size() < sizeof(message_type)) {
            return { ReadState::Later, {} };
        }

        io->read_uint8(&message_type);
        current_message_type = message_type;

        // The header is complete, so the length can be checked
        // before any payload is buffered.
        if (!is_message_length_correct(tor_, message_type, *current_message_len)) {
            logdbg(
                this,
                fmt::format(
                    "bad msg: '{:s}' ({:d}) with len {:d}, disconnecting",
                    BtPeerMsgs::debug_name(message_type),
                    static_cast<int>(message_type),
                    *current_message_len));
            current_message_len.reset();
            current_message_type.reset();
            disconnect_soon();
            return { ReadState::Err, {} };
        }
    }
```

libtransmission/peer-msgs.cc
```diff
@@ -1758,6 +1755,18 @@ ReadResult tr_peerMsgsImpl::can_read_impl(tr_peerIo* io)
...
+    if (!is_message_length_correct(tor_, *current_message_type, *current_message_len)) {
+        logdbg(
+            this,
+            fmt::format(
+                "bad msg: '{:s}' ({:d}) with len {:d}, disconnecting",
+                BtPeerMsgs::debug_name(*current_message_type),
+                static_cast<int>(*current_message_type),
+                *current_message_len));
+        disconnect_soon();
+        return { ReadState::Err, {} };
```
ckerr: (claude)

This early return bypasses the resets at :1790-1792. The removed check reached them: it lived inside `process_peer_message`, so control returned to :1788 and fell through to `current_message_len.reset(); current_message_type.reset(); current_payload.clear();`.

Nothing else picks up the slack. `disconnect_soon()` (`peer-msgs.h:112`) only sets `is_disconnecting_`; teardown waits for peer-mgr's reconnect pulse. `tr_peerIo::can_read_wrapper` treats `ReadState::Err` as `err = true` and merely breaks its loop (`peer-io.cc:362`) — it does not close the socket or disable reads. So the next `EV_READ` re-enters `can_read_impl` with the stale `incoming_.id`/`length`, fails the identical check, and re-emits the same `logdbg`, repeatedly, while nothing drains the offending bytes. Bounded by `RcvBuf` = 256 KiB, so this is log churn and a retained buffer rather than growth — but it is a straight regression from the old path.

Anchor: lines 1766-1767. Skip this if you take the :1758 suggestion, which subsumes it.

```suggestion
        current_message_len.reset();
        current_message_type.reset();
        disconnect_soon();
        return { ReadState::Err, {} };
```

Separately, and lower stakes: this is the only one of six `publish(tr_peer_event::GotError(...))` sites in the file converted to `disconnect_soon()`. The PR body says that's deliberate groundwork, and it is behavior-neutral (`peer-mgr.cc:656-658` maps `EMSGSIZE` straight to `disconnect_soon()`), so this is scope rather than defect. Two costs worth weighing: it loses the swarm-level `tr_logAddDebugSwarm` at `peer-mgr.cc:659` that the five survivors still get, and it strands the errno's meaning — after this PR the only genuine message-size error no longer uses `EMSGSIZE`, while all five remaining `EMSGSIZE` sites (:1523, :1536, :1555, :1569, :1593) are FEXT-without-negotiation. A reader later deleting the now-misnamed `EMSGSIZE` arm of the peer-mgr allowlist would silently disable those.

libtransmission/peer-msgs.cc
ckerr, line 653: the precondition `process_peer_message` relies on is undocumented 

```suggestion
    // Precondition: `payload` is a complete message body whose length
    // can_read_impl() has already accepted for `id`.
    ReadResult process_peer_message(uint8_t id, MessageReader& payload);
```

ckerr (commented):
me: "I'd add a symbolic constant for clarity, but LGTM"

Claude: "LOL no. I found 11 problems."

Sorry @tearfur 😵

libtransmission/peer-msgs.cc
tearfur: `sizeof(id)` is the BitTorrent message type, different from ltep message ID.

libtransmission/peer-msgs.cc
tearfur: The first part about the early return is moot, as it is handled in the other review comment thread.

The second part about `GotError` is moot because of #238.

libtransmission/peer-msgs.cc
tearfur: 71fb852b0afd4bfbf493665dd9921066f8b3ce7e

libtransmission/peer-msgs.cc
tearfur: Oof this was a dumb one. a1a0c758fe511d77ac728d2e9509f484ce403ed5

libtransmission/peer-msgs.cc
tearfur: Oof another dumb one. 1b78b47da4d40a559b51271f847f4e657064fc7d

libtransmission/peer-msgs.cc
tearfur: 879d92dc6e69ff23e69b0e3def2d8dbce97c8c99

libtransmission/peer-msgs.cc
tearfur: Makes sense to move the `is_message_length_correct` call into `if (!current_message_type)`. 281794b1aee365622ed25555c686c734a53c350d

---

The suggested code resets `current_message_len` and `current_message_type`.

Not that this is bad idea, but I think the thing to do here is to drop the peerIo buffer and return early in `can_read()` if the peer is disconnecting, as there's no reason to process any more incoming messages from such peers. I've checked all `disconnect_soon()` call sites, and I think this holds everywhere.

I'd like a second set of brains to double verify on this.

0193b24906b358162649cd3bcfb750e9fb2a3c4a

---

The part about `Bitfield` is not making sense to me. What do you mean the peer did nothing wrong when the size of the bitfield is wrong?

libtransmission/peer-msgs.cc
tearfur: dba68d50d2f7ab463328eed0dea1028aa3c8a51f

libtransmission/peer-msgs.cc
ckerr: 0193b24 makes sense to me.

ckerr (approved):

PR #228, opened 2026-08-14
perf: avoid re-allocating `have_` on bitfield msg

tearfur:
## Description

Simple perf improvement by re-using the existing bitfield object.

Extracted from #216.

## Checklist

### DIFF
--- a/libtransmission/peer-msgs.cc
+++ b/libtransmission/peer-msgs.cc
@@ -1551,7 +1551,7 @@ ReadResult tr_peerMsgsImpl::process_peer_message(uint8_t id, MessageReader& payl
 
     case BtPeerMsgs::Bitfield:
         logtrace(this, "got a bitfield");
-        have_ = tr_bitfield{ tor_.has_metainfo() ? tor_.piece_count() : std::size(payload) * 8 };
+        have_.init_size(std::size(payload) * 8);
         {
             [[maybe_unused]] auto const set_raw_res = have_.set_raw(payload);
             TR_ASSERT(set_raw_res);


libtransmission/peer-msgs.cc
```diff
@@ -1551,7 +1551,7 @@ ReadResult tr_peerMsgsImpl::process_peer_message(uint8_t id, MessageReader& payl
 
     case BtPeerMsgs::Bitfield:
         logtrace(this, "got a bitfield");
-        have_ = tr_bitfield{ tor_.has_metainfo() ? tor_.piece_count() : std::size(payload) * 8 };
+        have_.init_size(std::size(payload) * 8);
```
ckerr, line 1554: 1. I don't think this actually has any performance win. `bitfield` is supposed to be the peer's first message, so I think `have_` will always be empty before calling `operator=` or `init_size()` here?

2. The old version is safer for BEP-breaking events eg a second bitfield message from the same peer. Looks like libtorrent disconnects on the second message but we don't. (We probably should, but that's out-of-scope here).

3. `init_size()`'s `[[nodiscard]]` return value is discarded here

tldr: maybe we should skip this one

tearfur: I've decided to drop this PR. You're right that this won't have performance improvements in normal situations.

My understanding of libtorrent different from you though. They don't drop the peer on a second Bitfield/HaveAll/HaveNone. They effectively overwrite the first bitfield with the second one by first reversing the bookkeeeping for the first bitfield by calling `libtorrent::torrent::peer_lost`, then do the bookkeeping for the new bitfield.

PR #229, opened 2026-08-14
fix(wishlist): prevent duplicate bitfields inflating replication

tearfur:
## Description

Each peer should send exactly one of `bitfield`, `have all`, or `have none` immediately after the BitTorrent handshake.

However, to defend against clients that send these messages more than once, we decrement the piece replication using the old bitfield before accepting the new bitfield.

Notes: Added defense against peers that send duplicated bitfield messages.

## Checklist

### DIFF
--- a/libtransmission/peer-common.h
+++ b/libtransmission/peer-common.h
@@ -60,6 +60,7 @@ public:
         ClientSentCancel,
         ClientSentPieceData,
         ClientSentRequest,
+        ClientStaleBitfield,
         Error // generic
     };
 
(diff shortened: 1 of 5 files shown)


libtransmission/peer-common.h
```diff
@@ -189,6 +190,14 @@ class tr_peer_event
         event.length = loc_end.byte - loc_begin.byte;
         return event;
     }
+
+    [[nodiscard]] constexpr static auto StaleBitfield(tr_bitfield* bitfield) noexcept
+    {
+        auto event = tr_peer_event{};
+        event.type = Type::ClientStaleBitfield;
+        event.bitfield = bitfield;
```
ckerr: I might be missing it but it doesn't look like `tr_peer_event::bitfield` is ever read? Maybe we can remove this?

PR #230, opened 2026-08-14
build: require macOS 12

nevack:
> [!NOTE]
> This pull request was created by an AI agent.

## Summary
- raise the CMake minimum macOS deployment target to 12.0
- align the Qt CMake target with the project minimum
- update all maintained Xcode build configurations to target macOS 12.0

## Validation
- CMake configuration completed successfully with CMAKE_OSX_DEPLOYMENT_TARGET=12.0
- the Xcode project passed plutil validation and xcodebuild project listing
- git diff --check passed


nevack:
Xcode 27.0 beta sets deployment target of 12.0 as a hard requirement and refuses to build otherwise

qt/CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
 set_property(GLOBAL PROPERTY AUTOGEN_SOURCE_GROUP "Generated Files")
 
 # https://doc.qt.io/qt-6/macos.html#supported-versions
-set(CMAKE_OSX_DEPLOYMENT_TARGET 11)
```
tearfur, line 4: We can remove this as it was already set here:

https://github.com/retransmission/retransmission/blob/39b4057bc2c790f302c108a7cda3e5ddb86592e8/CMakeLists.txt#L14-L18

lolgear:
I don't understand (re)transmission release cycle.
macOS 27 is in beta yet.

macOS 11
macOS 12
macOS 13
macOS 14
macOS 15
macOS 26 ( Tahoe, we are here )
macOS 27 ( beta )

I guess (re)transmission should drop macOS 11 _after_ macOS 27 release. October-November 2026.

ckerr:
> I don't understand (re)transmission release cycle.

I don't think it's been discussed.

CMakeLists.txt
```diff
@@ -7,7 +7,7 @@ if(POLICY CMP0114)
 endif()
 
 # Value should follow latest stable Xcode's RECOMMENDED_MACOSX_DEPLOYMENT_TARGET
-set(MACOS_SUPPORT_MINIMUM 11.0)
+set(MACOS_SUPPORT_MINIMUM 12.0)
```
ckerr, line 12: The comment states the rule this value follows, but 12.0 doesn't come from the latest stable Xcode's RECOMMENDED_MACOSX_DEPLOYMENT_TARGET — it comes from Xcode 27 refusing to build anything lower. A maintainer who checks the stated source won't derive 12.0 from it and can't tell whether the value is current or stale. Keep the rule as the standing default, but lead with the constraint that actually pins the value.

```suggestion
# Xcode 27 rejects deployment targets below 12.0.
# Otherwise this follows the latest stable Xcode's RECOMMENDED_MACOSX_DEPLOYMENT_TARGET.
set(MACOS_SUPPORT_MINIMUM 12.0)
```

ckerr (commented):
(claude)

- docs/Building-Transmission.md still documents macOS 11

- macosx/DefaultAppHelper.mm:48, 79, 128 : At a 12.0 deployment target these @available(macOS 12, *) tests are always true, so the #ifdef __MAC_12_0 else-branches are unreachable. They also call LSCopyDefaultRoleHandlerForContentType (line 63) and LSSetDefaultRoleHandlerForContentType (line 95), both API_DEPRECATED(..., macos(10.4, 12.0)) — clang gates -Wdeprecated-declarations on the deployment target, so this file starts warning on every macOS build, and the warnings can't be fixed while the branches exist. Delete the #ifdef/@available scaffolding and the else-branches.

- macosx/PowerManager.mm:69, 87, 120 — [fix here] · @available(macOS 12) guards always true at a 12.0 floor, as is the else if (@available(macOS 12, *)) arm in macosx/PrefsController.mm:775 — whose final else (779–781) becomes unreachable while still constructing a /System/Library/PreferencePanes/Notifications.prefPane URL. Collapse them.

qt/CMakeLists.txt
nevack: Addressed in https://github.com/retransmission/retransmission/pull/230/changes/aa6aa1d9453b0f5d2cbdf6c9e13e49bc28316ebc

nevack:
> (claude)
> 
> * docs/Building-Transmission.md still documents macOS 11
> * macosx/DefaultAppHelper.mm:48, 79, 128 : At a 12.0 deployment target these @available(macOS 12, *) tests are always true, so the #ifdef __MAC_12_0 else-branches are unreachable. They also call LSCopyDefaultRoleHandlerForContentType (line 63) and LSSetDefaultRoleHandlerForContentType (line 95), both API_DEPRECATED(..., macos(10.4, 12.0)) — clang gates -Wdeprecated-declarations on the deployment target, so this file starts warning on every macOS build, and the warnings can't be fixed while the branches exist. Delete the #ifdef/@available scaffolding and the else-branches.
> * macosx/PowerManager.mm:69, 87, 120 — [fix here] · @available(macOS 12) guards always true at a 12.0 floor, as is the else if (@available(macOS 12, *)) arm in macosx/PrefsController.mm:775 — whose final else (779–781) becomes unreachable while still constructing a /System/Library/PreferencePanes/Notifications.prefPane URL. Collapse them.

I initially wanted this to be done separately after merging this PR.
Addressed in https://github.com/retransmission/retransmission/pull/230/changes/5e4307ba6a0388573935d8f407a113d5240d0b47 and https://github.com/retransmission/retransmission/pull/230/changes/084649a826ddbfd0feabbf8de7151f2c89de87ee

CMakeLists.txt
nevack: Addressed in https://github.com/retransmission/retransmission/pull/230/changes/aa6aa1d9453b0f5d2cbdf6c9e13e49bc28316ebc

tearfur (approved):
CMake changes LGTM.

ckerr (approved):
LGTM as well

Coeur:
From lolgear:
>I guess (re)transmission should drop macOS 11 after macOS 27 release. October-November 2026.

1. Apple has probably always been releasing their major updates around the second week of September.
2. Even in beta, the newer Xcode is installed and in use by many developers, to prepare and support that OS upgrade. So as soon as it was offered to developers during June's WWDC, it's legitimate to support it.

Coeur (approved):

PR #233, opened 2026-08-14
Remove Legacy QuickLook Plugin

nevack:
This plugin system is not supported as of macOS 15.
New QuickLook appex extension is provided on macOS 12+ as a replacement.

Notes: Remove legacy QuickLook plugin for previewing torrent files as there's already modern QuickLook extension in place for the same functionality.


Transmission.xcodeproj/project.pbxproj
```diff
@@ -3,29 +3,19 @@
 	archiveVersion = 1;
 	classes = {
 	};
-	objectVersion = 54;
+	objectVersion = 90;
```
lolgear: No.
It should be something like 70 ( Xcode 16 ).

docs/Building-Transmission.md
```diff
@@ -6,8 +6,8 @@ While the prerequisites for running the app are:
  * macOS 11.0 or newer
 
 The actual software prerequisites for building the app are based on GitHub Actions [macos-14](https://github.com/actions/runner-images/blob/main/images/macos/macos-14-arm64-Readme.md):
- * macOS 14.8.3 or newer
- * Xcode 15.0.1 or newer
+ * macOS 15.2 or newer
+ * Xcode 16.3 or newer
```
Coeur: This feels like the version bump is done in order to delete older code?

Could we maybe figure out clear rules about which versions of macOS we desire to support, follow those rules, and do cleanup such as removing "legacy" code only then?

Aka separate pull requests:
1. One to update the minimum supported version based on clear agreement.
2. Then one to perform cleanup as needed.

Coeur: From what I recall in Transmission, we wanted to roughly follow the supported version offered by GitHub actions.

Coeur: in #224, I saw you write:
>I have a plan to update minumum Xcode and deployment target in a separate PR, stay tuned.

I agree we should bump Xcode to something higher. Let's just do it in its own PR, for easy reference.

Coeur: Oh, I just found your macOS bump PR here:
- #230

nevack:
Cleaned up unrelated changes, ready to merge as https://github.com/retransmission/retransmission/pull/230 is merged.
I would like at least one approval for this to go in 😄

lolgear:
LGTM

lolgear (approved):

PR #235, opened 2026-08-18
perf: shrink FileListNode with a class cluster

lolgear:
This PR reduces memory footprint of FileListNode by splitting it into OnlyFileListNode ( files ) and OnlyFolderListNode.

Results using `class_getInstanceSize`.

| Class | Before Refactor | After Refactor |
|---|---|---|
| File Nodes (OnlyFileListNode) | ~64-72 bytes | 48 bytes |
| Folder Nodes (OnlyFolderListNode) | ~72 bytes | 56 bytes |

### DIFF
--- a/macosx/FileListNode.h
+++ b/macosx/FileListNode.h
@@ -14,24 +14,23 @@
 @property(nonatomic, weak, readonly) Torrent* torrent;
 
 @property(nonatomic, readonly) uint64_t size;
-@property(nonatomic, readonly) NSImage* icon;
 @property(nonatomic, readonly) BOOL isFolder;
 @property(nonatomic, readonly) NSMutableArray<FileListNode*>* children;
 
 @property(nonatomic, readonly) NSIndexSet* indexes;
 
-- (instancetype)initWithFolderName:(NSString*)name path:(NSString*)path torrent:(Torrent*)torrent;
-- (instancetype)initWithFileName:(NSString*)name
-                            path:(NSString*)path
-                            size:(uint64_t)size
-                           index:(NSUInteger)index
-                         torrent:(Torrent*)torrent;
-
 - (void)insertChild:(FileListNode*)child;
 - (void)insertIndex:(NSUInteger)index withSize:(uint64_t)size;
 
-@property(nonatomic, readonly) NSString* description;
-
 - (BOOL)updateFromOldName:(NSString*)oldName toNewName:(NSString*)newName inPath:(NSString*)path;
 
+@end
+
+@interface FileListNode (Creation)
++ (instancetype)createWithFolderName:(NSString*)name path:(NSString*)path torrent:(Torrent*)torrent;
++ (instancetype)createWithFileName:(NSString*)name
+                              path:(NSString*)path
+                              size:(uint64_t)size
+                             index:(NSUInteger)index
+                           torrent:(Torrent*)torrent;
 @end
--- a/macosx/FileListNode.mm
+++ b/macosx/FileListNode.mm
@@ -10,52 +10,37 @@
 
 #import "FileListNode.h"
 
-@interface FileListNode ()
+@interface FileListNode () {
+  @protected
+    uint64_t _size;
+}
+- (instancetype)initWithName:(NSString*)name path:(NSString*)path torrent:(Torrent*)torrent size:(uint64_t)size;
+@end
 
-@property(nonatomic, readonly) NSMutableIndexSet* indexesInternal;
-@property(nonatomic) NSImage* iconInternal;
+@interface OnlyFileListNode : FileListNode {
+    NSIndexSet* _indexesInternal;
+}
+@end
 
+@interface OnlyFolderListNode : FileListNode {
+    NSMutableIndexSet* _indexesInternal;
+    NSMutableArray* _childrenInternal;
+}
 @end
 
 @implementation FileListNode
 
-- (instancetype)initWithFolderName:(NSString*)name path:(NSString*)path torrent:(Torrent*)torrent
+- (NSMutableArray<FileListNode*>*)children
 {
-    if ((self = [self initWithFolder:YES name:name path:path torrent:torrent])) {
-        _children = [[NSMutableArray alloc] init];
-        _size = 0;
-    }
-
-    return self;
-}
-
-- (instancetype)initWithFileName:(NSString*)name
-                            path:(NSString*)path
-                            size:(uint64_t)size
-                           index:(NSUInteger)index
-                         torrent:(Torrent*)torrent
-{
-    if ((self = [self initWithFolder:NO name:name path:path torrent:torrent])) {
-        _size = size;
-        [_indexesInternal addIndex:index];
-    }
-
-    return self;
+    return nil;
 }
 
 - (void)insertChild:(FileListNode*)child
 {
-    NSAssert(_isFolder, @"method can only be invoked on folders");
-
-    [_children addObject:child];
 }
 
 - (void)insertIndex:(NSUInteger)index withSize:(uint64_t)size
 {
-    NSAssert(_isFolder, @"method can only be invoked on folders");
-
-    [self.indexesInternal addIndex:index];
-    _size += size;
 }
 
 - (id)copyWithZone:(NSZone*)zone
@@ -66,25 +51,17 @@
 
 - (NSString*)description
 {
-    if (!_isFolder) {
-        return [NSString stringWithFormat:@"%@ (%ld)", _name, _indexesInternal.firstIndex];
-    } else {
-        return [NSString stringWithFormat:@"%@ (folder: %@)", _name, _indexesInternal];
-    }
+    return @"";
 }
 
-- (NSImage*)icon
+- (BOOL)isFolder
 {
-    if (!_iconInternal) {
-        _iconInternal = [NSWorkspace.sharedWorkspace
-            iconForFileType:_isFolder ? NSFileTypeForHFSTypeCode(kGenericFolderIcon) : _name.pathExtension];
-    }
-    return _iconInternal;
+    return NO;
 }
 
 - (NSIndexSet*)indexes
 {
-    return _indexesInternal;
+    return nil;
 }
 
 - (BOOL)updateFromOldName:(NSString*)oldName toNewName:(NSString*)newName inPath:(NSString*)path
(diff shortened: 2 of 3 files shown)


macosx/FileListNode.h
```diff
@@ -14,24 +14,23 @@
...
-- (instancetype)initWithFileName:(NSString*)name
-                            path:(NSString*)path
-                            size:(uint64_t)size
-                           index:(NSUInteger)index
-                         torrent:(Torrent*)torrent;
-
 - (void)insertChild:(FileListNode*)child;
 - (void)insertIndex:(NSUInteger)index withSize:(uint64_t)size;
 
-@property(nonatomic, readonly) NSString* description;
```
lolgear, line 33: Already satisfied as "requirement" of "NSObjectProtocol" ( NSObject class )

ckerr:
@lolgear just a heads-up I'm going to have limited time for reviews for a few days. Should be back to normal soon.

macosx/FileListNode.h
```diff
@@ -14,24 +14,23 @@
...
 
-- (instancetype)initWithFolderName:(NSString*)name path:(NSString*)path torrent:(Torrent*)torrent;
-- (instancetype)initWithFileName:(NSString*)name
-                            path:(NSString*)path
-                            size:(uint64_t)size
-                           index:(NSUInteger)index
-                         torrent:(Torrent*)torrent;
-
 - (void)insertChild:(FileListNode*)child;
 - (void)insertIndex:(NSUInteger)index withSize:(uint64_t)size;
```
ckerr, line 23: Should these two be moved below tot he `@interface FileListNode` section?

macosx/FileListNode.mm
ckerr:
```suggestion
```

AppKit no longer needed in this file

macosx/FileListNode.mm
```diff
@@ -10,52 +10,37 @@
...
+@interface OnlyFileListNode : FileListNode {
+    NSIndexSet* _indexesInternal;
+}
+@end
 
+@interface OnlyFolderListNode : FileListNode {
+    NSMutableIndexSet* _indexesInternal;
+    NSMutableArray* _childrenInternal;
+}
 @end
```
ckerr: (claude)

macosx/FileListNode.mm:20, 25 — naming diverges from the split added one commit earlier (new)
BaseFileNameCellView.h names this exact shape BaseFileNameCellView / FileNameCellView / FolderNameCellView, and it landed immediately before this PR's base. Here the base carries no abstract marker and the subclasses take an Only prefix that already means something else in this codebase — @"OnlyFileCell" / @"OnlyFolderCell" are cell-view identifier strings at FileOutlineController.mm:212,219.

Suggest FileListFileNode / FileListFolderNode (keeping the FileList prefix — ObjC class names are process-global, unlike a C++ anonymous namespace).

macosx/FileListNode.mm
```diff
@@ -10,52 +10,37 @@
...
-@property(nonatomic) NSImage* iconInternal;
+@interface OnlyFileListNode : FileListNode {
+    NSIndexSet* _indexesInternal;
+}
+@end
 
+@interface OnlyFolderListNode : FileListNode {
+    NSMutableIndexSet* _indexesInternal;
+    NSMutableArray* _childrenInternal;
+}
```
ckerr, line 29: _(claude)_

undeclared initializers + duplicated `_indexesInternal` + lost generic *(new)*

Three issues in one block; combined suggestion below.

**a. The initializers are declared in no `@interface`** — the most serious finding in the PR.
`-initWithFolderName:` (line 157) and `-initWithFileName:` (line 121) exist only inside their
`@implementation` blocks, so `+createWithFolderName:` at line 202 resolves them purely
because those blocks are parsed first. Move `@implementation FileListNode (Creation)` above
them — a cosmetic "public API first" reorder, or the split into separate files a cluster
invites — and the build fails:

```
error: no visible @interface for 'OnlyFolderListNode' declares the selector 'initWithFolderName:'
```

Reproduced with clang by two reviewers independently. Under ARC the receiver is statically
typed via `+alloc`, so this is a hard error, not a tolerable warning.

**b. `_indexesInternal` is declared twice with different static types**, and `-indexes` is
written three times. One `NSIndexSet*` in the class extension is byte-identical in object
size (48 / 56 either way) and removes ~14 lines.

**c. `_childrenInternal` lost its element type** — `main`'s `_children` was auto-synthesized
from `@property NSMutableArray<FileListNode*>*` and so *was* generic; the hand-written ivar
drops it, making `insertChild:`'s `addObject:` unchecked while every caller stays typed.


```diff
 @interface FileListNode () {
   @protected
     uint64_t _size;
+    NSIndexSet* _indexesInternal;
 }
 - (instancetype)initWithName:(NSString*)name path:(NSString*)path torrent:(Torrent*)torrent size:(uint64_t)size;
 @end

-@interface OnlyFileListNode : FileListNode {
-    NSIndexSet* _indexesInternal;
-}
+@interface OnlyFileListNode : FileListNode
+- (instancetype)initWithFileName:(NSString*)name
+                            path:(NSString*)path
+                            size:(uint64_t)size
+                           index:(NSUInteger)index
+                         torrent:(Torrent*)torrent;
 @end

 @interface OnlyFolderListNode : FileListNode {
-    NSMutableIndexSet* _indexesInternal;
-    NSMutableArray* _childrenInternal;
+    NSMutableArray<FileListNode*>* _childrenInternal;
 }
+- (instancetype)initWithFolderName:(NSString*)name path:(NSString*)path torrent:(Torrent*)torrent;
 @end

Follow-on hunks for (b) — base accessor gains the ivar, both overrides go away:

```diff
 - (NSIndexSet*)indexes
 {
-    return nil;
+    return _indexesInternal;
 }
```
```diff
-- (NSIndexSet*)indexes
-{
-    return _indexesInternal;
-}
-
 - (void)insertChild:(FileListNode*)child
 {
     NSAssert(NO, @"method can only be invoked on folders");
 }
```
```diff
-- (NSIndexSet*)indexes
-{
-    return _indexesInternal;
-}
-
 - (NSMutableArray<FileListNode*>*)children
 {
     return _childrenInternal;
 }
```
```diff
 - (void)insertIndex:(NSUInteger)index withSize:(uint64_t)size
 {
-    [_indexesInternal addIndex:index];
+    [(NSMutableIndexSet*)_indexesInternal addIndex:index];
     _size += size;
 }
```

Both subclass initializers keep assigning `_indexesInternal` unchanged. The base `-indexes`
implementation must stay: `indexes` is still a `readonly` property in the header, so
deleting the getter would make clang auto-synthesize a backing ivar on every node and undo
the PR's memory saving. The same is true of the `children` and `isFolder` stubs — they are
load-bearing, not dead code.


```

File: macosx/FileListNode.mm [hunk 5]
```diff
@@ -10,52 +10,37 @@
...
-    [_children addObject:child];
 }
 
 - (void)insertIndex:(NSUInteger)index withSize:(uint64_t)size
 {
-    NSAssert(_isFolder, @"method can only be invoked on folders");
-
-    [self.indexesInternal addIndex:index];
-    _size += size;
 }
```
ckerr, line 49: (claude)

`main` had `NSAssert(_isFolder, @"method can only be invoked on folders")` in both of these.
The PR's base versions are silent in every configuration, Debug included — while the file
subclass 100 lines below asserts. The base is now the weakest path, and the assertion text
is duplicated.

`doesNotRecognizeSelector:` is the better guard on both counts: it survives
`NS_BLOCK_ASSERTIONS` (defined in the Xcode `Release` config at `project.pbxproj:5061`,
where `NSAssert` compiles to nothing), and it lets `OnlyFileListNode`'s two overrides be
deleted outright.

```diff
 - (void)insertChild:(FileListNode*)child
 {
+    [self doesNotRecognizeSelector:_cmd];
 }

 - (void)insertIndex:(NSUInteger)index withSize:(uint64_t)size
 {
+    [self doesNotRecognizeSelector:_cmd];
 }
```

Then delete the now-redundant overrides in `OnlyFileListNode`:

```diff
-- (void)insertChild:(FileListNode*)child
-{
-    NSAssert(NO, @"method can only be invoked on folders");
-}
-
-- (void)insertIndex:(NSUInteger)index withSize:(uint64_t)size
-{
-    NSAssert(NO, @"method can only be invoked on folders");
-}
-
 - (NSString*)description
```

macosx/FileListNode.h
lolgear: It's better to keep them in main interface because overriding methods in different ( named category ) part of interface is a bit tricky.

```objective-c
@interface FileListNode (Actions)
- (void)insertChild:(FileListNode*)child;
- (void)insertIndex:(NSUInteger)index withSize:(uint64_t)size;Collapse comment;
@end
```

I prefer to keep them in main public interface.
However, let's imagine how it could be done in the most "objective-cish" way.

Let's extract methods.

### A possible implementation for composition of interfaces in class cluster.

Consider you have a folder and a file.
Folder has two additional methods.

Let's extract these methods into interface ( interface and protocol in objective-c are the same for reader ).

```objective-c
/// FileListNode.h
@protocol FolderActionsProtocol
- (void)insertChild:(FileListNode*)child;
- (void)insertIndex:(NSUInteger)index withSize:(uint64_t)size;
@end

@interface FileListNode

/// Define a method which will return nothing OR object which has methods from FolderActionsProtocol.
- (_Nullable id<FolderActionsProtocol>)folder;
@end
```

We already have two subclasses and we know that only folder can do stuff with children nodes.

```objective-c
@implementation FileListNode

/// By default, base class doesn't know about these methods.
- (_Nullable id<FolderActionsProtocol>)folder {
	return nil;
}

@end

@implementation FileListFolderNode
- (void)insertChild:(FileListNode*)child {
	/// ...
}

- (void)insertIndex:(NSUInteger)index withSize:(uint64_t)size {
	/// ...
}

/// But only folder _can_ insert children.
- (_Nullable id<FolderActionsProtocol>)folder {
	return self;
}
@end
```

macosx/FileListNode.mm
```diff
@@ -2,60 +2,50 @@
...
-    if ((self = [self initWithFolder:NO name:name path:path torrent:torrent])) {
-        _size = size;
-        [_indexesInternal addIndex:index];
-    }
+                         torrent:(Torrent*)torrent;
+@end
 
-    return self;
+@interface FileListFolderNode : FileListNode {
+    NSMutableArray* _childrenInternal;
```
ckerr:
```suggestion
    NSMutableArray<FileListNode*>* _childrenInternal;
```

lolgear:
@ckerr 
Could we merge it?

lolgear:
@ckerr 
Ready for merge?

ckerr:
@lolgear oop apparently I left a review comment in draft :frowning_face:

macosx/FileListNode.mm
lolgear: Done

PR #237, opened 2026-08-19
Prevent repeated fatal peer-I/O callbacks

### LINKED ISSUE #236: Repeated fatal peer-I/O callbacks before asynchronous peer removal
### What is the issue?

A single failed peer connection can invoke the fatal error callback multiple times before the peer is removed asynchronously.

This is visible in debug logs as repeated messages for the same peer, often within the same millisecond:

```text
setting [...] is_disconnecting_ flag because we got [(107) Socket is not connected]
```

These messages do not always represent separate connection failures. The same failed peer-I/O object may be processed repeatedly by subsequent bandwidth and I/O passes before its removal is completed.

## Investigation

This issue was found during a broader investigation into long-standing, difficult-to-diagnose networking and peer-management problems on high-load seedboxes with tens of thousands of loaded torrents.

A large temporary diagnostic patch was developed specifically for this investigation. It adds dozens of aggregated metrics covering:

* incoming and outgoing peer lifecycles;
* handshake outcomes;
* actual transport errors and their direction;
* callback counts per peer session;
* buffered and post-handshake input;
* wire messages and piece payload;
* reconnect intervals;
* candidate selection and connection attempts;
* peer creation and removal.

The diagnostics patch does not change peer admission, peer limits, candidate selection, queueing, choking, or disconnect behavior.

The key distinction made by the diagnostics is between:

* `transport_error_count`: the total number of fatal error callbacks;
* `transport_error_sessions`: the number of distinct peer sessions that received at least one fatal callback.

## Results

Before the experimental fix:

| Direction | Fatal callbacks | Affected sessions | Callbacks per session |
| --------- | --------------: | ----------------: | --------------------: |
| Incoming  |          70,219 |            24,087 |                 2.915 |
| Outgoing  |          30,872 |            14,092 |                 2.190 |

One failed incoming peer therefore generated almost three fatal callbacks on average.

After the experimental fix:

| Direction | Fatal callbacks | Affected sessions | Callbacks per session |
| --------- | --------------: | ----------------: | --------------------: |
| Incoming  |          27,744 |            27,744 |                 1.000 |
| Outgoing  |             715 |               715 |                 1.000 |

A later nine-hour validation run produced the same invariant:

```text
incoming: 67602 / 67602 = 1.000
outgoing:   259 /   259 = 1.000
```

The ratio remained exactly `1.000` despite the much larger observation window.

This confirms that the repeated messages were caused by repeated fatal callbacks for the same failed connection, rather than only by large numbers of independent connection failures.

## Expected behavior

Each socket instance should report its first non-retryable fatal transport error once.

A replacement socket or a new connection should have an independent error state and must still be able to report its own failure normally.

## Impact

Under ordinary workloads this may mainly appear as duplicate debug output. Under high peer churn, however, it also causes unnecessary repeated processing of already failed I/O objects by the bandwidth and peer-management paths.

It additionally makes diagnostics misleading because callback counts no longer correspond to the number of failed peer sessions.

## Related upstream reports

The same triplicate callback symptom was reported in [transmission/transmission#5287](https://github.com/transmission/transmission/issues/5287). That issue correctly connected the repetitions with the main loop and the two phases of the bandwidth pulse, but it remains open and the underlying problem has not been resolved.

[transmission/transmission#7849](https://github.com/transmission/transmission/pull/7849) attempted to suppress repeated downstream disconnect handling. It was closed without being merged after the discussion concluded that the change primarily addressed the debug-log symptom rather than the underlying repeated I/O processing.

A follow-up report, [transmission/transmission#8683](https://github.com/transmission/transmission/issues/8683), was subsequently opened to track a root-cause fix and then closed as a duplicate of #5287.

Large `ERANGE`/`EMSGSIZE`/`ENOTCONN` log floods were also reported in [transmission/transmission#5829](https://github.com/transmission/transmission/issues/5829). That report may include a separate µTP issue, but it explicitly references the same triplicate TCP callback behavior.

None

_No response_

h1de0x:
Fixes #236.

Peer removal is asynchronous. After a fatal transport error is reported, the bandwidth allocator can revisit the same `tr_peerIo` before the peer manager removes it. The failed socket may therefore invoke the error callback repeatedly during subsequent bandwidth and I/O passes.

This change adds a per-socket fatal-error latch in the common error-callback path. After the first fatal error:

* subsequent error callbacks from the same socket are suppressed;
* further read, write, and flush operations are stopped;
* bandwidth pulses cannot re-enable events on the failed socket.

The latch is reset when a replacement socket is installed, preserving the existing socket-replacement behavior and outgoing µTP-to-TCP fallback.

Detailed investigation and high-load before/after measurements are available in #236.

Notes: Avoid repeated disconnect processing and duplicate debug messages when a peer connection fails.

## Checklist

* [x] I have manually written or manually audited all changes in this PR and vouch for them.
* [x] I have explained user-facing changes in a 'Notes:' paragraph for the release notes script.

### DIFF
--- a/libtransmission/peer-io.cc
+++ b/libtransmission/peer-io.cc
@@ -176,6 +176,7 @@ void tr_peerIo::set_socket(std::shared_ptr<tr_peer_socket> socket_in)
     close(); // tear down the previous socket, if any
 
     socket_ = std::move(socket_in);
+    fatal_error_seen_ = false;
     socket_address_ = socket_->socket_address();
 
     socket_->set_read_cb([weak = weak_from_this()] {
@@ -276,7 +277,7 @@ size_t tr_peerIo::try_write(size_t max)
 {
     static auto constexpr Dir = tr_direction::Up;
 
-    if (max == 0U) {
+    if (max == 0U || fatal_error_seen_) {
         return {};
     }
 
@@ -325,7 +326,7 @@ void tr_peerIo::can_read_wrapper(size_t bytes_transferred)
 {
     // try to consume the input buffer
 
-    if (can_read_ == nullptr) {
+    if (can_read_ == nullptr || fatal_error_seen_) {
         return;
     }
 
@@ -371,7 +372,7 @@ size_t tr_peerIo::try_read(size_t max)
 {
     static auto constexpr Dir = tr_direction::Down;
 
-    if (max == 0U) {
+    if (max == 0U || fatal_error_seen_) {
         return {};
     }
 
@@ -424,9 +425,9 @@ void tr_peerIo::set_enabled(tr_direction dir, bool is_enabled)
     }
 
     if (dir == tr_direction::Up) {
-        socket_->set_write_enabled(is_enabled);
+        socket_->set_write_enabled(is_enabled && !fatal_error_seen_);
     } else {
-        socket_->set_read_enabled(is_enabled);
+        socket_->set_read_enabled(is_enabled && !fatal_error_seen_);
     }
 }
 
--- a/libtransmission/peer-io.h
+++ b/libtransmission/peer-io.h
@@ -144,6 +144,10 @@ public:
 
     size_t flush(tr_direction dir, size_t byte_limit)
     {
+        if (fatal_error_seen_) {
+            return {};
+        }
+
         return dir == tr_direction::Down ? try_read(byte_limit) : try_write(byte_limit);
     }
 
@@ -313,6 +317,13 @@ private:
 
     void call_error_callback(tr_error const& error)
     {
+        // Peer removal is asynchronous. Until it happens, the bandwidth allocator
+        // can try to flush this I/O again, so report only the first fatal error.
+        if (fatal_error_seen_) {
+            return;
+        }
+
+        fatal_error_seen_ = true;
         if (got_error_ != nullptr) {
             got_error_(this, error, user_data_);
         }
@@ -368,6 +379,7 @@ private:
     bool const client_is_seed_;
     bool const is_incoming_;
 
+    bool fatal_error_seen_ = false;
     bool dht_supported_ = false;
     bool extended_protocol_supported_ = false;
     bool fast_extension_supported_ = false;


ckerr:
@h1de0x this looks like a review more for @tearfur anyway but fwiw I'm going to have limited availability for review for a few days, will be back soon.

tearfur (changes requested):
This leaks logic from `tr_peerMsgs` into `tr_peerIo`. `tr_peerIo` should't need to care whether the error callback considers this error fatal, but `tr_peerIo` in this PR is now assuming that it is always fatal.

Something like #239 maintains the encapsulation principles better, and we still get logs for peer I/O errors that are not processed. Wdyt?

h1de0x:
@tearfur I agree with your encapsulation concern. In the current code, however, I couldn't find an error-callback path that can retry on the same socket. `try_read()` and `try_write()` invoke the callback only when `can_retry_from_error()` returns false, while the µTP callbacks report terminal socket conditions. A handshake fallback may continue the connection attempt, but it does so by installing a replacement socket, which resets the per-socket state. So the current behavior is consistent with the existing call sites, even though the callback contract does not make this explicit.

Would an explicit per-socket method for disabling further I/O make that separation clearer? `tr_peerMsgsImpl::got_error()` could call `io->disable_io()` unconditionally before any branch based on `is_disconnecting()`. This matters because the peer may already be disconnecting for an unrelated reason when its first transport error occurs.

`tr_peerIo` would then own only the mechanism: an `io_disabled_` flag preventing further flushes, reads, writes, error callbacks, and event re-enabling. `call_error_callback()` would only check this flag, not set it or classify the error, and `set_socket()` would reset it for a replacement socket.

The measurements in #236 show that this approach reduced fatal-error callbacks to exactly one per affected peer session. I have not repeated those measurements with #239. However, based on its code, I would expect the repeated I/O attempts and callbacks to remain because #239 only prevents repeated peer-level handling.

If this approach sounds reasonable, I can rebase and update #237 on top of #238, incorporating the guard from #239.

tearfur:
Why is there a need to artificially limit fatal error callbacks to one per peer connection? IMO it's overkill for just eliminating duplicated logs.

I'd like to raise another point apart from encapsulation, which is that I want `tr_peerIo` to be transparent about the errors it got from the underlying stack. This makes it easier for me to troubleshoot connection issues in my experience.

If the current way our code works will trigger multiple fatal errors before the connection actually closes, then so be it, `tr_peerIo` will honestly reflect that fact. Owners of `tr_peerIo` object has the freedom to handle this however they want. If there is a bigger problem than just aesthetic issues like duplicated logs, then the underlying issue should be fixed instead.

I acknowledge that this is a balancing act of hiding quirks behind abstractions against keeping information transparent, there is no absolute answer. While this PR leans towards the former approach, I favour the latter in general in this project. The reason being, our classes are meant to be used by ourselves. I know how they work and I want to take advantage of the fact. I don't need to make them pretty for customers.

I opened #238 for a similar reason, which is that I don't like the fact that multiple scenarios are being generalised into a handful of ambiguous errno for the consumer.

h1de0x:
@tearfur Thanks, that clarifies the design preference.

My intention was not to treat the duplicate logs themselves as the underlying problem. The callback counts in #236 were meant to show that the bandwidth allocator can initiate I/O on the same failed socket multiple times before asynchronous peer removal is completed. I noticed this while investigating peer behavior and how the client manages peer connections internally.

However, I agree that suppressing callbacks inside `tr_peerIo` trades away information that may be useful for troubleshooting, which is not the right tradeoff given the transparency you want from this class.

It may still be worth investigating whether the bandwidth or peer-lifecycle code could avoid initiating redundant I/O once the owner has handled a fatal error and scheduled the peer for removal, while keeping `tr_peerIo` transparent about every error it actually receives from the underlying stack. That would address the source of the repeated work without making `tr_peerIo` less transparent.

tearfur:
In any case, thanks for the effort in investigating this. Keep em coming if you like.

PR #241, opened 2026-08-21
fix: acquire lock before calling `tr_torrent::torrents()`

tearfur:
## Description

Regression from #138. CC @ckerr.

Add missing lock in `tr_session::compute_busy_torrent_count` before calling `tr_torrent::torrents()`.

Discovered from CI run https://github.com/retransmission/retransmission/actions/runs/32444522362/job/96661590185?pr=215.

<details>
<summary>fsanitize error</summary>

```
  ==12991==ERROR: AddressSanitizer: heap-use-after-free on address 0x502000032a98 at pc 0x55b861e6c45d bp 0x7f73b45f3e10 sp 0x7f73b45f3e08
  READ of size 8 at 0x502000032a98 thread T1
      #0 0x55b861e6c45c in tr_session::compute_busy_torrent_count() const /home/runner/work/retransmission/retransmission/libtransmission/session.cc:608:32
      #1 0x55b861e6bc81 in tr_session::on_now_timer() /home/runner/work/retransmission/retransmission/libtransmission/session.cc:528:31
      #2 0x55b861eaa83d in tr_session::tr_session(std::basic_string_view<char, std::char_traits<char>>, tr_variant::Map const&)::$_0::operator()() const /home/runner/work/retransmission/retransmission/libtransmission/session.cc:2028:51
      #3 0x55b861eaa7c0 in void std::__invoke_impl<void, tr_session::tr_session(std::basic_string_view<char, std::char_traits<char>>, tr_variant::Map const&)::$_0&>(std::__invoke_other, tr_session::tr_session(std::basic_string_view<char, std::char_traits<char>>, tr_variant::Map const&)::$_0&) /usr/bin/../lib/gcc/x86_64-linux-gnu/14/../../../../include/c++/14/bits/invoke.h:61:14
      #4 0x55b861eaa6d0 in std::enable_if<is_invocable_r_v<void, tr_session::tr_session(std::basic_string_view<char, std::char_traits<char>>, tr_variant::Map const&)::$_0&>, void>::type std::__invoke_r<void, tr_session::tr_session(std::basic_string_view<char, std::char_traits<char>>, tr_variant::Map const&)::$_0&>(tr_session::tr_session(std::basic_string_view<char, std::char_traits<char>>, tr_variant::Map const&)::$_0&) /usr/bin/../lib/gcc/x86_64-linux-gnu/14/../../../../include/c++/14/bits/invoke.h:111:2
[...]

PR #242, opened 2026-08-21
fix: race condition in `tr_sessionGetBusyTorrentCount()`

ckerr:
## Description

Stacked on #250. Supersedes #241, CC @tearfur
Fixes a regression introduced by #138.

Fixes a race condition and improves performance by replacing `tr_sessionGetBusyTorrentCount()` with `tr_sessionIsBusy()`, which answers the question the count existed for -- "should we be inhibiting desktop hibernation?" -- without a session lock and without periodically walking the torrents.

`tr_session` keeps the answer current with relaxed atomics maintained by rare events: counts of started torrents and running verify jobs, an activity stamp on transfer/verify/start, and a busy window that mirrors the queue-stalled settings. `session_stats` gains a `busy` bool so RPC clients relay the daemon's answer instead of re-implementing the policy or comparing the daemon's clock against their own.

This modifies the RPC contract, but it's non-breaking since the modified pieces never shipped.

## Checklist

### DIFF
--- a/libtransmission/peer-mgr.cc
+++ b/libtransmission/peer-mgr.cc
@@ -615,7 +615,7 @@ public:
             break;
 
         case tr_peer_event::Type::ClientGotPieceData:
-            on_client_got_piece_data(s->tor, event.length, tr_time());
+            s->tor->add_downloaded(event.length, tr_time());
             break;
 
         case tr_peer_event::Type::ClientGotHave:
@@ -866,7 +866,7 @@ private:
 
         switch (event.type) {
         case tr_peer_event::Type::ClientGotPieceData:
-            on_client_got_piece_data(s->tor, event.length, tr_time());
+            s->tor->add_downloaded(event.length, tr_time());
             break;
 
         default:
@@ -912,11 +912,6 @@ private:
             TR_ASSERT_MSG(false, "This should be unreachable code");
             break;
         }
-    }
-
-    static void on_client_got_piece_data(tr_torrent* const tor, uint32_t const sent_length, time_t const now)
-    {
-        tor->add_downloaded(sent_length, now);
     }
 
     void on_got_port(tr_peerMsgs* const msgs, tr_peer_event const& event)
--- a/libtransmission/peer-mgr.cc
+++ b/libtransmission/peer-mgr.cc
@@ -608,11 +608,8 @@ public:
         case tr_peer_event::Type::ClientSentPieceData:
             {
                 auto* const tor = s->tor;
-
-                tor->bytes_uploaded_ += event.length;
+                tor->add_uploaded(event.length, tr_time());
                 tr_announcerAddBytes(tor, TR_ANN_UP, event.length);
-                tor->set_date_active(tr_time());
-                tor->session->add_uploaded(event.length);
             }
 
             break;
@@ -919,9 +916,7 @@ private:
 
     static void on_client_got_piece_data(tr_torrent* const tor, uint32_t const sent_length, time_t const now)
     {
-        tor->bytes_downloaded_ += sent_length;
-        tor->set_date_active(now);
-        tor->session->add_downloaded(sent_length);
+        tor->add_downloaded(sent_length, now);
     }
 
     void on_got_port(tr_peerMsgs* const msgs, tr_peer_event const& event)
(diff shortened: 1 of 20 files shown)


tearfur:
Just checking in to acknowledge that I will review this PR, but it's going to take a few days.

ckerr:
I found a regression in this and am moving it to a draft.

I think the basic _shape_ of this is right (keeping an activity date so that we don't need to lock + walk torrents in the upkeep pulse) but it still needs more work. This current iteration @ eef4d58 breaks the keep-awake for stalled torrents.

libtransmission/peer-mgr.cc
```diff
@@ -919,9 +916,7 @@ class tr_swarm
 
     static void on_client_got_piece_data(tr_torrent* const tor, uint32_t const sent_length, time_t const now)
```
tearfur: *(nit)* we can inline this helper.

libtransmission/session.h
```diff
@@ -1431,9 +1473,16 @@ struct tr_session {
...
+    // is_busy() state, readable lock-free from any thread.
+    // The counts are maintained by the torrent start/stop and verify flows;
+    // date_active_ is stamped by transfers (session and web threads),
+    // verified pieces (verify thread), and torrent starts (session thread);
+    // busy_window_ mirrors the queue-stalled settings and is refreshed
+    // once per second in on_now_timer().
+    std::atomic<size_t> n_started_torrents_;
+    std::atomic<size_t> n_verify_jobs_;
+    std::atomic<time_t> date_active_{ 0 };
+    std::atomic<time_t> busy_window_{ 0 };
```
tearfur, line 1485: This value does not change unless session settings `queue_stalled_enabled` or `queue_stalled_minutes` changed. The calculation is trivial too. There's little point in caching the value, and re-computing it every second.

Maybe rename `compute_busy_window` to `busy_window` and call it every time we need this value?

libtransmission/torrent.cc
```diff
@@ -609,6 +609,7 @@ void tr_torrent::start(bool bypass_queue, std::optional<bool> has_any_local_data
     }
 
     is_running_ = true;
+    session->add_started_torrent();
```
tearfur, line 613: Might be a good idea to add `TR_ASSERT(!is_running_)` before `is_running_ = true` to ensure the correctness of this line.

tests/qt/session-test.cc
```diff
@@ -83,6 +83,7 @@ private slots:
     static void initTestCase()
     {
         TR_QT_SKIP_UNLESS_SIGNALS_WORK();
+        qRegisterMetaType<int64_t>("int64_t");
```
tearfur: Looks to be carried over accidentally from an old version of #250.

IIUC we don't need this anymore after #272.

libtransmission/torrent.cc
ckerr: Added in af9b7a468ab787f56a5f5b581520b1e65c229015

libtransmission/peer-mgr.cc
ckerr: I'm not sure it's adding much value, so I tried removing it instead and just calling `tor->add_downloaded()` directly.  59b8b3dd2eb9666b3e8ee7fd9c9aed9bfe79980d

tests/qt/session-test.cc
ckerr: removed in d6ab4f488d62fcda333e696f170dfe537101b2c1

libtransmission/session.h
ckerr: The problem is how to recalculate it on demand without UB? That reintroduces the same kind of problem that we were just trying to solve (though it's a smaller one, since it's a single session field rather than a per-torrent field).

These atomics are invoked with `std::memory_order_relaxed` which basically means "make it as fast as possible, this isn't a critical race, we're just trying to avoid UB".

libtransmission/peer-mgr.cc
tearfur: Yes that's exactly what I meant, not adding the `inline` specifier. 😄

libtransmission/session.h
tearfur: Oooh I missed the fact that calling `tr_session::queueStalledMinutes` can cause a race too. Then yeah I think this is a reasonable way to do it for now.

I think for a future PR, we can have `tr::SessionSettings` store some fields with `std::atomic` and augment the serializer code as such. I will work on it after merging this one.

tearfur (approved):

libtransmission/session.h
tearfur: Oh `std::atomic` is not copy constructable or move constructable, that won't work.

PR #251, opened 2026-08-24
refactor: remove torrents cells from XIB

lolgear:
## Description

Finalizing torrents and group cells extraction from xib.

### DIFF
--- a/macosx/TorrentTableView.mm
+++ b/macosx/TorrentTableView.mm
@@ -141,10 +141,10 @@ static NSTimeInterval const kToggleProgressSeconds = 0.175;
         if (![rowItem isKindOfClass:[TorrentGroup class]]) {
             if (minimal) {
                 SmallTorrentCell* smallCell = [self viewAtColumn:0 row:row makeIfNecessary:NO];
-                [(TorrentCellControlButton*)smallCell.fControlButton resetImage];
+                [smallCell.fControlButton resetImage];
             } else {
                 TorrentCell* torrentCell = [self viewAtColumn:0 row:row makeIfNecessary:NO];
-                [(TorrentCellControlButton*)torrentCell.fControlButton resetImage];
+                [torrentCell.fControlButton resetImage];
             }
         }
     }];
@@ -205,11 +205,10 @@ static NSTimeInterval const kToggleProgressSeconds = 0.175;
 
         TorrentCell* torrentCell;
         if (minimal) {
-            // TODO(lolgear): Remove small torrent cell from xib.
-            torrentCell = [outlineView makeViewWithIdentifier:@"NewSmallTorrentCell" owner:self];
+            torrentCell = [outlineView makeViewWithIdentifier:@"SmallTorrentCell" owner:self];
             if (!torrentCell) {
                 torrentCell = [[SmallTorrentCell alloc] initWithFrame:NSMakeRect(0, 0, NSWidth(outlineView.bounds), self.rowHeight)];
-                torrentCell.identifier = @"NewSmallTorrentCell";
+                torrentCell.identifier = @"SmallTorrentCell";
                 [torrentCell layoutSubtreeIfNeeded];
             }
 
@@ -236,11 +235,10 @@ static NSTimeInterval const kToggleProgressSeconds = 0.175;
                 torrentCell.fRevealButton.hidden = YES;
             }
         } else {
-            // TODO(lolgear): Remove torrent cell from xib.
-            torrentCell = [outlineView makeViewWithIdentifier:@"NewTorrentCell" owner:self];
+            torrentCell = [outlineView makeViewWithIdentifier:@"TorrentCell" owner:self];
             if (!torrentCell) {
                 torrentCell = [[TorrentCell alloc] initWithFrame:NSMakeRect(0, 0, NSWidth(outlineView.bounds), self.rowHeight)];
-                torrentCell.identifier = @"NewTorrentCell";
+                torrentCell.identifier = @"TorrentCell";
                 [torrentCell layoutSubtreeIfNeeded];
             }
 
@@ -250,7 +248,7 @@ static NSTimeInterval const kToggleProgressSeconds = 0.175;
             NSImage* fileImage = torrent.icon;
             if (error) {
                 NSRect frame = torrentCell.fIconView.frame;
-                NSImage* resultImage = [[NSImage alloc] initWithSize:NSMakeSize(frame.size.height, frame.size.width)];
+                NSImage* resultImage = [[NSImage alloc] initWithSize:frame.size];
                 [resultImage lockFocus];
 
                 // draw fileImage
@@ -313,11 +311,10 @@ static NSTimeInterval const kToggleProgressSeconds = 0.175;
         return torrentCell;
     } else {
         TorrentGroup* group = (TorrentGroup*)item;
-        // TODO(lolgear): Remove group cell from xib.
-        GroupCell* groupCell = [outlineView makeViewWithIdentifier:@"NewGroupCell" owner:self];
+        GroupCell* groupCell = [outlineView makeViewWithIdentifier:@"GroupCell" owner:self];
         if (!groupCell) {
             groupCell = [[GroupCell alloc] initWithFrame:NSZeroRect];
-            groupCell.identifier = @"NewGroupCell";
+            groupCell.identifier = @"GroupCell";
         }
 
         NSInteger groupIndex = group.groupIndex;
--- a/macosx/TorrentTableView.mm
+++ b/macosx/TorrentTableView.mm
@@ -213,10 +213,6 @@ static NSTimeInterval const kToggleProgressSeconds = 0.175;
                 [torrentCell layoutSubtreeIfNeeded];
             }
 
-            ((TorrentCellControlButton*)torrentCell.fControlButton).torrentCell = torrentCell;
-            ((TorrentCellRevealButton*)torrentCell.fRevealButton).torrentCell = torrentCell;
-            ((TorrentCellActionButton*)torrentCell.fActionButton).torrentCell = torrentCell;
-
             // set torrent icon or error badge
             torrentCell.fIconView.image = error ? [NSImage imageNamed:NSImageNameCaution] : torrent.icon;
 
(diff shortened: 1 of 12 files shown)


lolgear:
@ckerr 
Let's ask Claude?
image

ckerr:
> @ckerr Let's ask Claude?

5732AFABE758958935DD9E61532DCFA7

ckerr:
Me: "looks OK to me." 

Claude: "lol no"

Review is all-Claude this time.

One question I have about its findings: The strings overwritten each time we update translations from Transifex, so don't bother removing dead translations. But ... why do we have dead translations now? Are there strings that used to be translated and now aren't but should be...? If so, that sounds like a problem on its own

---

# PR #251 review — findings

Two correctness issues (both pre-existing, but inside the method this PR edits) and six
cleanups directly actionable in this PR. One candidate was investigated and dropped: the
`setButtonType:` call in `configureViews` does clobber the action button's
`setHighlightsBy:NSNoCellMask`, but that ordering is identical before and after this PR —
the deleted xib prototypes were already unreachable (the code requests `New*` reuse
identifiers), so the PR deletes only dead code there and changes no observable behavior.

---

## `macosx/TorrentTableView.mm:224` — reused cell keeps the prior row's hover state *(pre-existing, in touched method)*

In the small-view branch of `outlineView:viewForTableColumn:item:`, the reset of
`fTorrentStatusField.hidden` / `fControlButton.hidden` / `fRevealButton.hidden` lives only in
the `else` of `if (self.fHoverEventDict)`. When the dict is non-nil but the row being prepared
is not the hovered row, the inner `if (row == hoverRow)` has no `else`, so nothing is written
and a cell recycled from a previously hovered row keeps its hover UI. No `prepareForReuse`
override exists to compensate.

Repro: in SmallView, hover row A (status field hides, control/reveal buttons appear), then
scroll or otherwise trigger `reloadVisibleRows` while the hover dict is still set; the recycled
cell shows pause/reveal buttons and no status text on a non-hovered row.

Fix: give the inner `if` an `else` that mirrors the outer `else`'s three assignments (or move
the reset before the hover check so every pass starts from the non-hovered state).

---

## `macosx/TorrentTableView.mm:208` — completed "Remove … from xib" TODOs left in place

The three comments `TODO(lolgear): Remove small torrent cell from xib.` (208),
`Remove torrent cell from xib.` (239), and `Remove group cell from xib.` (316) describe exactly
the removal this PR performs. At PR head they point at xib prototypes that no longer exist.
Delete all three in this PR.

---

## `macosx/*.lproj/MainMenu.strings` — orphaned keys for deleted xib objects

All 21 localized `MainMenu.strings` files still carry translation entries keyed to object IDs
this PR deletes from `MainMenu.xib` (e.g. `"zIM-6D-jsa.title"` for the removed Group Title
field, plus `x8g-Nv-jcD`, `HmT-FI-amY`, `y9O-gd-sXA`, `oSz-TE-8jj`, `lmp-Y3-ZQO`, `kbX-Av-eP4`,
`Zmj-A9-NPf`). Runtime-inert — unmatched keys are never looked up — but nothing in CI
regenerates these files (translations sync manually via Transifex), so the stale keys sit in
the repo producing audit/ibtool noise until removed. Dropping the orphaned entries belongs in
this PR alongside the xib deletion.

---

## `macosx/SmallTorrentCell.mm:28` — layout constants duplicated from `TorrentCell.mm`

`SmallTorrentCell.mm` redefines `kActionButtonSize` (16), `kPriorityViewSize` (12),
`kButtonSize` (14), and `kButtonsSpacing` (3) with the same names and values as its superclass
file `TorrentCell.mm`. Both cells share the same hover buttons and priority icon, so these are
one set of design values in two files with nothing tying them together — resizing a button in
one file silently leaves the other view mode at the old value. Hoist the shared metrics next to
`TorrentCell` (shared header or base-class constants) and keep only the genuinely different
values (icon size, indicator size, offsets) file-local.

---

## `macosx/TorrentTableView.mm:209` — the "New" reuse-identifier prefix is now meaningless

`@"NewSmallTorrentCell"` (209, 212), `@"NewTorrentCell"` (240, 243), and `@"NewGroupCell"`
(317, 320) existed to avoid colliding with the xib prototypes' unprefixed identifiers, which
this PR deletes. The prefix now distinguishes against nothing and permanently misnames the
cells, and each identifier is a bare literal spelled twice — a typo in one of the six silently
defeats reuse (`makeViewWithIdentifier:` returns nil and every row allocates a fresh cell).
Nothing else in the repo claims the unprefixed names and nothing persists these strings, so
renaming to `SmallTorrentCell`/`TorrentCell`/`GroupCell` — ideally one named constant per cell
type — is behavior-neutral.

---

## `macosx/TorrentCell.h:10` — button properties typed `NSButton*`, forcing casts

`fActionButton`/`fControlButton`/`fRevealButton` stay declared as plain `NSButton*` (this PR
touches exactly these lines to strip `IBOutlet`), yet `configureViews` now always constructs
the concrete `TorrentCellActionButton`/`ControlButton`/`RevealButton` classes, and
`TorrentTableView.mm:144`/`147` must cast to call `resetImage`. With the nib-injection path
gone, nothing can assign a different class, so retyping the three properties to their concrete
classes removes both existing casts and the ones this shape will keep attracting. `TorrentCell.h`
needs only the three button imports (or forward declarations); `TorrentTableView.mm` already
imports all three headers.

---

## `macosx/TorrentCell.mm:24` — multiple constants encode one design value

`kStackViewToProgressFieldSpacing`, `kProgressFieldToProgressBarSpacing`, and
`kProgressBarToStatusFieldSpacing` (24–26) are all the same 1.0 row gap; changing the cell's
row spacing takes three coordinated edits. Likewise `SmallTorrentCell.mm:26-27`:
`kStatusFieldTrailingOffset` and `kRevealButtonTrailingOffset` are the same −3.0 offset against
the same trailing anchor, and the status field and reveal button swap at that position on
hover, so the two values must stay equal — two independently editable constants hide that
invariant. One `kRowSpacing` and one shared trailing-offset constant express the intent.

---

## `macosx/TorrentTableView.mm:253` — badge image size transposes width and height *(pre-existing, in touched method; latent)*

The error-badge canvas is created with `NSMakeSize(frame.size.height, frame.size.width)` —
arguments transposed from the icon view's frame. Invisible today because the only path reaching
this code pins the icon square (36×36), but if the icon dimensions ever diverge, a torrent with
an error gets a badge image with swapped dimensions, clipping or misplacing the composited
icon-plus-caution-badge. Swap the arguments while the method is being touched.

lolgear:
@ckerr

Before involving into this "very productive" dialogue with Claude ( ironically and unironically productive ), could you add Claude as agent to repository?

In Swift repository smoke-test bot exists which you can call by (at)smoke-test or something like this and ask him for running tests.

I wish agents or GitHub Actions have this functionality but for Claude or any LLM.

Maybe not as "agent" but as a bot..

ckerr:
> could you add Claude as agent to repository?

@lolgear how do I do that...?

lolgear:
I asked Gemini ( of course I asked A.I. for this question ) and it gives three options.

Silence!
Gemini speaks:

Here is a brief summary of the three options to add an AI agent or bot to your GitHub repository:

## 1. Built-in GitHub Copilot Agents (Native & Context-Aware)
If you already use GitHub Copilot, you can create a custom agent directly in your repo. You simply create a markdown file at .github/agents/vite-reviewer.agent.md and define its rules and system prompt inside. It will automatically understand your Vite configuration, and you can invoke it in GitHub or VS Code to review PRs or generate code. Alternatively, you can use a .github/copilot-instructions.md file to feed specific guidelines to the native Copilot Code Review feature.

## 2. GitHub Apps / Marketplace Bots (Fastest Setup)
You can install a ready-made, autonomous AI bot from the GitHub Marketplace (such as CodeRabbit, Qodo/CodiumAI, or Sweep). After granting it access to your repository, you configure its behavior via a YAML file (e.g., .coderabbit.yaml). The bot will automatically trigger on every Pull Request, leave line-by-line comments, and can even commit fixes directly for your Vite project.

## 3. GitHub Actions with Custom Scripts (Full Control & Flexibility)
If you want to use your own API keys (OpenAI, Anthropic, etc.) and have complete control over the logic, you can set up a custom GitHub Actions workflow. By creating a .github/workflows/ai-review.yml file, you can trigger a review every time a PR is opened or updated. You can use popular open-source actions like anc95/ChatGPT-Code-Review and pass your own custom system prompt tailored specifically for Vite and TypeScript.

ckerr:
None of those appear to be free?

I'm paying for my own Claude and am already regularly hitting usage limits just on my own code. I don't want to plug that into GitHub! :laughing:

lolgear:
@ckerr
Now we know that Opensource is dead for robots.

lolgear:
@ckerr 
Yeah, another one..
If somebody touches resources with localization ( XIB or localized resources as Localizable.strings ), then anything on CI checks that these resources are fine?

Group and Torrent (cells/views) have dynamic localization, it makes sense.
But CI script could check it.

macosx/TorrentTableView.mm
ckerr: This TODO item is done now & can be removed

```suggestion
```

ckerr: https://developer.apple.com/documentation/foundation/nsmakesize(_:_:)?language=objc says the first arg is width, the second is height. This pre-existing code has them in the wrong order.

```diff
- NSImage* resultImage = [[NSImage alloc] initWithSize:NSMakeSize(frame.size.height, frame.size.width)];
+ NSImage* resultImage = [[NSImage alloc] initWithSize:frame.size];
```

Not new to this PR. Introduced in transmission/transmission/pull/5147. 

@lolgear please fix this either in this PR or in a followup, and also open a bugfix PR in transmission/transmission since this bug exists upstream as well.

lolgear: Let's fix it in next PR.

lolgear: Ok, done in this PR.

lolgear: Oh no, I forgot :(
Ok, done in this PR.

ckerr: Either way is fine with me. Thanks for the fix.

Please do open an upstream fix, transmission/transmission needs this one too :bowing_man:

ckerr (approved):

lolgear:
bugifx
https://github.com/transmission/transmission/pull/9071

PR #253, opened 2026-08-24
perf: hoist `tr_time()`

ckerr:
## Description

In cases where `tr_time()` is being called inside a loop, or back-to-back, hoist it into a local variable so there's only one call instead of N. This is consistent with how the rest of the codebase hoists the current time into a local variable. It's also a small performance win, though `tr_time()` is pretty inexpensive already.

## Checklist

### DIFF
--- a/tests/qt/session-test.cc
+++ b/tests/qt/session-test.cc
@@ -83,6 +83,7 @@ private slots:
     static void initTestCase()
     {
         TR_QT_SKIP_UNLESS_SIGNALS_WORK();
+        qRegisterMetaType<int64_t>("int64_t");
     }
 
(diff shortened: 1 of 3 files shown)


tearfur:
@ckerr Merge conflicts.

tests/qt/session-test.cc
```diff
@@ -83,6 +83,7 @@ private slots:
     static void initTestCase()
     {
         TR_QT_SKIP_UNLESS_SIGNALS_WORK();
+        qRegisterMetaType<int64_t>("int64_t");
```
tearfur, line 86: Ah, this line got in in the end. 😅

PR #254, opened 2026-08-24
fix: apply honor-session-limits change to both directions

jgalat:
## Description

`tr_torrentUseSessionLimits()` used short-circuit evaluation to combine two setter calls. When the upload setting changed, the download setter was skipped, leaving the torrent unexpectedly subject to the session download limit.

Evaluate both directions before checking whether either changed, and add regression coverage for disabling and re-enabling the limits.

This is a parallel port of transmission/transmission#8960.

Notes: Fixed per-torrent session speed-limit overrides so they apply to both downloads and uploads.

## Checklist

### DIFF
--- a/tests/libtransmission/torrent-test.cc
+++ b/tests/libtransmission/torrent-test.cc
@@ -7,7 +7,7 @@
 #include <cstddef>
 #include <ranges>
 
-#include <libtransmission/transmission.h>
+#include <libtransmission/types.h>
 
 #include <libtransmission/torrent.h>
 
--- a/tests/libtransmission/torrent-test.cc
+++ b/tests/libtransmission/torrent-test.cc
@@ -7,9 +7,8 @@
 #include <cstddef>
 #include <ranges>
 
-#include <libtransmission/types.h>
-
 #include <libtransmission/torrent.h>
+#include <libtransmission/types.h>
 
 #include "test-fixtures.h"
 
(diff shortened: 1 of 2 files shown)


ckerr:
argh cannot click correctly today, mean to assign co-review to me and to @tearfur

tests/libtransmission/torrent-test.cc
```diff
@@ -7,6 +7,8 @@
 #include <cstddef>
 #include <ranges>
 
+#include <libtransmission/transmission.h>
```
tearfur: Include `libtransmission/types.h` instead.

jgalat: fixed on 89c4de1

tests/libtransmission/torrent-test.cc
```diff
@@ -7,6 +7,8 @@
 #include <cstddef>
 #include <ranges>
 
+#include <libtransmission/types.h>
```
tearfur: Our usual convention will order this line below `#include <libtransmission/torrent.h>`, please do that.

Other than that the PR looks good, nice catch!

tearfur (approved):

tests/libtransmission/torrent-test.cc
jgalat: thanks! fixed on c6a39d0

PR #256, opened 2026-08-24
fix: quit hangs during UPnP gateway validation

jjoanna2-debug:
## Description

`UPNP_GetValidIGD()` can block while fetching a router's device description. Retransmission currently runs that call on the session thread after discovery finishes, so quitting during validation can leave the app waiting on router HTTP or DNS.

This moves discovery and validation into the existing worker. The worker returns an owned result: successful URL ownership transfers once, while abandoned results clean themselves up. The debug log also keeps miniupnpc's version-specific validation status and distinguishes a failed discovery from a failed validation.

This is a parallel port of [transmission/transmission#8984](https://github.com/transmission/transmission/pull/8984), opened after [ckerr suggested trying the fix in Retransmission](https://github.com/transmission/transmission/pull/8984#issuecomment-5397509871). The symbolicated stack in [transmission/transmission#4759](https://github.com/transmission/transmission/issues/4759) identifies `UPNP_GetValidIGD()` as the blocking quit path.

Notes: Fixed a quit hang during UPnP gateway validation.

## Verification

- `uvx --from 'clang-format~=22.0' clang-format --dry-run --Werror libtransmission/port-forwarding-upnp.cc`
- `git diff --check`
- Fresh CMake/Ninja configuration and `all-tests` build completed.
- 805/805 enabled non-blocklist tests passed; the repository-disabled tests and one locale-dependent skip were unchanged.
- The five `LT.BlocklistDownloadTest.*` loopback-download failures returned `No Response (0)` on both this branch and an untouched `upstream/main` worktree.
- The final source compiled against miniupnpc 2.2.1/API 17 and the vendored API 21 headers.

## Checklist

- [x] I have explained user-facing changes in a `Notes:` paragraph above.

### DIFF
--- a/libtransmission/port-forwarding-upnp.cc
+++ b/libtransmission/port-forwarding-upnp.cc
@@ -39,6 +39,7 @@
 namespace
 {
 // miniupnpc added named IGD statuses in 2.3.0. Keep supported 2.2.x system builds compatible.
+// NOLINTBEGIN(readability-identifier-naming)
 #if !defined(UPNP_NO_IGD)
 inline constexpr auto UPNP_NO_IGD = 0;
 inline constexpr auto UPNP_CONNECTED_IGD = 1;
@@ -51,6 +52,7 @@ inline constexpr auto UPNP_DISCONNECTED_IGD = 2;
 inline constexpr auto UPNP_UNKNOWN_DEVICE = 3;
 #endif
 #endif
+// NOLINTEND(readability-identifier-naming)
 
 enum class UpnpState : uint8_t {
     Idle,
--- a/libtransmission/port-forwarding-upnp.cc
+++ b/libtransmission/port-forwarding-upnp.cc
@@ -8,8 +8,10 @@
 #include <chrono>
 #include <cstdint>
 #include <future>
+#include <memory>
 #include <optional>
 #include <string>
+#include <string_view>
 #include <thread>
 #include <utility>
 
@@ -36,12 +38,46 @@
 
 namespace
 {
+// miniupnpc added named IGD statuses in 2.3.0. Keep supported 2.2.x system builds compatible.
+#if !defined(UPNP_NO_IGD)
+inline constexpr auto UPNP_NO_IGD = 0;
+inline constexpr auto UPNP_CONNECTED_IGD = 1;
+#if (MINIUPNPC_API_VERSION >= 18)
+inline constexpr auto UPNP_PRIVATEIP_IGD = 2;
+inline constexpr auto UPNP_DISCONNECTED_IGD = 3;
+inline constexpr auto UPNP_UNKNOWN_DEVICE = 4;
+#else
+inline constexpr auto UPNP_DISCONNECTED_IGD = 2;
+inline constexpr auto UPNP_UNKNOWN_DEVICE = 3;
+#endif
+#endif
+
 enum class UpnpState : uint8_t {
     Idle,
     WillDiscover, // next action is upnpDiscover()
-    Discovering, // currently making blocking upnpDiscover() call in a worker thread
+    Discovering, // currently discovering and validating an IGD in a worker thread
     WillMap, // next action is UPNP_AddPortMapping()
     WillUnmap // next action is UPNP_DeletePortMapping()
+};
+
+struct UpnpDiscoveryResult {
+    UpnpDiscoveryResult() = default;
+    UpnpDiscoveryResult(UpnpDiscoveryResult&&) = delete;
+    UpnpDiscoveryResult(UpnpDiscoveryResult const&) = delete;
+    UpnpDiscoveryResult& operator=(UpnpDiscoveryResult&&) = delete;
+    UpnpDiscoveryResult& operator=(UpnpDiscoveryResult const&) = delete;
+
+    ~UpnpDiscoveryResult()
+    {
+        FreeUPNPUrls(&urls);
+    }
+
+    UPNPUrls urls = {};
+    IGDdatas data = {};
+    std::array<char, TrAddrStrlen> lanaddr = {};
+    int error = 0;
+    int igd_status = UPNP_NO_IGD;
+    bool did_discover_device = false;
 };
 } // namespace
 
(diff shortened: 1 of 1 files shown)


jjoanna2-debug:
@tearfur, this is the focused one-file port of transmission/transmission#8984 onto current main. Could you take a look when you have time?

libtransmission/port-forwarding-upnp.cc
```diff
@@ -195,14 +216,77 @@ void tr_upnpDeletePortMapping(tr_upnp const* handle, char const* proto, tr_port
...
+    UPNP_IGD_VALID_CONNECTED = 1,
+#if (MINIUPNPC_API_VERSION >= 18)
+    UPNP_IGD_VALID_PRIVATE = 2,
+    UPNP_IGD_VALID_NOT_CONNECTED = 3,
+    UPNP_IGD_INVALID = 4
+#else
+    UPNP_IGD_VALID_NOT_CONNECTED = 2,
+    UPNP_IGD_INVALID = 3
+#endif
+};
```
ckerr: This version of the enum is clearly better than what was there before, since it fixes a bug that gave the wrong enum value for miniupnpc >= 18. (And the vendored version in third-party is 21 💀)

I wonder if this change just hits reset on the same problem though. It can break again the next time upstream's enum values change.

Would we lose anything by dropping the enum and using `UPNP_NO_IGD`, `UPNP_CONNECTED_IGD`, `UPNP_PRIVATEIP_IGD`, etc directly?

libtransmission/port-forwarding-upnp.cc
```diff
@@ -195,14 +216,77 @@ void tr_upnpDeletePortMapping(tr_upnp const* handle, char const* proto, tr_port
...
 
-auto* discover_thread_func(std::string bindaddr) // NOLINT performance-unnecessary-value-param
+char const* igd_status_string(int status)
```
ckerr: _(nit)_

- House style to use `noexcept`, `constexpr` and to define args as `const` when they are unmodified
- Prefer `std::string_view` over raw `char*` so {fmt} doesn't have to recalcualte string lengths

```suggestion
[[nodiscard]] constexpr std::string_view igd_status_string(int const status) noexcept
```

libtransmission/port-forwarding-upnp.cc
```diff
@@ -195,14 +216,77 @@ void tr_upnpDeletePortMapping(tr_upnp const* handle, char const* proto, tr_port
...
+auto discover_and_validate_thread_func(std::string bindaddr) // NOLINT performance-unnecessary-value-param
 {
     // If multicastif is not NULL, it will be used instead of the default
     // multicast interface for sending SSDP discover packets.
     char const* multicastif = std::empty(bindaddr) ? nullptr : bindaddr.c_str();
-    return upnp_discover(2000, multicastif);
+    auto* const devlist = upnp_discover(2000, multicastif);
+    auto result = std::make_unique<UpnpDiscoveryResult>();
+    result->did_discover_device = devlist != nullptr;
+    if (devlist != nullptr) {
```
tearfur: *(nit)*

```suggestion
    if (result->did_discover_device) {
```

libtransmission/port-forwarding-upnp.cc
tearfur: Looks to me we should use miniupnpc's macro constants instead.

jjoanna2-debug: Updated in 32718afff to use miniupnpc’s named constants directly. Retransmission still supports miniupnpc 2.2.1/API 17, which predates those macros, so the file defines a version-aware fallback only when the header does not provide them. I compiled the final source against both 2.2.1 and the vendored API 21 headers.

libtransmission/port-forwarding-upnp.cc
jjoanna2-debug: Applied the suggested signature in 32718afff.

libtransmission/port-forwarding-upnp.cc
jjoanna2-debug: Changed the condition to use `result->did_discover_device` in 32718afff.

tearfur:
@jjoanna2-debug Got some clang-tidy warnings to silence. Just surround these variables with `// NOLINTBEGIN(readability-identifier-naming)` and  `// NOLINTEND(readability-identifier-naming)`.

```
  /home/runner/work/retransmission/retransmission/libtransmission/port-forwarding-upnp.cc:43:23: error: invalid case style for constexpr variable 'UPNP_NO_IGD' [readability-identifier-naming,-warnings-as-errors]
     43 | inline constexpr auto UPNP_NO_IGD = 0;
        |                       ^~~~~~~~~~~
        |                       UpnpNoIgd
  /home/runner/work/retransmission/retransmission/libtransmission/port-forwarding-upnp.cc:44:23: error: invalid case style for constexpr variable 'UPNP_CONNECTED_IGD' [readability-identifier-naming,-warnings-as-errors]
     44 | inline constexpr auto UPNP_CONNECTED_IGD = 1;
        |                       ^~~~~~~~~~~~~~~~~~
        |                       UpnpConnectedIgd
  /home/runner/work/retransmission/retransmission/libtransmission/port-forwarding-upnp.cc:50:23: error: invalid case style for constexpr variable 'UPNP_DISCONNECTED_IGD' [readability-identifier-naming,-warnings-as-errors]
     50 | inline constexpr auto UPNP_DISCONNECTED_IGD = 2;
        |                       ^~~~~~~~~~~~~~~~~~~~~
        |                       UpnpDisconnectedIgd
  /home/runner/work/retransmission/retransmission/libtransmission/port-forwarding-upnp.cc:51:23: error: invalid case style for constexpr variable 'UPNP_UNKNOWN_DEVICE' [readability-identifier-naming,-warnings-as-errors]
     51 | inline constexpr auto UPNP_UNKNOWN_DEVICE = 3;
        |                       ^~~~~~~~~~~~~~~~~~~
        |                       UpnpUnknownDevice
```

jjoanna2-debug:
Fixed in 99031feb8: wrapped the miniupnpc 2.2.x compatibility constants in the requested readability-identifier-naming suppression. The targeted clang-tidy 22 check passes locally.

tearfur (approved):

tearfur:
Noting that this PR also handles cases where `upnpDiscover()` returns `nullptr`, which we didn't handle before.

...How did we not get SEGV reports in these parts of the code before?

jjoanna2-debug:
miniupnpc is why: `UPNP_GetValidIGD()` returns `UPNP_NO_IGD` immediately for a null device list, and `freeUPNPDevlist()` is null-safe too. So the old path avoided a segfault through those library checks. This change avoids calling validation in that case and logs “no discovery response” separately from an IGD validation failure.

ckerr (approved):

PR #261, opened 2026-08-25
refactor: migrate drag overlay view to appkit views.

lolgear:
## Description
This PR refactors `DragOverlayView` by migrating it from old-school manual pixel drawing (`drawRect:` with `lockFocus`) to a clean, modern architecture using core AppKit subviews (`NSImageView`, `NSTextField`) and Auto Layout. 

## Motivation and Context
* **Deprecated API Removal:** Removed `lockFocus`, which has been deprecated by Apple and causes rendering issues on newer macOS versions.
* **Performance Improvement:** Replaced CPU-bound manual image drawing with GPU-accelerated Core Animation layers (`wantsLayer = YES`).
* **Code Simplification:** Removed manual string size calculations (`sizeWithAttributes:`) and hardcoded frame math. Layout is now fully handled by declarative Auto Layout constraints.

## Changes
* Removed `fBadge` image generation and custom `drawRect:` logic.
* Removed `NSShadow` properties to clean up view rendering.
* Configured flexible text constraints ensuring proper alignment and native middle-truncation (`NSLineBreakByTruncatingMiddle`).


macosx/DragOverlayView.mm
```diff
@@ -21,81 +18,86 @@ @implementation DragOverlayView
...
-    [self.fBadge unlockFocus];
-
-    self.needsDisplay = YES;
+    // Activate Auto Layout constraints
+    [NSLayoutConstraint activateConstraints:@[
+        // Center the badgeContainer within DragOverlayView
+        [self.badgeContainer.centerXAnchor constraintEqualToAnchor:self.centerXAnchor],
+        [self.badgeContainer.centerYAnchor constraintEqualToAnchor:self.centerYAnchor],
+
+        // Fixed badge dimensions from the original spec
```
ckerr:
```suggestion
        // Fixed badge dimensions
```

macosx/DragOverlayView.mm
```diff
@@ -4,15 +4,12 @@
...
 @interface DragOverlayView ()
 
-@property(nonatomic) NSImage* fBadge;
-
-@property(nonatomic, readonly) NSDictionary* fMainLineAttributes;
-@property(nonatomic, readonly) NSDictionary* fSubLineAttributes;
+@property(nonatomic, strong) NSView* badgeContainer;
+@property(nonatomic, strong) NSImageView* iconImageView;
+@property(nonatomic, strong) NSTextField* titleLabel;
+@property(nonatomic, strong) NSTextField* subtitleLabel;
```
ckerr: macosx/ has a convention of fVariableName and we should follow that here.

Also these are all set in the init func and never reassigned, so we can promote from strong to readonly:

```Objc++
@property(nonatomic, readonly) NSView* fBadgeContainer;
@property(nonatomic, readonly) NSImageView* fIconImageView;
@property(nonatomic, readonly) NSTextField* fTitleLabel;
@property(nonatomic, readonly) NSTextField* fSubtitleLabel;
```

macosx/DragOverlayView.mm
```diff
@@ -4,15 +4,12 @@
 
 #import "DragOverlayView.h"
 
-static CGFloat const kPadding = 10.0;
-static CGFloat const kIconWidth = 64.0;
```
ckerr, line 8: The named layout constants kPadding and kIconWidth were deleted and their values inlined as repeated literals in the constraint list. `10.0` now appears twice with opposite signs (lines 77 and 84), `64.0` twice (79, 80), and `2.0` twice for two unrelated offsets (86, 92). 

Please use symbolic constants for these instead of hardcoding magic values 👍

macosx/DragOverlayView.mm
```diff
@@ -21,81 +18,86 @@ @implementation DragOverlayView
...
+        [self.iconImageView.leadingAnchor constraintEqualToAnchor:self.badgeContainer.leadingAnchor constant:10.0],
+        [self.iconImageView.centerYAnchor constraintEqualToAnchor:self.badgeContainer.centerYAnchor],
+        [self.iconImageView.widthAnchor constraintEqualToConstant:64.0],
+        [self.iconImageView.heightAnchor constraintEqualToConstant:64.0],
+
+        // Main label layout constraints (spacing from icon: 5.0, padding right: 10.0)
+        [self.titleLabel.leadingAnchor constraintEqualToAnchor:self.iconImageView.trailingAnchor constant:5.0],
+        [self.titleLabel.trailingAnchor constraintEqualToAnchor:self.badgeContainer.trailingAnchor constant:-10.0],
+        // Position the main label slightly above the badge center line
+        [self.titleLabel.bottomAnchor constraintEqualToAnchor:self.badgeContainer.centerYAnchor constant:2.0],
```
ckerr: The comment states the main label is placed above the badge center line, but a positive constant on bottomAnchor places its bottom edge below the center.


(claude)

`[self.titleLabel.bottomAnchor constraintEqualToAnchor:self.badgeContainer.centerYAnchor constant:2.0]` moves the label's bottom edge 2 points *down* from the center — AppKit Auto Layout is flip-agnostic, which macosx/TorrentCell.mm:29 makes explicit (`static CGFloat const kStatusFieldBottomOffset = -1.0; // inverted for constraints.`). The constant is right for reproducing the old geometry, but the comment on line 85 says the opposite direction, so the next person to nudge the layout will move it the wrong way or 'fix' the sign and shift the whole two-line block.

ckerr (changes requested):

macosx/DragOverlayView.mm
lolgear: If Retransmission is a new modern project then it is better to follow modern Apple Standards.

Gemini:
> * Legacy C++ / MacApp: Early Macintosh frameworks written in C++ used f to denote a class field, distinguishing it from local variables or parameters.

> * Objective-C (Older Codebases): Some developers carried this habit into Objective-C to differentiate instance variables (fName) from local variables (name), though the community eventually standardized on using an underscore prefix (_name) for instance variables.

> * Modern Apple Standards (Swift & Objective-C): Apple's official guidelines mandate standard camelCase for variables and properties, leveraging self.variableName when disambiguation from local parameters is required.

macosx/DragOverlayView.mm
lolgear: Done

macosx/DragOverlayView.mm
```diff
@@ -21,81 +18,86 @@ @implementation DragOverlayView
...
+        // 3. Title
+        _titleLabel = [[NSTextField alloc] initWithFrame:NSZeroRect];
+        _titleLabel.translatesAutoresizingMaskIntoConstraints = NO;
+        _titleLabel.editable = NO;
+        _titleLabel.selectable = NO;
+        _titleLabel.drawsBackground = NO;
+        _titleLabel.bordered = NO;
+        _titleLabel.font = [NSFont boldSystemFontOfSize:18.0];
+        _titleLabel.textColor = NSColor.whiteColor;
+        _titleLabel.lineBreakMode = NSLineBreakByTruncatingMiddle;
```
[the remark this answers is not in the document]
lolgear: Added.

macosx/DragOverlayView.mm
```diff
@@ -21,81 +18,86 @@ @implementation DragOverlayView
...
+
+        // 3. Title
+        _titleLabel = [[NSTextField alloc] initWithFrame:NSZeroRect];
+        _titleLabel.translatesAutoresizingMaskIntoConstraints = NO;
+        _titleLabel.editable = NO;
+        _titleLabel.selectable = NO;
+        _titleLabel.drawsBackground = NO;
+        _titleLabel.bordered = NO;
+        _titleLabel.font = [NSFont boldSystemFontOfSize:18.0];
+        _titleLabel.textColor = NSColor.whiteColor;
```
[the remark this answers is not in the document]
lolgear: NSShadow was deprecated and "removed" in iOS 7.
Be modern, Copilot.

macosx/DragOverlayView.mm
lolgear: Done everything in StackViews with minimum usage of layout constraints.

lolgear:
@ckerr 
Could you check it?

lolgear:
@nevack @Coeur 
Could you check it?

macosx/DragOverlayView.mm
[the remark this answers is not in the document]
ckerr: (opus 5 extra re-review)

NSShadow is a current AppKit class with no deprecation — and this is a macOS-only file, so an iOS deprecation wouldn't govern it either way. What iOS 7 actually changed was the visual idiom (and it deprecated the old UITextAttributeTextShadow* dictionary keys, superseded by NSShadowAttributeName, which is itself still current). "We prefer the flatter look" is a perfectly good reason to drop it; "the API is gone" isn't, and the distinction matters because the first one still needs a Notes: line.

macosx/DragOverlayView.mm
```diff
@@ -4,15 +4,28 @@
...
-static CGFloat const kIconWidth = 64.0;
+// Layout
 
-@interface DragOverlayView ()
+// Badge
+static CGFloat const kBadgeWidth = 325.0;
+static CGFloat const kBadgeHeight = 84.0;
+
+// MainStack
+static CGFloat const kMainStackInset = 10;
```
ckerr: Two values escaped the new constants block: `kMainStackInset = 10` drops the decimal that every sibling carries (`325.0`, `84.0`, `5.0`, `64.0`, `2.0`), and `cornerRadius = 15.0` at line 42 is the one geometry number left inline — so the badge's size is at line 10 and its radius 30 lines away.

```suggestion
static CGFloat const kMainStackInset = 10.0;
```

macosx/DragOverlayView.mm
```diff
@@ -4,15 +4,28 @@
...
 
-@property(nonatomic) NSImage* fBadge;
+// TextStack
+static CGFloat const kTextStackSpacing = 2.0;
+
+@interface DragOverlayView ()
 
-@property(nonatomic, readonly) NSDictionary* fMainLineAttributes;
-@property(nonatomic, readonly) NSDictionary* fSubLineAttributes;
+@property(nonatomic, readonly) NSView* badgeContainer;
```
ckerr: `badgeContainer` is only ever touched inside `initWithFrame:` — `setOverlay:` uses just the icon and the two labels. So the property synthesizes a getter nothing calls, and advertises an instance-lifetime relationship the class no longer has.

A local `NSView* badgeContainer` in the initializer says what's true. The other three genuinely outlive init and should stay as they are.

macosx/DragOverlayView.mm
```diff
@@ -21,81 +34,90 @@ @implementation DragOverlayView
...
-            NSParagraphStyleAttributeName : paragraphStyle
-        };
-
-        _fSubLineAttributes = @{
-            NSForegroundColorAttributeName : NSColor.whiteColor,
-            NSFontAttributeName : smallFont,
-            NSShadowAttributeName : stringShadow,
-            NSParagraphStyleAttributeName : paragraphStyle
-        };
+        // 1. Background badge container
```
ckerr: The `// 1.` … `// 7.` headers are now carrying structure that method boundaries used to: `setupConstraints` was folded back into the initializer, which is 78 lines of view creation, configuration and constraint activation.

`TorrentCell` — which this file now mirrors closely, down to the constants block — splits the same work three ways (`configureViews`, `configurePriorities`, `setupConstraints`, declared at TorrentCell.h:32-34). Doing that here would let the numbered headers go away rather than grow with each new view.

Per CLAUDE.md, the headers that restate their own line (`// 2. Icon` above `_iconImageView = ...`) cost every future reader something and give nothing back. The `edgeInsets` comment at line 87 is the opposite kind and worth keeping — that AppKit's stack-view insets push inward is a real API fact you can't read off the code.

macosx/DragOverlayView.mm
```diff
@@ -21,81 +34,90 @@ @implementation DragOverlayView
...
+        // 4. Subtitle
+        _subtitleLabel = [[NSTextField alloc] initWithFrame:NSZeroRect];
+        _subtitleLabel.editable = NO;
+        _subtitleLabel.selectable = NO;
+        _subtitleLabel.drawsBackground = NO;
+        _subtitleLabel.bordered = NO;
+        _subtitleLabel.font = [NSFont systemFontOfSize:14.0];
+        _subtitleLabel.textColor = NSColor.whiteColor;
+        _subtitleLabel.maximumNumberOfLines = 1;
+        _subtitleLabel.lineBreakMode = NSLineBreakByTruncatingMiddle;
```
ckerr: These two blocks differ only in their font, so every other property has to be kept in sync by hand — `maximumNumberOfLines` just had to be added to both.

`+[NSTextField labelWithString:]` (macOS 10.12+; the project floor is 12.0 per CMakeLists.txt:10) covers `editable`, `selectable`, `drawsBackground` and `bordered` in one call, and additionally clears `bezeled` — a separate `NSCell` flag from `bordered` that nothing here sets. The shared remainder fits the loop idiom TorrentCell.mm:86 already uses for its three labels.

No `translatesAutoresizingMaskIntoConstraints` line is needed, since the stack views set it on the views handed to them.

```suggestion
        _titleLabel = [NSTextField labelWithString:@""];
        _titleLabel.font = [NSFont boldSystemFontOfSize:18.0];

        _subtitleLabel = [NSTextField labelWithString:@""];
        _subtitleLabel.font = [NSFont systemFontOfSize:14.0];

        for (NSTextField* label in @[ _titleLabel, _subtitleLabel ]) {
            label.textColor = NSColor.whiteColor;
            label.maximumNumberOfLines = 1;
            label.lineBreakMode = NSLineBreakByTruncatingMiddle;
        }
```

macosx/DragOverlayView.mm
```diff
@@ -21,81 +34,90 @@ @implementation DragOverlayView
...
+
+        // 5. Vertical TextStack
+        NSStackView* textStackView = [NSStackView stackViewWithViews:@[ _titleLabel, _subtitleLabel ]];
+        textStackView.orientation = NSUserInterfaceLayoutOrientationVertical;
+        textStackView.alignment = NSLayoutAttributeLeading;
+        textStackView.spacing = kTextStackSpacing;
+
+        // Prevent the vertical stack from stretching awkwardly in height
+        [textStackView setContentHuggingPriority:NSLayoutPriorityDefaultHigh forOrientation:NSLayoutConstraintOrientationVertical];
```
ckerr, line 65: This line is 131 columns, against `ColumnLimit: 128` in `.clang-format`.

macosx/DragOverlayView.mm
ckerr: We should be consistent with rest of the macOS code's style. Please keep the `f` prefix for now.

macosx/DragOverlayView.mm
lolgear: I don't understand because I use clang-format 22.1.8.
All checks passed.

lolgear: Ok, I see.

In Objective-C section

```
PenaltyExcessCharacter: 1
```

If I change it to 2, it rewrites a lot of files.


@ckerr 
What to do here?
Only at PenaltyExcessCharacter: 5 it rewrites this DragOverlayView.

ckerr: Ugh, let's not rewrite unrelated files. If the linter is happy, let's leave it as-is. :man_shrugging:

macosx/DragOverlayView.mm
ckerr: We should be consistent with rest of the macOS code's style. Please keep the f prefix for now.

macosx/DragOverlayView.mm
ckerr: Still needs `static CGFloat const kBadgeCornerRadius = 15.0;` in the `// Badge` section at line 9

lolgear: Added.

ckerr:
Code looks OK. Can you attach some before/after screenshots @lolgear ?

lolgear:
## Results

### Before

Before Снимок экрана — 2026-09-05 в 18 41 00

### After

After Снимок экрана — 2026-09-05 в 18 38 06

lolgear:
@ckerr 
Let's merge it

ckerr (approved):

ckerr:
@lolgear I usually ban users who include screenshots of those kinds of torrents. Don't do that.

PR #264, opened 2026-08-25
refactor: encapsulate group cell internal views.

lolgear:
## Description
This PR significantly improves the encapsulation and layout architecture of `GroupCell` in the macOS interface. It modernizes how group status data (speeds and ratio) is managed and rendered.

### Key Changes:
* **Enhanced Encapsulation**: Removed internal UI subviews (`NSTextField` and `NSImageView` instances) from the public header (`GroupCell.h`). 
* **Clean API Surface**: Exposed high-level, business-logic-driven methods (`setDownloadSpeed:uploadSpeed:ratio:`, `setDisplayRatio:`, `setTooltipForTorrentsCount:`) to handle data updates, keeping the rendering logic internal to the cell.
* **UI Simplification**: Combined separate icon/text pairs into borderless `NSButton` components and encapsulated them inside a private `NSStackView` to automate layout and drastically reduce manual Autolayout boilerplate.
* **Modernized Property Names**: Renamed old properties (`fGroupIndicatorView` -> `indicatorView`) to match modern objective-c naming conventions.

## Motivation and Context
Previously, external controllers had to interact directly with the cell's underlying subviews to update the UI. By encapsulating these views and providing a dedicated public API, we reduce coupling, enforce better separation of concerns, and make the UI layer much more maintainable.

### DIFF
--- a/macosx/GroupCell.mm
+++ b/macosx/GroupCell.mm
@@ -3,6 +3,14 @@
 // License text can be found in the licenses/ folder.
 
 #import "GroupCell.h"
+#import "NSStringAdditions.h"
+
+@interface GroupCell ()
+@property(nonatomic, readonly) NSStackView* stackView;
+@property(nonatomic, readonly) NSButton* downloadButton;
+@property(nonatomic, readonly) NSButton* uploadButton;
+@property(nonatomic, readonly) NSButton* ratioButton;
+@end
 
 @implementation GroupCell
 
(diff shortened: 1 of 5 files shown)


macosx/GroupCell.mm
```diff
@@ -20,98 +28,112 @@ - (void)configureViews
...
-        view.selectable = NO;
-        view.bordered = NO;
-        view.font = [NSFont boldSystemFontOfSize:NSFont.smallSystemFontSize];
-        view.drawsBackground = NO;
+    auto downloadButton = [NSButton buttonWithTitle:@"" image:[NSImage imageNamed:@"DownArrowGroupTemplate"] target:nil action:nil];
+    downloadButton.toolTip = NSLocalizedString(@"Download speed", "Torrent table -> group row -> tooltip");
+
+    auto uploadButton = [NSButton buttonWithTitle:@"" image:[NSImage imageNamed:@"UpArrowGroupTemplate"] target:nil action:nil];
+    uploadButton.toolTip = NSLocalizedString(@"Upload speed", "Torrent table -> group row -> tooltip");
+    uploadButton.image.accessibilityDescription = NSLocalizedString(@"UL", "Torrent -> status image");
```
lolgear, line 56: Both uploadRate element ( image in button or image view ) and ratio element have accessibility description for image.

However, downloadRate element doesn't have it.
I guess it is a bug.

But resources don't contain appropriate string.
The most appropriate string is "DL".

See

```
// Spanish Localizable.strings
/* inspector -> peer table -> header
   status bar -> status label
   Torrent -> status image
   Torrent -> status string */
"UL" = "↑";
```

```
// Spanish Localizable.strings
/* inspector -> peer table -> header
   inspector -> web seed table -> header
   status bar -> status label
   Torrent -> status string */
"DL" = "↓";
```

lolgear:
@Coeur @nevack 
Could you check it?

lolgear:
@ckerr 
Could you check it?

macosx/GroupCell.mm
```diff
@@ -20,89 +40,74 @@ - (void)configureViews
...
+
+    auto ratioButton = [NSButton buttonWithTitle:@"" image:[NSImage imageNamed:@"YingYangGroupTemplate"] target:nil action:nil];
+    ratioButton.toolTip = NSLocalizedString(@"Ratio", "Torrent table -> group row -> tooltip");
+    ratioButton.image.accessibilityDescription = NSLocalizedString(@"Ratio", "Torrent -> status image");
+
```
ckerr, line 61: (Fable 5.1, High) These are enabled `NSButton`s with nil target/action, and `NSTableView`'s `validateProposedFirstResponder:forEvent:` lets an enabled non-text control take the click. A click on a speed icon or number now ends inside `-[NSButton mouseDown:]` (flashes pressed, sends nothing) and never reaches `-[TorrentTableView mouseDown:]`: no row selection, no double-click expand/collapse, and no `toggleGroupRowRatio` — the exact spot the toggle exists for. The old non-editable `NSTextField`/`NSImageView` fell through to the table. `TorrentCellControlButton.mm` shows the same routing: in-cell buttons receive the event and have to hand focus back to the table themselves.

Fixes that keep the tooltips working (a `hitTest:` returning nil would also lose the button tooltips, since the tooltip manager resolves the owner via hit-testing): keep passive `NSImageView` + `NSTextField` pairs (a nested `NSStackView` per stat keeps the layout win), or have `TorrentTableView` override `validateProposedFirstResponder:forEvent:` to return NO for responders inside a `GroupCell`.

macosx/GroupCell.mm
```diff
@@ -20,89 +40,74 @@ - (void)configureViews
...
-        [self.fGroupUploadAndRatioField.leadingAnchor constraintEqualToAnchor:self.fGroupUploadAndRatioView.trailingAnchor],
-        [self.fGroupUploadAndRatioField.trailingAnchor constraintEqualToAnchor:self.trailingAnchor constant:-5],
-        [self.fGroupUploadAndRatioField.centerYAnchor constraintEqualToAnchor:self.centerYAnchor],
-        [self.fGroupUploadAndRatioField.widthAnchor constraintGreaterThanOrEqualToConstant:60],
+        [self.fLeadingStackView.leadingAnchor constraintEqualToAnchor:self.leadingAnchor],
+        [self.fLeadingStackView.centerYAnchor constraintEqualToAnchor:self.centerYAnchor],
+        [self.fIndicatorView.widthAnchor constraintEqualToConstant:kIndicatorSize],
+        [self.fIndicatorView.heightAnchor constraintEqualToConstant:kIndicatorSize],
+
+        [self.fTrailingStackView.leadingAnchor constraintGreaterThanOrEqualToAnchor:self.fTitleField.trailingAnchor],
```
ckerr, line 105: (Fable 5.1, High) This `>=` changes what `frameForTitle` means to `pointInGroupStatusRect:`. Before, `title.trailing == downloadView.leading` plus the value fields' hugging 251 vs the title's 250 stretched the title up to the download icon, so `point.x > NSMaxX(title.frame)` meant "on the status widgets". Now nothing pulls the title past its text width (the leading stack hugs at 250 and only this inequality references the title's trailing edge), so every click in the blank gap between the group name and the speeds flips `DisplayGroupRowRatio` and reloads all visible rows. Together with the button issue above, the toggle now fires everywhere except on the widgets.

Suggest letting the cell answer the question from the trailing stack's frame (see the `frameForTitle` comment below), or restoring an equality so the title fills the gap.

macosx/GroupCell.mm
```diff
@@ -20,89 +40,74 @@ - (void)configureViews
...
+        [self.fTrailingStackView.trailingAnchor constraintEqualToAnchor:self.trailingAnchor],
+        [self.fTrailingStackView.centerYAnchor constraintEqualToAnchor:self.centerYAnchor],
+        [self.fTrailingStackView.heightAnchor constraintEqualToConstant:kTrailingStackSize],
     ]];
```
ckerr, line 113: (Fable 5.1, High) The `>= 60` minimum width on both value fields is gone and nothing constrains the buttons' widths, so the trailing stack packs from the right at intrinsic width. `updateUI` reloads visible rows every second; when "987 KB/s" becomes "1.02 MB/s" the upload button changes width and the download icon and number slide with it, and icons no longer line up across group rows. A `widthAnchor >= 60` per button (or a fixed width plus `monospacedDigitSystemFontOfSize:weight:`) restores stable columns.

Cosmetic drift to be aware of: indicator→title spacing is now `NSStackView`'s default 8pt (was 5), and the 16pt icon box became 14pt via `kTrailingInsets`.

macosx/GroupCell.mm
```diff
@@ -20,89 +40,74 @@ - (void)configureViews
...
+
+    auto uploadButton = [NSButton buttonWithTitle:@"" image:[NSImage imageNamed:@"UpArrowGroupTemplate"] target:nil action:nil];
+    uploadButton.toolTip = NSLocalizedString(@"Upload speed", "Torrent table -> group row -> tooltip");
+    uploadButton.image.accessibilityDescription = NSLocalizedString(@"UL", "Torrent -> status image");
+
+    auto ratioButton = [NSButton buttonWithTitle:@"" image:[NSImage imageNamed:@"YingYangGroupTemplate"] target:nil action:nil];
+    ratioButton.toolTip = NSLocalizedString(@"Ratio", "Torrent table -> group row -> tooltip");
+    ratioButton.image.accessibilityDescription = NSLocalizedString(@"Ratio", "Torrent -> status image");
+
+    for (NSButton* button in @[ downloadButton, uploadButton, ratioButton ]) {
```
ckerr, line 62: (Fable 5.1, High) Accessibility regression from using buttons as labels: VoiceOver reads each stat as e.g. "1.2 MB/s, button" — an actionable control that does nothing — and the `image.accessibilityDescription` values above ("UL"/"Ratio") are inert because a titled `NSButton` takes its AX label from the title. Full Keyboard Access also tabs through three inert buttons per group row (`refusesFirstResponder` is unset).

If the buttons stay: `refusesFirstResponder = YES`, `accessibilityRole = NSAccessibilityStaticTextRole`, and set `accessibilityLabel` on the button (e.g. "Upload speed, 1.2 MB/s") rather than on the image.

macosx/GroupCell.mm
```diff
@@ -20,89 +40,74 @@ - (void)configureViews
...
+        button.imageScaling = NSImageScaleProportionallyDown;
+        button.imagePosition = NSImageLeft;
+        button.bordered = NO;
+        button.font = [NSFont boldSystemFontOfSize:NSFont.smallSystemFontSize];
+        button.contentTintColor = NSColor.secondaryLabelColor;
```
ckerr, line 67: (Fable 5.1, High) `NSTableCellView` forwards `backgroundStyle` only to direct subviews that are controls or implement `setBackgroundStyle:`; `NSStackView` is neither, and this line pins an explicit tint anyway. On a selected group row the arrow/yin-yang template icons therefore stay `secondaryLabelColor` on the accent background, whereas the old direct-subview `NSImageView`s got the emphasized rendering. (The speed text was gray on selection before too.) `TorrentCell.mm` handles this by swapping `contentTintColor` in `setBackgroundStyle:`; the override below only recolors the title.

macosx/GroupCell.mm
```diff
@@ -111,7 +116,48 @@ - (void)setBackgroundStyle:(NSBackgroundStyle)backgroundStyle
...
+        tooltipGroup = NSLocalizedString(@"%lu transfers", "Torrent table -> group row -> tooltip");
+        tooltipGroup = [NSString localizedStringWithFormat:tooltipGroup, count];
+    }
+    self.toolTip = tooltipGroup;
+}
+
+- (CGRect)frameForTitle
+{
+    return self.fTitleField.frame;
 }
```
ckerr, line 176: (Fable 5.1, High) `fTitleField.frame` is now in `fLeadingStackView`'s coordinate space (the stack is its superview), not the cell's. It coincides today only because the stack's leading is pinned to the cell's leading with no constant; an inset on the stack, or any use of the rect's y, silently mis-hit-tests.

Better altitude: keep the geometry inside the cell and expose the question the table asks, e.g. `- (BOOL)isPointInStatusArea:(NSPoint)pointInCell` computed from `fTrailingStackView.frame` (a direct subview, so already cell coordinates); `TorrentTableView` passes `[groupCell convertPoint:point fromView:self]`. That also fixes the blank-gap toggle above.

macosx/TorrentGroup.mm
```diff
@@ -60,4 +77,27 @@ - (CGFloat)downloadRate
     return rate;
 }
 
+- (TorrentGroupData*)aggregatedData
```
ckerr: (Fable 5.1, High) After this PR `ratio`, `uploadRate` and `downloadRate` (lines 49–78, declared in `TorrentGroup.h` 23–25) have no callers — `TorrentTableView.mm` was their only user — so the same aggregation now exists twice. Either delete the three, or define them as one-liners over `aggregatedData` so the two cannot drift (a later "exclude paused torrents" tweak landing in only one of them).

macosx/TorrentGroup.h
```diff
@@ -6,6 +6,12 @@
 
 @class Torrent;
 
+@interface TorrentGroupData : NSObject
+@property(nonatomic, readonly) CGFloat ratio;
+@property(nonatomic, readonly) CGFloat uploadRate;
+@property(nonatomic, readonly) CGFloat downloadRate;
+@end
```
ckerr: (Fable 5.1, High) `TorrentGroupData` is a heap-allocated `NSObject` with a private-category initializer carrying three doubles that the only caller unpacks immediately, allocated once per group row per second. A plain struct returned by value does the job with no class, no init and no allocation:

```objc
typedef struct
{
    CGFloat ratio;
    CGFloat uploadRate;
    CGFloat downloadRate;
} TorrentGroupData;

@property(nonatomic, readonly) TorrentGroupData aggregatedData;
```

and `aggregatedData` ends with `return { tr_getRatio(uploaded, total_size), uploadRate, downloadRate };`, which also drops the `auto result = …; return result;` at `TorrentGroup.mm` 99–100.

macosx/GroupCell.mm
```diff
@@ -20,89 +40,74 @@ - (void)configureViews
...
+    uploadButton.toolTip = NSLocalizedString(@"Upload speed", "Torrent table -> group row -> tooltip");
+    uploadButton.image.accessibilityDescription = NSLocalizedString(@"UL", "Torrent -> status image");
+
+    auto ratioButton = [NSButton buttonWithTitle:@"" image:[NSImage imageNamed:@"YingYangGroupTemplate"] target:nil action:nil];
+    ratioButton.toolTip = NSLocalizedString(@"Ratio", "Torrent table -> group row -> tooltip");
+    ratioButton.image.accessibilityDescription = NSLocalizedString(@"Ratio", "Torrent -> status image");
+
+    for (NSButton* button in @[ downloadButton, uploadButton, ratioButton ]) {
+        button.imageScaling = NSImageScaleProportionallyDown;
+        button.imagePosition = NSImageLeft;
```
ckerr: (Fable 5.1, High) The old layout was fully leading/trailing-anchored, so the icon sat on the leading side of its number and mirrored under RTL (the app ships `he.lproj`). `NSImageLeft` is absolute, so in RTL the arrow now lands on the trailing side. Same class of issue for `kLeadingInsets`/`kTrailingInsets`: `NSEdgeInsets` are left/right, so the 11pt and 5pt margins do not mirror either, whereas the leading/trailing anchor constants they replaced did.

```suggestion
        button.imagePosition = NSImageLeading;
```

macosx/GroupCell.mm
```diff
@@ -111,7 +116,48 @@ - (void)setBackgroundStyle:(NSBackgroundStyle)backgroundStyle
...
+{
+    self.fTitleField.stringValue = title;
+}
+
+- (void)updateDownloadSpeed:(CGFloat)downloadSpeed uploadSpeed:(CGFloat)uploadSpeed ratio:(CGFloat)ratio
+{
+    _fDownloadButton.title = [NSString stringForSpeed:downloadSpeed];
+    _fUploadButton.title = [NSString stringForSpeed:uploadSpeed];
+    _fRatioButton.title = [NSString stringForRatio:ratio];
+}
```
ckerr, line 152: (Fable 5.1, High) This formats all three strings on every reload although only two buttons are ever visible; the old code formatted just the displayed pair. Passing `displayRatio` in here (or folding `updateDisplayRatio:` into this method) skips the hidden one.

Nit: these are the only `_fX` ivar reads in the class; everything else goes through `self.`.

ckerr (changes requested):
Reviewed with fable 5.1, high effort, with the prompt:

> code-reviewPR 264. . Ignore pre-existing issues unless this PR makes them worse. Add review comments/suggestions inline in the GitHub review as drafts. Preface each comment/suggestion with "(Fable 5.1, High)"

macosx/GroupCell.mm
lolgear: Hm, no.
These buttons are not only expected, they also WILL toggle ratio/speed.
In next PRs.

macosx/GroupCell.mm
lolgear: Done.

macosx/GroupCell.mm
lolgear: This is intentional.

macosx/TorrentGroup.h
lolgear: Mmm, ok.
Done.

macosx/GroupCell.mm
lolgear: Switching to leadingStackView.

macosx/GroupCell.mm
lolgear: 16 pt icon box became 14pt via `kTrailingInsets`.
What does it mean?
Icon is exactly 14pt x 14pt.
`kTrailingInsets` are used only for trailing stack.

macosx/GroupCell.mm
```diff
@@ -20,98 +41,138 @@ - (void)configureViews
...
+        button.font = [NSFont boldSystemFontOfSize:NSFont.smallSystemFontSize];
+        button.contentTintColor = NSColor.secondaryLabelColor;
+        button.lineBreakMode = NSLineBreakByClipping;
     }
 
-    for (NSView* view in @[ indicatorView, titleField, downloadView, downloadField, uploadAndRatioView, uploadAndRatioField ]) {
+    auto leadingStackView = [[NSStackView alloc] initWithFrame:NSZeroRect];
+    [leadingStackView addArrangedSubview:indicatorView];
+    [leadingStackView addArrangedSubview:titleField];
+    leadingStackView.edgeInsets = kLeadingInsets;
```
ckerr: (Opus 5, xhigh) `imagePosition` is fixed, but the insets are still geometric. `NSEdgeInsets` is left/right and `NSStackView` does not mirror it, so under RTL (the app ships `he.lproj`) the stack reverses its arranged views while the 11pt margin stays on the left. `kTrailingInsets` on line 81 has the same problem. Put both margins back on the leading and trailing anchor constants at lines 100 and 106, which do mirror.

macosx/GroupCell.mm
```diff
@@ -20,98 +41,138 @@ - (void)configureViews
...
+    } else {
+        tooltipGroup = NSLocalizedString(@"%lu transfers", "Torrent table -> group row -> tooltip");
+        tooltipGroup = [NSString localizedStringWithFormat:tooltipGroup, count];
+    }
+    self.toolTip = tooltipGroup;
+}
+
+- (CGRect)frameForTitle
+{
+    return self.fLeadingStackView.frame;
```
ckerr: (Opus 5, xhigh) `frameForTitle` now returns the leading stack rather than the title, so the name misleads its only caller, which reads the rect as "where the title ends". Replace it with the predicate that caller actually wants:

```objc
- (BOOL)isPointInStatusArea:(NSPoint)pointInCell
{
    return pointInCell.x >= NSMinX(self.fTrailingStackView.frame);
}
```

`fTrailingStackView` is a direct subview, so its frame is already in cell coordinates, and asking about the status area rather than the title also settles the blank-gap toggle reported on line 105.

macosx/TorrentGroup.mm
```diff
@@ -29,35 +29,27 @@ - (NSInteger)groupOrderValue
...
-{
-    CGFloat rate = 0.0;
-    for (Torrent* torrent in self.torrents) {
-        rate += torrent.downloadRate;
-    }
+    CGFloat ratio = tr_getRatio(uploaded, total_size);
 
-    return rate;
+    auto result = (TorrentGroupData){ ratio, uploadRate, downloadRate };
+    return result;
```
ckerr: (Opus 5, xhigh) The cast makes this a C compound literal, which `-Wgnu-compound-literal` warns on; `-Wgnu` is enabled project-wide at CMakeLists.txt:826. Aggregate-initialize the return value instead, which also drops the temporary.

```suggestion
    return { ratio, uploadRate, downloadRate };
```

ckerr (changes requested):
Re-reviewed b8c963f4d (Opus 5, xhigh).

Fixed since be67e2d46, threads resolved: the 60pt button widths and hugging priorities are back; `setBackgroundStyle:` tints the buttons; `frameForTitle` returns a direct subview, so it is in cell coordinates; `TorrentGroupData` is a struct; the dead `ratio`/`uploadRate`/`downloadRate` accessors are gone; `imagePosition` is `NSImageLeading`.

Still blocking, in the open threads on `GroupCell.mm`:

- The three `NSButton`s take the click, so clicking the status area no longer toggles the ratio display.
- `frameForTitle` ends at the title text, so clicking the blank gap toggles it instead.
- The buttons read as actionable controls to VoiceOver and Full Keyboard Access.

`updateDownloadSpeed:uploadSpeed:ratio:` still formats all three strings although one button is always hidden. The ivar nit in that thread is fixed.

PR #268, opened 2026-08-26
perf: optimize cells reloading in hovering related events

lolgear:
### Description
This PR optimizes the cell reloading logic when handling hover-related events in the UI. Previously, hovering over elements triggered redundant or full cell redraws, leading to unnecessary CPU usage and potential UI stuttering in large torrent lists. 

### Changes
* Optimized the conditions under which cells are reloaded during mouse hover events.
* Reduced redundant cell updates to ensure smoother scrolling and UI interaction.
* Fixed unnecessary layout recalculations triggered by hover states.

### Impact
* **Performance:** Lower CPU consumption during list interactions.
* **UX:** Smoother hovering effect and more responsive scrolling, especially visible in configurations with hundreds of active torrents.

### Testing Done
* Verified that hover states (tooltips, highlights) still render correctly.
* Checked for regressions in cell selection and context menu triggers.

#### Before

https://github.com/user-attachments/assets/b4683a3a-050e-4039-ac75-c9c460dd453b

#### After

https://github.com/user-attachments/assets/03783f90-ec82-469b-b793-32bd12ced6b2

### DIFF
--- a/macosx/TorrentTableView.mm
+++ b/macosx/TorrentTableView.mm
@@ -28,6 +28,31 @@ static CGFloat const kErrorImageSize = 20.0;
 
 static NSTimeInterval const kToggleProgressSeconds = 0.175;
 
+@interface TorrentTableViewHoveringData : NSObject<NSCopying>
+@property(nonatomic, nullable, copy) NSNumber* hoveredRow;
+@property(nonatomic, nullable, copy) NSString* statusText;
+@property(nonatomic, readonly) BOOL isHovered;
+@end
+@implementation TorrentTableViewHoveringData
+- (BOOL)isHoveredForRow:(NSInteger)row
+{
+    return self.hoveredRow != nil && self.hoveredRow.integerValue == row;
+}
+
+- (BOOL)isHovered
+{
+    return self.hoveredRow != nil && self.hoveredRow.integerValue != -1;
+}
+
+- (nonnull id)copyWithZone:(nullable NSZone*)zone
+{
+    auto copy = (TorrentTableViewHoveringData*)[[self.class alloc] init];
+    copy.hoveredRow = self.hoveredRow;
+    copy.statusText = self.statusText;
+    return copy;
+}
+@end
+
 @interface TorrentTableView ()
 
 @property(nonatomic) IBOutlet Controller* fController;
@@ -45,7 +70,7 @@ static NSTimeInterval const kToggleProgressSeconds = 0.175;
 @property(nonatomic) BOOL fActionPopoverShown;
 @property(nonatomic) NSView* fPositioningView;
 
-@property(nonatomic) NSDictionary* fHoverEventDict;
+@property(nonatomic) TorrentTableViewHoveringData* hoveringData;
 
 @end
 
@@ -54,6 +79,7 @@ static NSTimeInterval const kToggleProgressSeconds = 0.175;
 - (instancetype)initWithCoder:(NSCoder*)decoder
 {
     if ((self = [super initWithCoder:decoder])) {
+        self.hoveringData = [[TorrentTableViewHoveringData alloc] init];
         _fDefaults = NSUserDefaults.standardUserDefaults;
 
         NSData* groupData;
@@ -220,15 +246,11 @@ static NSTimeInterval const kToggleProgressSeconds = 0.175;
                 torrent.shortStatusString :
                 torrent.remainingTimeString;
 
-            if (self.fHoverEventDict) {
-                NSInteger row = [self rowForItem:item];
-                NSInteger hoverRow = [self.fHoverEventDict[@"row"] integerValue];
-
-                if (row == hoverRow) {
-                    torrentCell.fTorrentStatusField.hidden = YES;
-                    torrentCell.fControlButton.hidden = NO;
-                    torrentCell.fRevealButton.hidden = NO;
-                }
+            NSInteger row = [self rowForItem:item];
+            if ([self.hoveringData isHoveredForRow:row]) {
+                torrentCell.fTorrentStatusField.hidden = YES;
+                torrentCell.fControlButton.hidden = NO;
+                torrentCell.fRevealButton.hidden = NO;
             } else {
                 torrentCell.fTorrentStatusField.hidden = NO;
                 torrentCell.fControlButton.hidden = YES;
@@ -270,12 +292,12 @@ static NSTimeInterval const kToggleProgressSeconds = 0.175;
 
             // set torrent status
             NSString* status;
-            if (self.fHoverEventDict) {
+            if (self.hoveringData.isHovered) {
                 NSInteger row = [self rowForItem:item];
-                NSInteger hoverRow = [self.fHoverEventDict[@"row"] integerValue];
+                NSInteger hoverRow = self.hoveringData.hoveredRow.integerValue;
 
                 if (row == hoverRow) {
-                    status = self.fHoverEventDict[@"string"];
+                    status = self.hoveringData.statusText;
                 }
             }
             torrentCell.fTorrentStatusField.stringValue = status ?: torrent.statusString;
@@ -582,10 +604,12 @@ static NSTimeInterval const kToggleProgressSeconds = 0.175;
     NSInteger row = [self rowForView:view];
     Torrent* torrent = [self itemAtRow:row];
 
+    auto previouslyHovered = (TorrentTableViewHoveringData*)[self.hoveringData copy];
     BOOL minimal = [self.fDefaults boolForKey:@"SmallView"];
     if (minimal) {
         if ([view isKindOfClass:[SmallTorrentCell class]]) {
-            self.fHoverEventDict = @{ @"row" : [NSNumber numberWithInteger:row] };
+            self.hoveringData.hoveredRow = @(row);
+            self.hoveringData.statusText = nil;
         } else if ([view isKindOfClass:[TorrentCellActionButton class]]) {
             SmallTorrentCell* smallCell = [self viewAtColumn:0 row:row makeIfNecessary:NO];
             smallCell.fIconView.hidden = YES;
@@ -611,11 +635,19 @@ static NSTimeInterval const kToggleProgressSeconds = 0.175;
         }
 
         if (statusString) {
-            self.fHoverEventDict = @{ @"string" : statusString, @"row" : [NSNumber numberWithInteger:row] };
+            self.hoveringData.hoveredRow = @(row);
+            self.hoveringData.statusText = statusString;
         }
     }
 
-    [self reloadVisibleRows];
+    if (previouslyHovered.isHovered) {
+        NSMutableIndexSet* indexSet = [[NSMutableIndexSet alloc] init];
+        [indexSet addIndex:previouslyHovered.hoveredRow.integerValue];
+        [indexSet addIndex:row];
+        [self reloadDataForRowIndexes:indexSet columnIndexes:[NSIndexSet indexSetWithIndex:0]];
+    } else {
+        [self reloadDataForRowIndexes:[NSIndexSet indexSetWithIndex:row] columnIndexes:[NSIndexSet indexSetWithIndex:0]];
+    }
 }
 
 - (void)hoverEventEndedForView:(id)view
@@ -635,8 +667,17 @@ static NSTimeInterval const kToggleProgressSeconds = 0.175;
     }
 
     if (update) {
-        self.fHoverEventDict = nil;
-        [self reloadDataForRowIndexes:[NSIndexSet indexSetWithIndex:row] columnIndexes:[NSIndexSet indexSetWithIndex:0]];
+        auto previouslyHovered = (TorrentTableViewHoveringData*)[self.hoveringData copy];
+        self.hoveringData.hoveredRow = nil;
+        self.hoveringData.statusText = nil;
+        if (previouslyHovered.isHovered) {
+            NSMutableIndexSet* indexSet = [[NSMutableIndexSet alloc] init];
+            [indexSet addIndex:previouslyHovered.hoveredRow.integerValue];
+            [indexSet addIndex:row];
+            [self reloadDataForRowIndexes:indexSet columnIndexes:[NSIndexSet indexSetWithIndex:0]];
+        } else {
+            [self reloadDataForRowIndexes:[NSIndexSet indexSetWithIndex:row] columnIndexes:[NSIndexSet indexSetWithIndex:0]];
+        }
     }
 }
 
--- a/macosx/TorrentTableView.mm
+++ b/macosx/TorrentTableView.mm
@@ -571,7 +571,7 @@ static NSTimeInterval const kToggleProgressSeconds = 0.175;
 - (void)updateStatusTextForTorrent:(nullable Torrent*)torrent
 {
     NSInteger const row = [self rowForItem:torrent];
-    
+
     // if torrent is nil, row will be -1.
     if (row < 0) {
         return;
@@ -586,7 +586,7 @@ static NSTimeInterval const kToggleProgressSeconds = 0.175;
 {
     BOOL minimal = [self.fDefaults boolForKey:@"SmallView"];
     NSInteger row = [self rowForView:view];
-    
+
     if (minimal) {
         if ([view isKindOfClass:[TorrentCellActionButton class]]) {
             SmallTorrentCell* smallCell = [self viewAtColumn:0 row:row makeIfNecessary:NO];
@@ -631,7 +631,7 @@ static NSTimeInterval const kToggleProgressSeconds = 0.175;
 {
     BOOL minimal = [self.fDefaults boolForKey:@"SmallView"];
     NSInteger row = [self rowForView:view];
-    
+
     if (minimal) {
         if ([view isKindOfClass:[TorrentCellActionButton class]]) {
             SmallTorrentCell* smallCell = [self viewAtColumn:0 row:row makeIfNecessary:NO];
@@ -639,12 +639,12 @@ static NSTimeInterval const kToggleProgressSeconds = 0.175;
         }
         return;
     }
-    
+
     auto previousHoveredTorrent = self.hoveringData.hoveredTorrent;
-    
+
     self.hoveringData.hoveredTorrent = nil;
     self.hoveringData.statusText = nil;
-    
+
     [self updateStatusTextForTorrent:previousHoveredTorrent];
 }
 
--- a/macosx/TorrentTableView.mm
+++ b/macosx/TorrentTableView.mm
@@ -28,28 +28,15 @@ static CGFloat const kErrorImageSize = 20.0;
 
 static NSTimeInterval const kToggleProgressSeconds = 0.175;
 
-@interface TorrentTableViewHoveringData : NSObject<NSCopying>
+@interface TorrentTableViewHoveringData : NSObject
 @property(nonatomic, nullable, copy) NSNumber* hoveredRow;
 @property(nonatomic, nullable, copy) NSString* statusText;
 @property(nonatomic, readonly) BOOL isHovered;
 @end
 @implementation TorrentTableViewHoveringData
-- (BOOL)isHoveredForRow:(NSInteger)row
-{
-    return self.hoveredRow != nil && self.hoveredRow.integerValue == row;
-}
-
 - (BOOL)isHovered
 {
-    return self.hoveredRow != nil && self.hoveredRow.integerValue != -1;
-}
-
-- (nonnull id)copyWithZone:(nullable NSZone*)zone
-{
-    auto copy = (TorrentTableViewHoveringData*)[[self.class alloc] init];
-    copy.hoveredRow = self.hoveredRow;
-    copy.statusText = self.statusText;
-    return copy;
+    return self.hoveredRow != nil;
 }
 @end
 
@@ -588,6 +575,12 @@ static NSTimeInterval const kToggleProgressSeconds = 0.175;
     return YES;
 }
 
+- (void)handleIsHoveringForRow:(NSInteger)row statusText:(NSString*)text
+{
+    auto view = (TorrentCell*)[self viewAtColumn:0 row:row makeIfNecessary:NO];
+    view.fTorrentStatusField.stringValue = text;
+}
+
 - (void)hoverEventBeganForView:(id)view
 {
     BOOL minimal = [self.fDefaults boolForKey:@"SmallView"];
@@ -598,8 +591,6 @@ static NSTimeInterval const kToggleProgressSeconds = 0.175;
 
     NSInteger row = [self rowForView:view];
     Torrent* torrent = [self itemAtRow:row];
-
-    auto previouslyHovered = (TorrentTableViewHoveringData*)[self.hoveringData copy];
 
     NSString* statusString;
     if ([view isKindOfClass:[TorrentCellRevealButton class]]) {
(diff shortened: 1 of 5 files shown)


macosx/TorrentTableView.mm
```diff
@@ -611,11 +635,19 @@ - (void)hoverEventBeganForView:(id)view
...
     }
 
-    [self reloadVisibleRows];
+    if (previouslyHovered.isHovered) {
+        NSMutableIndexSet* indexSet = [[NSMutableIndexSet alloc] init];
+        [indexSet addIndex:previouslyHovered.hoveredRow.integerValue];
+        [indexSet addIndex:row];
+        [self reloadDataForRowIndexes:indexSet columnIndexes:[NSIndexSet indexSetWithIndex:0]];
+    } else {
+        [self reloadDataForRowIndexes:[NSIndexSet indexSetWithIndex:row] columnIndexes:[NSIndexSet indexSetWithIndex:0]];
```
ckerr: Is it possible for `row` (created on line 604) to be `-1` here?

If so, we'll get `[NSIndexSet indexSetWithIndex:-1] ` throwing an `NSRangeException`.

macosx/TorrentTableView.mm
```diff
@@ -582,10 +604,12 @@ - (void)hoverEventBeganForView:(id)view
     NSInteger row = [self rowForView:view];
     Torrent* torrent = [self itemAtRow:row];
 
+    auto previouslyHovered = (TorrentTableViewHoveringData*)[self.hoveringData copy];
     BOOL minimal = [self.fDefaults boolForKey:@"SmallView"];
     if (minimal) {
         if ([view isKindOfClass:[SmallTorrentCell class]]) {
-            self.fHoverEventDict = @{ @"row" : [NSNumber numberWithInteger:row] };
+            self.hoveringData.hoveredRow = @(row);
```
ckerr: - Is it possible for `row` to be -1 here? If so, seems problematic to store -1 as the hovered row?

- I don't see where `hoveredRow` gets cleared anywhere?  Hover row N near the bottom, then remove or filter torrents s.t. there are fewer visible rows, and hoveredRow == N survives 💥

macosx/TorrentTableView.mm
```diff
@@ -220,15 +246,11 @@ - (NSView*)outlineView:(NSOutlineView*)outlineView viewForTableColumn:(NSTableCo
...
-            if (self.fHoverEventDict) {
-                NSInteger row = [self rowForItem:item];
-                NSInteger hoverRow = [self.fHoverEventDict[@"row"] integerValue];
-
-                if (row == hoverRow) {
-                    torrentCell.fTorrentStatusField.hidden = YES;
-                    torrentCell.fControlButton.hidden = NO;
-                    torrentCell.fRevealButton.hidden = NO;
-                }
+            NSInteger row = [self rowForItem:item];
```
ckerr: We used to only look up the row when `self.fHoverEventDict` existed, now we're doing it unconditionally.  Is this necessary?

macosx/TorrentTableView.mm
```diff
@@ -635,8 +667,17 @@ - (void)hoverEventEndedForView:(id)view
...
+        auto previouslyHovered = (TorrentTableViewHoveringData*)[self.hoveringData copy];
+        self.hoveringData.hoveredRow = nil;
+        self.hoveringData.statusText = nil;
+        if (previouslyHovered.isHovered) {
+            NSMutableIndexSet* indexSet = [[NSMutableIndexSet alloc] init];
+            [indexSet addIndex:previouslyHovered.hoveredRow.integerValue];
+            [indexSet addIndex:row];
+            [self reloadDataForRowIndexes:indexSet columnIndexes:[NSIndexSet indexSetWithIndex:0]];
+        } else {
+            [self reloadDataForRowIndexes:[NSIndexSet indexSetWithIndex:row] columnIndexes:[NSIndexSet indexSetWithIndex:0]];
```
ckerr: This copies lines 633-649 verbatim, can we extract-method here?

macosx/TorrentTableView.mm
```diff
@@ -582,10 +604,12 @@ - (void)hoverEventBeganForView:(id)view
     NSInteger row = [self rowForView:view];
     Torrent* torrent = [self itemAtRow:row];
 
+    auto previouslyHovered = (TorrentTableViewHoveringData*)[self.hoveringData copy];
```
ckerr: We don't need a deep copy of the entire row.

All we need is `NSNumber* previouslyHoveredRow = self.hoveringData.hoveredRow;` + a helper function `isHovered(NSNumber* row) { return  row != nil && row >= 0; }`

macosx/TorrentTableView.mm
```diff
@@ -635,8 +667,17 @@ - (void)hoverEventEndedForView:(id)view
     }
 
     if (update) {
-        self.fHoverEventDict = nil;
-        [self reloadDataForRowIndexes:[NSIndexSet indexSetWithIndex:row] columnIndexes:[NSIndexSet indexSetWithIndex:0]];
+        auto previouslyHovered = (TorrentTableViewHoveringData*)[self.hoveringData copy];
```
ckerr: same here, we only need the row, not the entire deep copy

ckerr (changes requested):
@lolgear I really love that you've been giving the macOS client all this perf attention so don't be discouraged by this, please keep the PRs coming --

but most times I've done a review on this, both human & CI review have turned up a lot of findings. If you're using an agent, are you using it to do an adversarial code review before opening the PR?

ckerr (commented):
(Claude)

- Don't reload rows at all for hover; mutate the live cells. A hover toggles three `hidden` flags and a status string, but `reloadDataForRowIndexes:` tears down and rebuilds the whole cell view to get there. The cheaper move is `viewAtColumn:row:makeIfNecessary:NO` and flip the properties on the existing view in place — the codebase already does exactly this in the action-button branch.  Keep `viewForTableColumn:` reading the hover model so real reloads (scroll-in, periodic refresh) reconstruct the appearance; that part of the PR's design is right.

- One mutator, not two pasted sequences. A single `setHoveredRow:statusText:` that snapshots the old row, rejects negative rows, updates the model, and touches only the changed cells. That makes "state changed → affected cells updated" an invariant instead of a per-call-site obligation, and structurally removes the -1 crash, the stale-index feed, and the duplication — findings that all exist because each call site re-implements the sequence.

- Keep the `updateTrackingAreas` resync; don't delete it. The synthetic enter/exit was compensating for a real AppKit gap (tracking areas only fire on pointer movement). It was expensive only because each synthetic event triggered a broad reload — once hover updates are cheap per-cell mutations, the resync costs almost nothing. The PR deleted the mechanism instead of fixing what made it expensive, which is where the stale-hover regression comes from. (Its enter half was already broken — blank `NSEvent` location — so I'd fix that while there.)

lolgear:
### Agents

No, I don't use the agent because running it on an Intel MacBook is resource-heavy (and sometimes not feasible).

Since nearly every developer now uses M1 or newer chips, it has its pros and cons.

On one hand, we leverage new technologies.

On the other hand, it makes it harder for the team to test how the client performs on older hardware.

### Feature

This feature is quite debatable. 

Hovering the mouse pointer over a torrent just to reveal it in Finder or pause it might not be intuitive for the user.

Having visible buttons seems much more user-friendly.

I'd suggest reconsidering or even removing this hover behavior.

To update the layout, we'd only need to add two or three constraints—and voila! The buttons will be fully visible, allowing users to pause or locate files effortlessly.

### Implementation

Yes, I will double-check some corner cases, but even without them, this approach seems more reliable than what we currently have on main.

However, here is the interesting part: most actions and event sources currently rely on a simple data reload for visible cells (e.g., a timer refreshing data every second).

Transitioning to updating existing cells might require some extra effort.

lolgear:
@ckerr 

I sorted out some stuff and it works a lot better for small torrent cell, yes.
No more unnecessary reloads.

However, for TorrentCell it is impossible in current state to do the same, because tooltips are shown as part of model ( literally in torrent subtitle field ).

This obstacle can be eliminated by shifting to standard macOS UI as "tooltip".

lolgear:
@ckerr 
Could you check it again?
Not the best UX solution, but it is very close to previous one ( without extensive reloading ).

macosx/TorrentTableView.mm
```diff
@@ -577,66 +575,73 @@ - (BOOL)validateMenuItem:(NSMenuItem*)menuItem
     return YES;
 }
 
+- (void)handleIsHoveringForRow:(NSInteger)row statusText:(NSString*)text
+{
+    auto view = (TorrentCell*)[self viewAtColumn:0 row:row makeIfNecessary:NO];
+    view.fTorrentStatusField.stringValue = text;
+}
```
ckerr:
```suggestion
// A row's status field shows the hover tooltip while that row is hovered and the
// torrent's own status otherwise, matching what outlineView:viewForTableColumn:
// renders when it builds the cell.
- (void)updateStatusTextForRow:(NSInteger)row
{
    if (row < 0 || row >= self.numberOfRows) {
        return;
    }

    auto torrent = (Torrent*)[self itemAtRow:row];
    if (![torrent isKindOfClass:[Torrent class]]) {
        return;
    }

    BOOL const hovered = self.hoveringData.hoveredRow != nil && self.hoveringData.hoveredRow.integerValue == row;
    auto cell = (TorrentCell*)[self viewAtColumn:0 row:row makeIfNecessary:NO];
    cell.fTorrentStatusField.stringValue = (hovered ? self.hoveringData.statusText : nil) ?: torrent.statusString;
}
```

macosx/TorrentTableView.mm
```diff
@@ -577,66 +575,73 @@ - (BOOL)validateMenuItem:(NSMenuItem*)menuItem
...
+        self.hoveringData.hoveredRow = @(row);
+        self.hoveringData.statusText = statusString;
+
+        [self handleIsHoveringForRow:row statusText:statusString];
     }
 
-    [self reloadVisibleRows];
+    if (previousHoveredRow != nil) {
+        [self handleIsHoveringForRow:previousHoveredRow.integerValue statusText:torrent.statusString];
+    }
```
ckerr:
```suggestion
    if (statusString) {
        self.hoveringData.hoveredRow = @(row);
        self.hoveringData.statusText = statusString;
    }

    if (previousHoveredRow != nil) {
        [self updateStatusTextForRow:previousHoveredRow.integerValue];
    }
    [self updateStatusTextForRow:row];
```

macosx/TorrentTableView.mm
```diff
@@ -577,66 +575,73 @@ - (BOOL)validateMenuItem:(NSMenuItem*)menuItem
...
+    auto previousHoveredRow = self.hoveringData.hoveredRow;
+
+    self.hoveringData.hoveredRow = nil;
+    self.hoveringData.statusText = nil;
+
+    if (previousHoveredRow != nil) {
+        NSInteger row = [self rowForView:view];
+        auto torrent = (Torrent*)[self itemAtRow:row];
+        [self handleIsHoveringForRow:previousHoveredRow.integerValue statusText:torrent.statusString];
     }
```
ckerr:
```suggestion
    if (previousHoveredRow != nil) {
        [self updateStatusTextForRow:previousHoveredRow.integerValue];
    }
```

macosx/TorrentTableView.mm
ckerr: Looks like the deep copy is gone but we still need the `isHovered` helper. We're just using the `nil` check instead of `nil` check + `row >= 0`.

macosx/TorrentTableView.mm
ckerr: Looks like bullet point 1 is fixed, but bullet point 2 still needs work:

`hoveredRow` is still cleared only by `hoverEventEndedForView:`, never on a data change. If row N survives a removal but now holds a different torrent, `TorrentTableView.mm:271` paints the stale tooltip onto it.

ckerr: (claude opus 5. Prompt: "What would be a clean fix for the uncleared `hoveredRow` issue")

Stop storing a row index. Store the hovered torrent, weakly — then there's nothing to clear.

The bug isn't a missing clear-hook; it's that hoveredRow is a position being used as an identity. Positions shift under sorting, filtering, and removal, so any clearing scheme has to hook every mutation path and will miss the next one someone adds. Identity doesn't shift.

```objc++
@interface TorrentTableViewHoveringData : NSObject
@property(nonatomic, weak) Torrent* hoveredTorrent;
@property(nonatomic, copy) NSString* statusText;
@end
```

`weak` is what does the work: when the torrent is removed, the reference nils itself and the hover state evaporates with no notification, no observer, no clearing call. Row shifts stop mattering because nothing compares rows any more.

The helper then derives the row instead of trusting one:

```objc++
- (void)updateStatusTextForTorrent:(Torrent*)torrent
{
    NSInteger const row = [self rowForItem:torrent];
    if (row < 0) {
        return;
    }

    auto cell = (TorrentCell*)[self viewAtColumn:0 row:row makeIfNecessary:NO];
    NSString* const hoverText = torrent == self.hoveringData.hoveredTorrent ? self.hoveringData.statusText : nil;
    cell.fTorrentStatusField.stringValue = hoverText ?: torrent.statusString;
}
```

and the handlers pass `torrent` / `previousTorrent` where they currently pass row numbers.

Four of the open problems dissolve rather than get fixed. The -1 sentinel has nowhere to live: `itemAtRow:-1` yields `nil`, storing `nil` means "not hovered", and that's already the encoding — so the unguarded `@(row)` store stops being a hazard. The `row >= numberOfRows` bound goes away because `rowForItem:` returns `-1` for a torrent that isn't displayed, which is the same guard derived instead of asserted. The `isKindOfClass:[Torrent class]` check goes away because the parameter is typed. And `viewForTableColumn:` drops its `rowForItem:` call for a pointer comparison:

```objc++
NSString* const status = torrent == self.hoveringData.hoveredTorrent ? self.hoveringData.statusText : nil;
torrentCell.fTorrentStatusField.stringValue = status ?: torrent.statusString;
```

which is cheaper on every reload than what's there now, and lets isHovered be deleted. The one cost is a weak read per cell, far below the row scan it replaces.

Precedent is already in the tree: `FileListNode.h:14` holds `@property(nonatomic, weak, readonly) Torrent* torrent`, and ARC is on for `macosx/`. `Torrent` defines no `isEqual:`, so pointer identity is the right comparison.

ckerr (changes requested):

macosx/TorrentTableView.mm
lolgear: O?
It is essentially NOT what user expects.
If somehow torrent moves up or down, it would be "hovered".

lolgear:
Ok, addressing both problems.

We could store `safeRange` in `HoveringData` ( and update it in `setTorrents:` ).

```objective-c
self.safeRange = NSRange(0, self.torrents.count);


- (BOOL)isHovered {
  return _hoveredRow != nil && self.safeRange.length > 0 && self.safeRange.contains(_hoveredRow.integerValue);
}

- (NSNumber*)hoveredRow {
   if (self.isHovered) { return _hoveredRow; }
   return nil;
}
```

Then, hoveringRow will return hoveringRow IF it safeRange contains this row OR nil.

ckerr:
> No, I don't use the agent because running it on an Intel MacBook is resource-heavy (and sometimes not feasible).

@lolgear  your attention to performance in the macOS UI is improving the project and I love contributors so interested in improving performance & improving the codebase. :heart: 

But we've had three review cycles on this PR alone and each one's yielded significant change requests. Given that you're submitting multiple PRs a day, I don't know if this kind of review cycle is sustainable -- each review takes significant time for me.

If an Electron-based UI like Claude Desktop is too heavyweight for your development machine, perhaps you could use a CLI ("Claude Code CLI", "OpenAI Codex CLI") or a browser (claude.ai/code, Codex Cloud) instead?

ckerr:
> Ok, addressing both problems.
> 
> We could store `safeRange` in `HoveringData` ( and update it in `setTorrents:` ).
> 
> ```objc
> self.safeRange = NSRange(0, self.torrents.count);
> 
> 
> - (BOOL)isHovered {
>   return _hoveredRow != nil && self.safeRange.length > 0 && self.safeRange.contains(_hoveredRow.integerValue);
> }
> 
> - (NSNumber*)hoveredRow {
>    if (self.isHovered) { return _hoveredRow; }
>    return nil;
> }
> ```
> 
> Then, hoveringRow will return hoveringRow IF it safeRange contains this row OR nil.

(claude opus 5 extra, prompt: "(1) would this work (2) is this better, worse, or about the same as the change you suggested above?")

No — it wouldn't work, and it's a step back from where the code already is. Three substantive problems, in order of severity.

**It doesn't address the case that's actually broken.** A range check answers "is this index in bounds," not "does this index still point at the torrent I was hovering." The out-of-bounds case is already handled — TorrentTableView.mm:583 has if (row < 0 || row >= self.numberOfRows) { return; }. What remains is the shift case: remove a torrent from the middle, or re-sort, and hovered row 5 is still perfectly in bounds while now holding a different torrent. safeRange says "safe" and the stale tooltip paints on the wrong row. So this adds a second bounds check next to the existing one and leaves the real gap open.

**In grouped mode the count is wrong outright.** fDisplayedTorrents holds TorrentGroup objects when SortByGroup is on — Controller.mm:3263 returns its count as the number of top-level children, with each group's torrents as sub-rows. So safeRange would be the number of groups (say 3) while the table displays 50 rows. Every hovered row past 3 reports not-hovered, and tooltips silently stop working in grouped view. numberOfRows is the only count in the right domain, and it's already what the existing guard uses.

**Masking instead of clearing resurrects dead hovers.** _hoveredRow is never cleared, only hidden by the getter. Hover row 9, remove torrents down to 5 — getter returns nil, correct. Add torrents back to 12, and the getter returns 9 again, reviving a hover that ended minutes ago.

Smaller things: there's no setTorrents: on TorrentTableView to hook (the only one in the tree is DragOverlayWindow's, which takes file paths); NSRange(0, n) isn't valid ObjC (NSMakeRange, or NSRange{0, n} in ObjC++); NSRange has no .contains() — it's a C struct, so it's NSLocationInRange(loc, range); and isHovered must read _hoveredRow directly forever, since a later tidy-up to self.hoveredRow makes it infinitely recursive.

**Versus the weak-torrent version: worse, and for a structural reason.** Both approaches share the same good instinct — push validity into the accessor so callers can't misuse the value. The difference is the predicate. safeRange validates a position, which requires the table to keep pushing state into HoveringData on every mutation path; miss one and you get a stale range that actively asserts safety, which is worse than a stale row. A weak Torrent* validates identity, and ARC does the invalidation for you: no hook to remember, no count to keep in sync, no resurrection, no grouped-mode special case — and it covers the shift case, which is the one still outstanding.

Worth relaying that the instinct is right and the predicate is the thing to change.

lolgear:
@ckerr 
Let's speak about this PR and first implementation.

First implementation is the same as previous with dictionary ( except it reloads one cell instead of N ).
It works and it is the same as previous implementation ( which was approved by everyone and it was in release of transmission with all hidden performance issues ).

Then you asked Claude and now we are here with totally different but the most performant implementation.

Was it worth?
I guess so, it is really cool now and it feels very natural.
I can scroll table at 60 fps + select any cell.

Now we have a bit pedantic issue with out-of-bounds.
Actually, it never ( or we don't know about it ) happened in the first simple implementation with dictionary. ( which also contains this bug ).

ckerr:
One remaining _new_ bug I see in this PR: `main` has `smallCell.fIconView.hidden = YES` in the minimal action-button branch, and this PR deletes it with no replacement, so in SmallView the translucent gear now composites over the still-visible torrent icon.

I guess it's true that the stale hover information edge cases were pre-existing, but I don't understand why that's a reason to not fix them.  :man_shrugging:

lolgear:
@ckerr 

Changed ID to torrent.

ckerr (approved):

lolgear:
@ckerr 
Can we merge it?

lolgear:
@ckerr 
Let's merge it?

macosx/TorrentTableView.mm
```diff
@@ -577,67 +565,87 @@ - (BOOL)validateMenuItem:(NSMenuItem*)menuItem
...
+
+    auto cell = (TorrentCell*)[self viewAtColumn:0 row:row makeIfNecessary:NO];
+    NSString* const hoverText = torrent == self.hoveringData.hoveredTorrent ? self.hoveringData.statusText : nil;
+    cell.fTorrentStatusField.stringValue = hoverText ?: torrent.statusString;
+}
+
 - (void)hoverEventBeganForView:(id)view
 {
+    BOOL minimal = [self.fDefaults boolForKey:@"SmallView"];
     NSInteger row = [self rowForView:view];
```
ckerr, line 588: LGTM other than needing a `row` safeguard here:

```suggestion
    NSInteger row = [self rowForView:view];

    if (row < 0) {
        return;
    }
```

`rowForView:` returns -1 when the view isn't currently associated with a row, which `TorrentCellActionButton` can hit since it tracks with `NSTrackingActiveAlways` and can get a mouse event after its row view has been recycled during a scroll.

macosx/TorrentTableView.mm
```diff
@@ -577,67 +565,87 @@ - (BOOL)validateMenuItem:(NSMenuItem*)menuItem
...
+    [self updateStatusTextForTorrent:torrent];
 }
 
 - (void)hoverEventEndedForView:(id)view
 {
-    NSInteger row = [self rowForView:[view superview]];
-
-    BOOL update = YES;
     BOOL minimal = [self.fDefaults boolForKey:@"SmallView"];
+    NSInteger row = [self rowForView:view];
```
ckerr, line 633: same here:

```suggestion
    NSInteger row = [self rowForView:view];

    if (row < 0) {
        return;
    }
```

lolgear: Done.

macosx/TorrentTableView.mm
lolgear: Done.

lolgear:
Anything to do here?

lolgear:
@ckerr 
Let's merge it

ckerr:
@lolgear STOP IT I've

> @ckerr Let's merge it

It's only been two hours since your last comment asking if there was anything else! I've done like six code reviews on your PRs already this morning, stop nagging!

macosx/TorrentTableView.mm
```diff
@@ -577,67 +565,95 @@ - (BOOL)validateMenuItem:(NSMenuItem*)menuItem
...
-            else {
-                if (NSApp.currentEvent.modifierFlags & NSEventModifierFlagOption) {
-                    statusString = NSLocalizedString(@"Resume the transfer right away", "Torrent cell -> button info");
-                } else if (torrent.waitingToStart) {
-                    statusString = NSLocalizedString(@"Stop waiting to start", "Torrent cell -> button info");
-                } else {
-                    statusString = NSLocalizedString(@"Resume the transfer", "Torrent cell -> button info");
-                }
+        return;
+    }
```
ckerr: Opus5, extra, with prompt 'OP has pushed more commits. Please update your review'

My earlier suggestion put this guard in the wrong place — that's my mistake, not yours; you applied exactly what I asked for.

The only thing in either handler that needs a valid row is the `viewAtColumn:0 row:row` call inside the `if (minimal)` branch. At the top of the method the guard also short-circuits the non-minimal paths, which don't need a row at all — and that causes a real regression in `hoverEventEndedForView:` below. Narrowing it to the branch that indexes with the row keeps the crash fix without the side effect:

```suggestion
    if (minimal) {
        if (row >= 0 && [view isKindOfClass:[TorrentCellActionButton class]]) {
            SmallTorrentCell* smallCell = [self viewAtColumn:0 row:row makeIfNecessary:NO];
            smallCell.fIconView.hidden = YES;
        }
        return;
    }
```

Minor here rather than harmful: previously `row == -1` gave a nil torrent, so a stray enter on a detached view cleared stale hover state. With the early return it leaves the old hover painted.

macosx/TorrentTableView.mm
```diff
@@ -577,67 +565,95 @@ - (BOOL)validateMenuItem:(NSMenuItem*)menuItem
...
-                SmallTorrentCell* smallCell = [self viewAtColumn:0 row:row makeIfNecessary:NO];
-                smallCell.fIconView.hidden = NO;
-            }
-            update = NO;
+        if ([view isKindOfClass:[TorrentCellActionButton class]]) {
+            SmallTorrentCell* smallCell = [self viewAtColumn:0 row:row makeIfNecessary:NO];
+            smallCell.fIconView.hidden = NO;
         }
+        return;
     }
```
ckerr: Same here — and this is where the placement actually bites:

```suggestion
    if (minimal) {
        if (row >= 0 && [view isKindOfClass:[TorrentCellActionButton class]]) {
            SmallTorrentCell* smallCell = [self viewAtColumn:0 row:row makeIfNecessary:NO];
            smallCell.fIconView.hidden = NO;
        }
        return;
    }
```

The non-minimal path below never reads `row`; it just clears `hoveredTorrent` and repaints. So an exit that really was delivered now gets dropped whenever the view has already been detached, and the tooltip stays painted.

That's reachable in a straight line, not just via a scroll race: `TorrentCellControlButton.mouseDown:` calls `[super mouseDown:]`, which synchronously runs `toggleControlForTorrent:` → `stopTorrents:` → `fullUpdateUI` → `applyFilter` → `removeItemsAtIndexes:`. With a filter active that excludes the just-paused torrent, the row view is detached before control returns, so the next statement's `hoverEventEndedForView:` sees `rowForView:` == -1 and returns before clearing. That torrent's status field then reads "Pause the transfer" instead of its real status, re-asserted every second by `updateUI`, until some unrelated hover clears it.

ckerr:
Not sure why this PR slipped through the cracks, but I was traveling for work 2 weeks ago so maybe that was it :shrug: sorry @lolgear

ckerr (approved):

PR #269, opened 2026-08-26
fix: report the real error when a file can't be opened

webpatser:
Same defect as transmission/transmission#9050, which came out of the transmission/transmission#9049 review; this port carries the review refinements along.

### The bug

`tr_open_files::get()` builds a `tr_error` for every failure branch, logs it, and throws it away. `get_fd()` can't tell an unopenable existing file from a missing one: it returns the empty optional without setting the caller's error, so `write_bytes()` treats the failed open as success and silently discards the data it was asked to write. The pieces are marked complete and later turn up corrupt on verify. The create branch guesses at the cause via `errno`, which `get()` never sets, so it reports whatever unrelated call last set it (on Windows it is never set: `file-win32.cc` reports via `GetLastError()` only).

### The fix

- `tr_open_files::get()` grows a `tr_error` out-param and reports the real failure from all four branches (dir-create, open, preallocate, truncate).
- `get_fd()` uses it for both the open-existing and the create branches, and prefixes the message with the filename via `tr_error::prefix_message()` so the torrent's local error still names the file that broke.
- The truncate branch's log line moves from warn to error to match its behavior: failing there has always aborted the open.

### Tests

New `inout-test.cc` regression test: warm the fd pool with a write that succeeds, evict it with `close_torrent()`, replace the file with a directory, then write another block through `tr_ioWrite`. The second write has to reopen the path, and it must report the failure to its caller; on unfixed `main` it returns success and the data is gone. Turning that error into the torrent's local error and stop is `on_block_written()`'s job since #202, and `TorrentDiskIoTest.failedWriteStopsTorrent` covers that side. Verified the test fails with the fix reverted, and the full libtransmission suite is green with it applied.

Notes: Fixed silent data loss when a torrent's file exists but can't be reopened for writing, e.g. after running out of file descriptors or a permissions change.

### DIFF
--- a/libtransmission/open-files.h
+++ b/libtransmission/open-files.h
@@ -94,7 +94,8 @@ public:
         bool writable,
         std::string_view filename,
         tr_file_preallocation allocation,
-        uint64_t file_size);
+        uint64_t file_size,
+        tr_error& error);
 
     void close_all();
     void close_torrent(tr_torrent_id_t tor_id);
--- a/tests/libtransmission/inout-test.cc
+++ b/tests/libtransmission/inout-test.cc
@@ -1,0 +1,70 @@
+// This file copyright Transmission authors and contributors.
+// It may be used under GPLv2 (SPDX: GPL-2.0-only), GPLv3 (SPDX: GPL-3.0-only),
+// or any future license endorsed by Mnemosaic LLC.
+// License text can be found in the licenses/ folder.
+
+#include <chrono>
+#include <functional>
+#include <future>
+#include <memory>
+#include <utility>
+#include <vector>
+
+#include <gtest/gtest.h>
+
+#include <libtransmission/transmission.h>
+
+#include <libtransmission/block-info.h>
+#include <libtransmission/file.h>
+#include <libtransmission/inout.h>
+#include <libtransmission/torrent.h>
+
+#include "test-fixtures.h"
+
+using InOutTest = tr::test::SessionTest;
+
+TEST_F(InOutTest, writeFailsWhenExistingFileCannotBeOpened)
+{
+    auto* const tor = zeroTorrentInit(ZeroTorrentState::Complete);
+    auto constexpr MaxWaitMsec = 5000;
+
+    auto const path = tr_torrentFindFile(tor, 0U);
+    ASSERT_FALSE(std::empty(path));
+
+    // tr_ioWrite() must run in the session thread; block until it's done
+    auto const run_in_session_thread = [this, MaxWaitMsec](std::function<void()> fn) {
+        auto const promise = std::make_shared<std::promise<void>>();
+        auto future = promise->get_future();
+        session_->run_in_session_thread([fn = std::move(fn), promise]() {
+            fn();
+            promise->set_value();
+        });
+        ASSERT_EQ(std::future_status::ready, future.wait_for(std::chrono::milliseconds{ MaxWaitMsec }));
+    };
+
+    auto err = tr_error_code_t{};
+    auto const write_block = [session = session_, tor, &err]() {
+        auto const buf = std::vector<uint8_t>(tr_block_info::BlockSize);
+        err = tr_ioWrite(*tor, session->openFiles(), tor->block_loc(0U), buf);
+    };
+
+    // Neither the fixture nor verify go through the fd pool, so warm it
+    // with a write that succeeds: this is the steady state of a torrent
+    // that's been receiving blocks from peers.
+    run_in_session_thread(write_block);
+    ASSERT_EQ(0, err);
+
+    // Evict the pooled fd, then make the path unopenable by replacing
+    // the file with a directory.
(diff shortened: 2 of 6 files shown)


tests/libtransmission/inout-test.cc
```diff
@@ -0,0 +1,54 @@
...
+    // make the torrent's first file unopenable by replacing it with a directory
+    auto const path = tr_torrentFindFile(tor, 0U);
+    ASSERT_FALSE(std::empty(path));
+    ASSERT_TRUE(tr_sys_path_remove(path.c_str()));
+    ASSERT_TRUE(tr_sys_dir_create(path.c_str(), 0, 0700));
+
+    // The write must fail and set a local error: reporting success would
+    // make the caller discard the data, leaving pieces that later fail
+    // verification. (Run in the session thread; drop the file descriptors
+    // that verify left open first, so the write has to reopen the
```
ckerr: This says the test drops "the file descriptors that verify left open", but verify never uses the tr_open_files pool, so `close_torrent()` clears nothing and the test does not exercise the pooled-fd path?

webpatser: @ckerr good catch, the comment was wrong. Verify opens with `tr_sys_file_open` directly and the fixture does the same, so the pool was empty and that `close_torrent()` was a no-op. The assertions still hit the fixed branch (pool miss, `find_file` sees the directory, `open_files.get()` fails), but only because nothing had warmed the pool.

Reworked the test in 35b3bb9 so it goes through the pool for real: a first `tr_ioWrite` on the intact file succeeds and caches the fd, then `close_torrent()` evicts it, the file gets replaced with a directory, and the second write has to reopen and fail. Sanity-checked the other way too: without the `close_torrent()` the second write sails through the cached fd and the test fails, so the eviction is now doing something.

webpatser:
Rebased onto main and cleaned up the CI board.

Three jobs were red on the last Sanity run, only one of them mine:

- `clang-tidy` (linux + windows core) flagged two redundant `.c_str()` calls in the new test. `tr_sys_path_remove()` and `tr_sys_dir_create()` take a `string_view`, so those are gone.
- The `completion-test.cc` nodiscard hit and the `interop-busy` readiness race were both already fixed by #279, so the rebase carries them in.

The rebase also picks up #202, which moved the local-error-and-stop reaction out of `tr_ioWrite()` into `on_block_written()`. Dropped the `tr_stat::Error` assertions from my test since `TorrentDiskIoTest.failedWriteStopsTorrent` owns that side now; what's left is the thing this test is actually about, a write that can't reopen an existing file has to tell its caller. Re-checked it the other way too: revert the fix and the test goes red on `err == 0`.

No production changes in this push, just the rebase and the test. Riding on the fresh CI.

libtransmission/open-files.h
```diff
@@ -31,7 +31,8 @@ class tr_open_files
         bool writable,
         std::string_view filename,
         tr_file_preallocation allocation,
-        uint64_t file_size);
+        uint64_t file_size,
+        tr_error* out_error = nullptr);
```
tearfur: Every caller in production code needs the error returned from this method, so let's simplify things a bit. Update test callers to fit them in.

We can remove the local `error` variable in the function body too.

```suggestion
        tr_error& error);
```

tearfur (changes requested):

libtransmission/open-files.h
webpatser: @tearfur done in 21bd2ff: `get()` takes `tr_error&` now, the null checks and the local `error` are gone, and the ten callers in `open-files-test.cc` pass one in. `inout.cc` hands its own out-param straight through. Sanity tests are green locally.

tearfur:
@webpatser Some new merge conflicts arose. Could you PTAL?

webpatser:
Rebased onto main and squashed to one commit, now 7d742e2.

#283 landed while this was in review and rewrote all the files this touches, so this is rebuilt on the new `Handle` API rather than replayed:

- `get()` takes a required `tr_error&` per @tearfur review. Every caller needs the error, so the null checks and the local `error` are gone.
- `get_file()` in `inout.cc` reworked on the new structure. The found branch passes the error through and falls through instead of returning, and the create block is now `else if (writable)` so a failed reopen of an existing file never slides into the create path. Tail keeps the specific message via `prefix_message` and only falls back to ENOENT when the file genuinely is not there.
- `open-files-test.cc` takes main as the base, 14 call sites pass an error. No assertion from main was dropped.

The regression test still bites: without the fix, `write_bytes` gets an empty handle with an unset error and `tr_ioWrite` returns 0, so the test fails. `OpenFilesTest.*` and `InOutTest.*` are green locally, 14 passing.

Notes: report the real error when an existing file cannot be reopened, instead of silently treating the write as successful.

tearfur (approved):
Thanks!

This fix has been a long time coming, because none of the active contributors can reproduce this. Nice work.

PR #270, opened 2026-08-26
fix: upper size limit for magnet metadata

tearfur:
## Description

Notes: Implemented size limits on peer messages to guard against memory exhaustion.

## Checklist

### DIFF
--- a/libtransmission/torrent-magnet.h
+++ b/libtransmission/torrent-magnet.h
@@ -34,7 +34,7 @@ public:
 
     [[nodiscard]] static constexpr auto is_valid_metadata_size(int64_t const size) noexcept
     {
-        return size > 0 && size <= MaxMetadataSize;
+        return 0 < size && size <= MaxMetadataSize;
     }
 
     bool set_metadata_piece(int64_t piece, void const* data, size_t len);
--- a/libtransmission/torrent-magnet.h
+++ b/libtransmission/torrent-magnet.h
@@ -34,7 +34,7 @@ public:
 
     [[nodiscard]] static constexpr auto is_valid_metadata_size(int64_t const size) noexcept
     {
-        return size > 0;
+        return size > 0 && size <= MaxMetadataSize;
     }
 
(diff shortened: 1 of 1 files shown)


libtransmission/torrent-magnet.h
```diff
@@ -73,4 +73,7 @@ class tr_metadata_download
     int64_t piece_count_ = {};
 
     std::string log_name_;
+
+    // Arbitrary upper limit on magnet metadata to defend against memory exhaustion attacks
+    static constexpr auto MaxMetadataSize = 8 * 1024 * 1024; // 8 MiB
```
ckerr, line 78: _(nit)_ Shouldn't be public unless it needs to be? Would either make this private or move it into `is_valid_metadata_size()`.

Otherwise LGTM

libtransmission/torrent-magnet.h
```diff
@@ -34,7 +34,7 @@ class tr_metadata_download
 
     [[nodiscard]] static constexpr auto is_valid_metadata_size(int64_t const size) noexcept
     {
-        return size > 0;
+        return size > 0 && size <= MaxMetadataSize;
```
ckerr: _(nit)_

This is really minor but when doing a "is-in-range" test like this please use this form for readability:

```c++
return 0 < size && size <= MaxMetadataSize;
```

libtransmission/torrent-magnet.h
tearfur: It *is* private though...?

libtransmission/torrent-magnet.h
tearfur: f6aeb8c68ac6eabbcb25a9d9a9abce4d751e17c3

libtransmission/torrent-magnet.h
ckerr: :dizzy_face:

ckerr (approved):

PR #271, opened 2026-08-26
perf: optimize file list reloading and fix cell destruction

lolgear:
## Description
This PR optimizes the data reloading pipeline inside the FileOutlineView (`NSOutlineView`) when switching between items (e.g., torrents/folders). 

Previously, using `reloadData` forced AppKit to completely drop and re-create `NSTableCellView` instances from scratch. This led to unnecessary layout passes, performance drops, and potential out-of-bounds crashes during mouse interaction loops.

### Key Changes:
1. **Moved Selection Reset**: `deselectAll:` is now strictly called *before* mutating the backing array, preventing AppKit from triggering delegate notifications or evaluating layout states on invalid/stale indexes.
2. **Introduced Batch Updates**: Replaced destructive full reloads with an atomic `beginUpdates`/`endUpdates` cycle using `NSTableViewAnimationEffectNone`.
3. **Preserved View Hierarchy**: AppKit now successfully retains existing `NSTableCellView` objects inside the view hierarchy and re-maps them directly via the reuse pool inside `viewForTableColumn:`, dramatically increasing performance.

## How to Test
1. Select a torrent with a large file list in the TorrentsTableView.
2. Observe instantaneous switching without UI stutters or visible cell "blinking".
3. Verify that `Instruments (Allocations / Layout)`

### DIFF
--- a/macosx/FileOutlineController.mm
+++ b/macosx/FileOutlineController.mm
@@ -59,8 +59,9 @@ typedef NS_ENUM(NSUInteger, FilePriorityMenuTag) { //
 
 - (void)setTorrent:(Torrent*)torrent
 {
-    // Save the previous row count to compute the diff for the batch update
-    NSInteger oldRowCount = self.fOutline.numberOfRows;
+    
+    // Close all folders.
+    [self.fOutline collapseItem:nil collapseChildren:YES];
 
     // 1. Update the underlying data source
     _torrent = torrent;
@@ -74,15 +75,16 @@ typedef NS_ENUM(NSUInteger, FilePriorityMenuTag) { //
     [self.fOutline beginUpdates];
 
     // Remove the old rows from the root (parent: nil)
-    if (oldRowCount > 0) {
-        NSIndexSet* oldIndexes = [NSIndexSet indexSetWithIndexesInRange:NSMakeRange(0, oldRowCount)];
+    NSInteger oldRootCount = self.fOutline.numberOfRows;
+    if (oldRootCount > 0) {
+        auto oldIndexes = [NSIndexSet indexSetWithIndexesInRange:NSMakeRange(0, oldRootCount)];
         [self.fOutline removeItemsAtIndexes:oldIndexes inParent:nil withAnimation:NSTableViewAnimationEffectNone];
     }
 
     // Insert the new rows. AppKit will re-map existing NSTableCellViews directly into these new indices
-    NSInteger newRowCount = self.fFileList.count;
+    NSUInteger newRowCount = self.fFileList.count;
     if (newRowCount > 0) {
-        NSIndexSet* newIndexes = [NSIndexSet indexSetWithIndexesInRange:NSMakeRange(0, newRowCount)];
+        auto newIndexes = [NSIndexSet indexSetWithIndexesInRange:NSMakeRange(0, newRowCount)];
         [self.fOutline insertItemsAtIndexes:newIndexes inParent:nil withAnimation:NSTableViewAnimationEffectNone];
     }
 
--- a/macosx/FileOutlineController.mm
+++ b/macosx/FileOutlineController.mm
@@ -59,19 +59,20 @@ typedef NS_ENUM(NSUInteger, FilePriorityMenuTag) { //
 
 - (void)setTorrent:(Torrent*)torrent
 {
-    
-    // Close all folders.
-    [self.fOutline collapseItem:nil collapseChildren:YES];
-
-    // 1. Update the underlying data source
-    _torrent = torrent;
-    [self.fFileList setArray:torrent.fileList ?: @[]];
+    // 1. Clear filter and perform necessary updates.
     self.filterText = nil;
 
-    // 2. Clear selection before mutating the layout to prevent selection artifacts
+    // 2. Close all folders (we need to calculate number of rows).
+    [self.fOutline collapseItem:nil collapseChildren:YES];
+
+    // 3. Clear selection before mutating the layout to prevent selection artifacts
     [self.fOutline deselectAll:nil];
 
-    // 3. Perform a safe batch update to preserve view hierarchy and reuse NSViews
+    // 4. Set new DataSource
+    _torrent = torrent;
+    [self.fFileList setArray:torrent.fileList ?: @[]];
+
+    // 5. Perform a safe batch update to preserve view hierarchy and reuse NSViews
     [self.fOutline beginUpdates];
 
     // Remove the old rows from the root (parent: nil)
@@ -81,7 +82,7 @@ typedef NS_ENUM(NSUInteger, FilePriorityMenuTag) { //
         [self.fOutline removeItemsAtIndexes:oldIndexes inParent:nil withAnimation:NSTableViewAnimationEffectNone];
     }
 
-    // Insert the new rows. AppKit will re-map existing NSTableCellViews directly into these new indices
+    // Insert the new rows.
     NSUInteger newRowCount = self.fFileList.count;
     if (newRowCount > 0) {
         auto newIndexes = [NSIndexSet indexSetWithIndexesInRange:NSMakeRange(0, newRowCount)];
--- a/macosx/FileOutlineController.mm
+++ b/macosx/FileOutlineController.mm
@@ -62,32 +62,28 @@ typedef NS_ENUM(NSUInteger, FilePriorityMenuTag) { //
     // 1. Clear filter and perform necessary updates.
     self.filterText = nil;
 
-    // 2. Close all folders (we need to calculate number of rows).
-    [self.fOutline collapseItem:nil collapseChildren:YES];
-
-    // 3. Clear selection before mutating the layout to prevent selection artifacts
+    // 2. Clear selection before mutating the layout to prevent selection artifacts
     [self.fOutline deselectAll:nil];
 
-    // 4. Set new DataSource
+    // 3. Remember old number of top items.
+    NSUInteger const oldRootCount = self.fFileList.count;
+    auto oldIndexes = [NSIndexSet indexSetWithIndexesInRange:NSMakeRange(0, oldRootCount)];
+
+    // 4. Set new dataSource
     _torrent = torrent;
     [self.fFileList setArray:torrent.fileList ?: @[]];
+
+    NSUInteger const newRowCount = self.fFileList.count;
+    auto newIndexes = [NSIndexSet indexSetWithIndexesInRange:NSMakeRange(0, newRowCount)];
 
     // 5. Perform a safe batch update to preserve view hierarchy and reuse NSViews
     [self.fOutline beginUpdates];
 
     // Remove the old rows from the root (parent: nil)
-    NSInteger oldRootCount = self.fOutline.numberOfRows;
-    if (oldRootCount > 0) {
-        auto oldIndexes = [NSIndexSet indexSetWithIndexesInRange:NSMakeRange(0, oldRootCount)];
-        [self.fOutline removeItemsAtIndexes:oldIndexes inParent:nil withAnimation:NSTableViewAnimationEffectNone];
-    }
+    [self.fOutline removeItemsAtIndexes:oldIndexes inParent:nil withAnimation:NSTableViewAnimationEffectNone];
 
     // Insert the new rows.
-    NSUInteger newRowCount = self.fFileList.count;
-    if (newRowCount > 0) {
-        auto newIndexes = [NSIndexSet indexSetWithIndexesInRange:NSMakeRange(0, newRowCount)];
-        [self.fOutline insertItemsAtIndexes:newIndexes inParent:nil withAnimation:NSTableViewAnimationEffectNone];
-    }
+    [self.fOutline insertItemsAtIndexes:newIndexes inParent:nil withAnimation:NSTableViewAnimationEffectNone];
 
     [self.fOutline endUpdates];
 }
--- a/macosx/FileOutlineController.mm
+++ b/macosx/FileOutlineController.mm
@@ -59,14 +59,34 @@ typedef NS_ENUM(NSUInteger, FilePriorityMenuTag) { //
 
 - (void)setTorrent:(Torrent*)torrent
 {
+    // Save the previous row count to compute the diff for the batch update
+    NSInteger oldRowCount = self.fOutline.numberOfRows;
+
+    // 1. Update the underlying data source
     _torrent = torrent;
-
     [self.fFileList setArray:torrent.fileList ?: @[]];
-
     self.filterText = nil;
 
-    [self.fOutline reloadData];
-    [self.fOutline deselectAll:nil]; //do this after reloading the data #4575
+    // 2. Clear selection before mutating the layout to prevent selection artifacts
+    [self.fOutline deselectAll:nil];
+
+    // 3. Perform a safe batch update to preserve view hierarchy and reuse NSViews
+    [self.fOutline beginUpdates];
+
+    // Remove the old rows from the root (parent: nil)
+    if (oldRowCount > 0) {
+        NSIndexSet* oldIndexes = [NSIndexSet indexSetWithIndexesInRange:NSMakeRange(0, oldRowCount)];
+        [self.fOutline removeItemsAtIndexes:oldIndexes inParent:nil withAnimation:NSTableViewAnimationEffectNone];
+    }
+
+    // Insert the new rows. AppKit will re-map existing NSTableCellViews directly into these new indices
+    NSInteger newRowCount = self.fFileList.count;
+    if (newRowCount > 0) {
+        NSIndexSet* newIndexes = [NSIndexSet indexSetWithIndexesInRange:NSMakeRange(0, newRowCount)];
+        [self.fOutline insertItemsAtIndexes:newIndexes inParent:nil withAnimation:NSTableViewAnimationEffectNone];
+    }
+
+    [self.fOutline endUpdates];
 }
 
 - (void)setFilterText:(NSString*)text


macosx/FileOutlineController.mm
```diff
@@ -59,14 +59,34 @@ - (FileOutlineView*)outlineView
 
 - (void)setTorrent:(Torrent*)torrent
 {
-    _torrent = torrent;
+    // Save the previous row count to compute the diff for the batch update
+    NSInteger oldRowCount = self.fOutline.numberOfRows;
```
ckerr: `numberOfRows` is the flattened count of *visible* rows, but `removeItemsAtIndexes:inParent:nil` indexes the root's **children**. The two diverge as soon as any folder is expanded, so switching torrents with an expanded folder asks AppKit to remove more root children than exist, and `endUpdates` fails its consistency check.

The count you want is the model's root count, sampled before the `setArray:` below.

```suggestion
    NSUInteger const oldRowCount = self.fFileList.count;
```

The comment on the line above needs to change with it — what's worth recording is that these indexes address the root's children, not rows. (`oldChildCount` would read better, but renaming here would break the suggestion.)

macosx/FileOutlineController.mm
```diff
@@ -59,14 +59,34 @@ - (FileOutlineView*)outlineView
...
 
+    // 1. Update the underlying data source
+    _torrent = torrent;
     [self.fFileList setArray:torrent.fileList ?: @[]];
-
     self.filterText = nil;
```
ckerr, line 63: `setFilterText:` is not a plain state reset. When `_filterText` is non-nil it runs its own `beginUpdates`/`moveItemAtIndex:`/`removeItemsAtIndexes:`/`insertItemsAtIndexes:`/`endUpdates` cycle against `self.fFileList` — which the line above just replaced wholesale. So it diffs the new torrent's list against itself and sends the resulting mutations to an outline that still holds the old torrent's rows.

That desync predates this PR, but the `reloadData` being removed here resynced the view unconditionally and hid it.

Reachable path: `AddWindowController.mm:245` and `:369` set `torrent = nil` when the Add window closes, and the user may have typed in the file filter field first (`AddWindowController.mm:253`).

Resetting the filter first — above the count capture, before anything else is mutated — keeps every subsequent step consistent.

macosx/FileOutlineController.mm
```diff
@@ -59,14 +59,34 @@ - (FileOutlineView*)outlineView
...
+    // 2. Clear selection before mutating the layout to prevent selection artifacts
+    [self.fOutline deselectAll:nil];
+
+    // 3. Perform a safe batch update to preserve view hierarchy and reuse NSViews
+    [self.fOutline beginUpdates];
+
+    // Remove the old rows from the root (parent: nil)
+    if (oldRowCount > 0) {
+        NSIndexSet* oldIndexes = [NSIndexSet indexSetWithIndexesInRange:NSMakeRange(0, oldRowCount)];
+        [self.fOutline removeItemsAtIndexes:oldIndexes inParent:nil withAnimation:NSTableViewAnimationEffectNone];
```
ckerr: `oldRowCount` was sampled at line 63, but `self.filterText = nil` on line 68 can itself remove and insert root rows in the outline. By the time this line runs the count can be stale in either direction, independently of the expanded-folder mismatch.

macosx/FileOutlineController.mm
```diff
@@ -59,14 +59,34 @@ - (FileOutlineView*)outlineView
...
+    // 1. Update the underlying data source
+    _torrent = torrent;
     [self.fFileList setArray:torrent.fileList ?: @[]];
-
     self.filterText = nil;
 
-    [self.fOutline reloadData];
-    [self.fOutline deselectAll:nil]; //do this after reloading the data #4575
+    // 2. Clear selection before mutating the layout to prevent selection artifacts
+    [self.fOutline deselectAll:nil];
```
ckerr, line 66: The line this replaces carried `//do this after reloading the data #4575`: the ordering was deliberate and fixed a reported bug. This drops both the ordering and the record of why it existed.

Concretely, `deselectAll:` synchronously fires `outlineViewSelectionDidChange:`, which reloads the Quick Look panel. At this point `fFileList` is already the new torrent's list while the outline still reports the old torrent's rows, so a delegate that maps selected rows back to items sees a mismatched pair — the shape of #4575.

If the new ordering is safe, the comment should say why #4575 no longer applies.

macosx/FileOutlineController.mm
```diff
@@ -59,14 +59,34 @@ - (FileOutlineView*)outlineView
...
+
+    // 3. Perform a safe batch update to preserve view hierarchy and reuse NSViews
+    [self.fOutline beginUpdates];
```
ckerr, line 80: Worth stating the premise, since the PR rests on it: `reloadData` does not drop and re-create `NSTableCellView` instances from scratch. It returns row views to the reuse queue and re-requests them through `outlineView:viewForTableColumn:item:` — the same `makeViewWithIdentifier:` pool this path uses.

Removing every root child and inserting a fresh set yields the same number of `viewForTableColumn:` calls and the same number of `heightOfRowByItem:` queries, plus batch-update bookkeeping. The trade is a one-line primitive that cannot desync for an equivalent that must match the view's cached state exactly — and the comments above are what that costs.

If there is an Instruments trace showing a real win, it belongs in the PR description. Otherwise this reads like it should stay `reloadData`.

Separately: the commit subject promises "fix cell destruction", but I don't see anything in the diff addressing the out-of-bounds crash the description mentions.

macosx/FileOutlineController.mm
```diff
@@ -59,14 +59,34 @@ - (FileOutlineView*)outlineView
...
+
+    // Remove the old rows from the root (parent: nil)
+    if (oldRowCount > 0) {
+        NSIndexSet* oldIndexes = [NSIndexSet indexSetWithIndexesInRange:NSMakeRange(0, oldRowCount)];
+        [self.fOutline removeItemsAtIndexes:oldIndexes inParent:nil withAnimation:NSTableViewAnimationEffectNone];
+    }
+
+    // Insert the new rows. AppKit will re-map existing NSTableCellViews directly into these new indices
```
ckerr: AppKit does not re-map existing cell views into the new indices. Every inserted row is re-requested through `outlineView:viewForTableColumn:item:`, which pulls from the reuse pool and reassigns `node`. This documents a mechanism that isn't there.

```suggestion
```

macosx/FileOutlineController.mm
```diff
@@ -59,14 +59,34 @@ - (FileOutlineView*)outlineView
 
 - (void)setTorrent:(Torrent*)torrent
 {
-    _torrent = torrent;
+    // Save the previous row count to compute the diff for the batch update
+    NSInteger oldRowCount = self.fOutline.numberOfRows;
 
+    // 1. Update the underlying data source
```
ckerr: CLAUDE.md: "Comments justify the current state of the code; commit messages justify the change" and "Write every comment for a reader who has only ever seen this version of the file."

The numbered narration — `// 1. Update the underlying data source`, `// 2. Clear selection before mutating the layout`, `// 3. Perform a safe batch update to preserve view hierarchy and reuse NSViews` — walks a reader through the change rather than justifying the code. "safe" and "preserve" only mean something against the `reloadData` version that this reader never sees.

What earns a comment here is the non-obvious constraint: the remove/insert indexes address the root's children, and the filter has to be reset before either count is taken. The step numbering doesn't.

macosx/FileOutlineController.mm
```diff
@@ -59,14 +59,34 @@ - (FileOutlineView*)outlineView
...
-    [self.fOutline reloadData];
-    [self.fOutline deselectAll:nil]; //do this after reloading the data #4575
+    // 2. Clear selection before mutating the layout to prevent selection artifacts
+    [self.fOutline deselectAll:nil];
+
+    // 3. Perform a safe batch update to preserve view hierarchy and reuse NSViews
+    [self.fOutline beginUpdates];
+
+    // Remove the old rows from the root (parent: nil)
+    if (oldRowCount > 0) {
```
ckerr: `NSMakeRange(0, 0)` yields an empty index set, and both `removeItemsAtIndexes:` and `insertItemsAtIndexes:` are no-ops on one. This guard and the one on line 84 can both go, along with a level of nesting.

macosx/FileOutlineController.mm
```diff
@@ -59,14 +59,34 @@ - (FileOutlineView*)outlineView
...
+        NSIndexSet* oldIndexes = [NSIndexSet indexSetWithIndexesInRange:NSMakeRange(0, oldRowCount)];
+        [self.fOutline removeItemsAtIndexes:oldIndexes inParent:nil withAnimation:NSTableViewAnimationEffectNone];
+    }
+
+    // Insert the new rows. AppKit will re-map existing NSTableCellViews directly into these new indices
+    NSInteger newRowCount = self.fFileList.count;
```
ckerr: `count` is `NSUInteger`, and the file spells unchanging locals `const` (e.g. `NSRange const removeRange` in `setFilterText:`).

```suggestion
    NSUInteger const newRowCount = self.fFileList.count;
```

ckerr (changes requested):
review: claude opus 5 with prompt "/review extra --comment https://github.com/retransmission/retransmission/pull/271"

macosx/FileOutlineController.mm
lolgear: `- (void)reloadData` drops all visibleCells.

Nothing in [documentation](https://developer.apple.com/documentation/appkit/nstableview/reloaddata()) says "put them in reuse queue/pool".

> For [NSView](https://developer.apple.com/documentation/appkit/nsview)-based table views, this method drops all the visible row views and cell views, and re-acquires them all.

macosx/FileOutlineController.mm
lolgear: I don't have access to previous tracker.
Also now it works totally different, because it doesn't modify reload anything (in previous versions this method `outlineViewSelectionDidChange` contains `reloadDataForCells`).
Now this method only reloads data in Quick Look Panel.

It makes sense.

First step: deselect item.
Second step: reload data.

macosx/FileOutlineController.mm
lolgear: Fixed it.

macosx/FileOutlineController.mm
ckerr: `collapseItem:nil collapseChildren:YES` does close the gap: with every root item collapsed, visible rows and root children are the same number, so `numberOfRows` stops overshooting.

It gets there by changing the view rather than by reading the right number, though, and that costs three things.

**Expansion state.** Collapse-all is user-visible, and `setTorrent:` runs on same-torrent resets too. `renameSelected:` posts `ResetInspector` (lines 405 and 413 of this file); `InfoWindowController.resetInfoForTorrent:` calls `resetInfo`, and `InfoFileViewController.setInfoForTorrents:` deliberately does not early-return on an unchanged array ("don't check if it's the same in case the metadata changed", `InfoFileViewController.mm:73`). So `setupInfo` re-enters `setTorrent:` with the identical `Torrent`. `reloadData` kept those folders open, because `Torrent.fileList` is built once and cached (`Torrent.mm:1813`) and NSOutlineView saw the same `FileListNode` objects still expanded. Now renaming a file inside an expanded folder snaps the whole tree shut.

**Distance.** The collapse is on line 66 and the count it corrects is read on line 79, past `deselectAll:` and `setArray:`. Neither site says it depends on the other, so a later reorder restores the original bug with no signal until `endUpdates` throws.

**Cost.** A recursive tree walk plus an un-batched row-removal pass, ahead of the batch update, in a PR about removing layout work.

`self.fFileList.count` sampled above the `setArray:` is the root child count directly: no view mutation, nothing to keep in sync, and the expanded folders survive.

(claude opus 5 extra)

macosx/FileOutlineController.mm
ckerr: Both guards are still here at 3ba01c20a — `if (oldRootCount > 0)` on line 80 and `if (newRowCount > 0)` on line 87.

For what it's worth, the codebase already relies on the empty-index-set no-op in four places, unguarded:

- `Controller.mm:2891` and `:2917` — `removeItemsAtIndexes:` and `insertItemsAtIndexes:` over `NSMakeRange(0, self.fDisplayedTorrents.count)`, `inParent:nil`, inside `beginUpdates`/`endUpdates`. Same class, same two methods, same construction; zero whenever nothing is displayed.
- `MessageWindowController.mm:423` — `clearLog:` on an already-empty log.
- Line 186 of this file — `itemsToAddIndexes` is empty on the common filter-reset path.

One thing to pair with the removal: `oldRootCount` is `NSInteger` while `NSMakeRange`'s length is `NSUInteger`, so the guard incidentally covers a negative value converting to a huge length. `numberOfRows` is never negative, so nothing is actually protected — but making the local `NSUInteger const`, as the thread on line 86 asks, closes that off properly.

(claude opus 5 extra)

macosx/FileOutlineController.mm
ckerr: Partly applied at 3ba01c20a: `newRowCount` is `NSUInteger` now, but neither local is `const`, and `oldRootCount` on line 79 is still `NSInteger`. The same quantity has two types in one method, and the signed one is what feeds `NSMakeRange`'s unsigned length.

(claude opus 5 extra)

macosx/FileOutlineController.mm
ckerr: On `reloadData` — you're right and I was wrong. The documentation says it drops the visible row and cell views rather than returning them to the reuse queue. Withdrawn.

What remains open is the rest of the thread. The description credits this change with a performance win and an out-of-bounds crash fix, and neither is evidenced: the diff contains nothing addressing a crash, and there is no measurement showing the remove/insert pair beats `reloadData`. The collapse-all added in 98f0ebb70 moves that comparison the wrong way — it walks the tree and removes every expanded subtree's rows in its own un-batched pass before the batch update even starts.

If there is an Instruments trace, it belongs in the description. If the crash fix is a separate concern, the commit subject should drop it.

(claude opus 5 extra)

lolgear:
@ckerr 

Could Claude write suggested solution without collapsing top items?
It seems unreasonable complex.

UPD:
Nevermind, it was a lot easier than I expected.

lolgear:
@ckerr 
Could you check PR?

macosx/FileOutlineController.mm
```diff
@@ -59,14 +59,33 @@ - (FileOutlineView*)outlineView
 
 - (void)setTorrent:(Torrent*)torrent
 {
-    _torrent = torrent;
+    // 1. Clear filter and perform necessary updates.
+    self.filterText = nil;
```
ckerr: `self.filterText = nil` runs `setFilterText:`'s whole diff cycle: an O(n*m) `findFileNode:` walk over the old torrent's list, plus a batch update of the outline. Lines 74, 83 and 86 then discard every bit of it.

Assigning the ivar is enough. `setArray:` restores the unfiltered list, and `oldRootCount` still matches the outline either way, because both count root children.

```suggestion
    // Drop the filter; the batch update below replaces every row it would rebuild.
    _filterText = nil;
```

macosx/FileOutlineController.mm
```diff
@@ -59,14 +59,33 @@ - (FileOutlineView*)outlineView
...
+    NSUInteger const oldRootCount = self.fFileList.count;
+    auto oldIndexes = [NSIndexSet indexSetWithIndexesInRange:NSMakeRange(0, oldRootCount)];
+
+    // 4. Set new dataSource
+    _torrent = torrent;
     [self.fFileList setArray:torrent.fileList ?: @[]];
 
-    self.filterText = nil;
+    NSUInteger const newRowCount = self.fFileList.count;
+    auto newIndexes = [NSIndexSet indexSetWithIndexesInRange:NSMakeRange(0, newRowCount)];
```
ckerr: `newRowCount` counts root children, not rows — `insertItemsAtIndexes:inParent:nil` indexes the root's children. `oldRootCount` above already carries the accurate name; conflating the two is what the earlier overshoot turned on.

```suggestion
    NSUInteger const newRootCount = self.fFileList.count;
    auto newIndexes = [NSIndexSet indexSetWithIndexesInRange:NSMakeRange(0, newRootCount)];
```

macosx/FileOutlineController.mm
```diff
@@ -59,14 +59,33 @@ - (FileOutlineView*)outlineView
...
 
-    [self.fOutline reloadData];
-    [self.fOutline deselectAll:nil]; //do this after reloading the data #4575
+    // 5. Perform a safe batch update to preserve view hierarchy and reuse NSViews
```
ckerr: "preserve view hierarchy and reuse NSViews" describes something that does not happen. Every inserted row is re-requested through `outlineView:viewForTableColumn:item:`, which builds or dequeues a view and reassigns `node`. A maintainer who believes this will keep the remove/insert pair through refactors that could simplify it.

The numbering has the same problem in smaller form: steps 3 and 4 each own an uncommented trailing statement now, so the sequence no longer matches the code, and it was already renumbered wholesale once for a single inserted line. Lines 82 and 85 restate the calls beneath them.

Keeping only what the code cannot say:

```suggestion
    [self.fOutline beginUpdates];
```

ckerr (commented):
The count fix is right: `self.fFileList.count` sampled above the `setArray:` is the root child count directly, it needs no view mutation, and dropping `collapseItem:` means expanded folders survive a torrent switch. Three threads resolved.

What's left is documentation, not logic.

The description still credits this change with a performance win and an out-of-bounds crash fix, and states that AppKit "retains existing `NSTableCellView` objects ... and re-maps them directly via the reuse pool". It does not: every inserted row goes back through `outlineView:viewForTableColumn:item:`, which builds or dequeues a view and reassigns `node`. Please drop that claim, add the Instruments trace if there is one, and either show the crash fix in the diff or take it out of the commit subject.

The PR also needs a `Notes:` paragraph — folders reload differently now — or the `notes:none` label. Commits 2 through 4 are missing a Conventional Commits `type:` prefix.

PR #273, opened 2026-08-27
build: require at least GCC 12

tearfur:
## Description

As Debian 11 goes EOL, the oldest GCC our CI covers is 12, so drop all the compatibility code for GCC 11 and below.
~Most notably, we don't need the fast_float library anymore!~

Notes: Require GCC 12 or above.

P.S. Jumping the gun here because I can't wait to drop some baggage.
~P.P.S. We're still alive! https://github.com/transmission/transmission/issues/6742#issuecomment-2020156031~

## Checklist

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -713,7 +713,6 @@ jobs:
             build-base \
             cmake \
             curl-dev \
-            fast_float \
             fmt-dev \
             gettext-dev \
             git \
@@ -1122,7 +1121,6 @@ jobs:
 
           if [ '${{ matrix.use-legacy-packages }}' = 'true' ]; then
             set -- "$@" \
-              -DUSE_SYSTEM_FAST_FLOAT=OFF \
               -DUSE_SYSTEM_FMT=OFF
           else
             set -- "$@" -DCMAKE_EXE_LINKER_FLAGS=-fuse-ld=mold
@@ -1570,7 +1568,6 @@ jobs:
               curl \
               cmake \
               dht \
-              fast_float \
               libevent \
               libfmt \
               libnatpmp \
@@ -1743,7 +1740,6 @@ jobs:
             pkg_add -I -v \
               curl \
               cmake \
-              fast-float \
               fmt \
               gtar-- \
               libevent \
@@ -1827,7 +1823,6 @@ jobs:
             until /usr/sbin/pkg_add -u \
               curl \
               cmake \
-              fast_float \
               fmtlib \
               libevent \
               libpsl \
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -1047,11 +1047,9 @@ jobs:
           - version: oldstable
             gtk: gtk3
             qt: qt5
-            use-legacy-packages: true
           - version: stable
             gtk: gtk4
             qt: qt6
-            use-legacy-packages: false
     container:
       image: debian:${{ matrix.version }}-slim
     steps:
@@ -1060,13 +1058,6 @@ jobs:
           echo '${{ toJSON(needs) }}'
           echo '${{ toJSON(runner) }}'
           cat /etc/os-release
-      - name: Warn if oldoldstable moved past Debian 11
-        if: ${{ matrix.version == 'oldoldstable' }}
-        run: |
-          . /etc/os-release
-          if [ "$VERSION_ID" != '11' ]; then
-            echo "::warning title=Debian oldoldstable changed::debian:oldoldstable currently resolves to Debian $VERSION_ID. Consider removing matrix.use-legacy-packages if legacy package workarounds are no longer needed."
-          fi
       - name: Get Composite Actions
         uses: actions/checkout@v6
         with:
@@ -1080,8 +1071,7 @@ jobs:
           enable-gtk: ${{ needs.what-to-make.outputs.make-gtk == 'true' && matrix.gtk || 'false' }}
           enable-qt: ${{ needs.what-to-make.outputs.make-qt == 'true' && matrix.qt || 'false' }}
           use-sudo: false
-          # Debian 11 (oldoldstable) has no mold package and its GCC 10 predates -fuse-ld=mold.
-          enable-mold: ${{ matrix.use-legacy-packages != true }}
+          enable-mold: true
           enable-ccache: true
       - name: Setup ccache
         uses: ./.github/actions/setup-ccache
@@ -1096,13 +1086,14 @@ jobs:
         run: mkdir src && tar xf transmission*.tar.* -C src --strip-components 1
       - name: Configure
         run: |
-          set -- \
+          cmake \
             -S src \
             -B obj \
             -G Ninja \
             -DCMAKE_C_COMPILER_LAUNCHER=ccache \
             -DCMAKE_CXX_COMPILER_LAUNCHER=ccache \
             -DCMAKE_BUILD_TYPE=Release \
+            -DCMAKE_EXE_LINKER_FLAGS=-fuse-ld=mold \
             -DCMAKE_INSTALL_PREFIX=pfx \
             -DENABLE_CLI=${{ (needs.what-to-make.outputs.make-cli == 'true') && 'ON' || 'OFF' }} \
             -DENABLE_DAEMON=${{ (needs.what-to-make.outputs.make-daemon == 'true') && 'ON' || 'OFF' }} \
@@ -1122,15 +1113,6 @@ jobs:
             -DUSE_SYSTEM_SMALL=OFF \
             -DUSE_SYSTEM_UTP=OFF \
             -DUSE_SYSTEM_WIDE_INTEGER=OFF
-
-          if [ '${{ matrix.use-legacy-packages }}' = 'true' ]; then
-            set -- "$@" \
-              -DUSE_SYSTEM_FMT=OFF
-          else
-            set -- "$@" -DCMAKE_EXE_LINKER_FLAGS=-fuse-ld=mold
-          fi
-
-          cmake "$@"
       - name: Build
         run: cmake --build obj --config Release
       - name: Test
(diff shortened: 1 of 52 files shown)


tearfur:
Aww man `std::from_chars` for floating point types is available starting from clang 20 and macOS 26.

We still have some years to wait before finally dropping fast_float.

.github/workflows/actions.yml
```diff
@@ -1041,14 +1041,16 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - version: oldoldstable
+          # Debian _typically_ releases every 2 years, and each has 5 years of support+LTS.
+          # So the `oldstable` tag is the oldest Docker tag we can use without ever ending up
+          # testing on an EOL release, while not needing periodic updates to the workflow.
+          # Ref: https://wiki.debian.org/DebianReleases
```
ckerr, line 1047: Do we maybe want to keep this to ensure we still build with GCC 12?

tests/libtransmission-app/converter-tests.cc
```diff
@@ -26,20 +26,12 @@ using tr::serializer::to_variant;
...
-
-constexpr std::chrono::sys_seconds make_sys_seconds(int year, int month, int day, int hour, int minute, int second)
-{
-    return std::chrono::sys_seconds{ std::chrono::seconds{ days_from_civil(year, month, day) * 86400 } } +
-        std::chrono::hours{ hour } + std::chrono::minutes{ minute } + std::chrono::seconds{ second };
+    return std::chrono::sys_days{
+        std::chrono::year_month_day{ std::chrono::year{ year }, std::chrono::month{ month }, std::chrono::day{ day } }
+    } + std::chrono::hours{ hour } +
+        std::chrono::minutes{ minute } + std::chrono::seconds{ second };
 }
```
ckerr, line 35: (opus 5 extra)

`parse_sys_seconds` swapped hand-rolled month-length and leap-year validation for `std::chrono::year_month_day::ok()`, and that is the only behaviour-bearing change in the PR -- but `sysSecondsRoundtrip` only feeds it valid dates.

`parse_int` bounds day at 1..31, so rejecting 2023-02-29, 2024-02-30 and 2024-04-31 rests entirely on the new `ymd.ok()` call. Drop that call and the whole suite still passes.

A few negative cases would pin it:

```cpp
EXPECT_FALSE(to_value<std::chrono::sys_seconds>(tr_variant{ "2023-02-29T00:00:00Z"sv }).has_value());
EXPECT_FALSE(to_value<std::chrono::sys_seconds>(tr_variant{ "2024-04-31T00:00:00Z"sv }).has_value());
EXPECT_TRUE(to_value<std::chrono::sys_seconds>(tr_variant{ "2024-02-29T00:00:00Z"sv }).has_value());
```

(The last one is the leap-year case the deleted `is_leap_year` handled by hand.)

libtransmission/macros.h
```diff
@@ -5,18 +5,6 @@
...
-#define TR_CONSTEXPR_VEC
-#endif
-
-#if __cpp_lib_constexpr_string >= 201907L
-#define TR_CONSTEXPR_STR constexpr
-#else
-#define TR_CONSTEXPR_STR
-#endif
-
 #if __cplusplus >= 202302L // _MSVC_LANG value for C++23 not available yet
```
ckerr, line 8: (Opus 5 extra)

While we're doing this, `libtransmission/peer-mse.cc:20-26` is the same species of compat code and survives this pass.

```cpp
// workaround bug in GCC < 10.4
// https://gcc.gnu.org/bugzilla/show_bug.cgi?id=99859
#if __GNUC__ < 10 || (__GNUC__ == 10 && __GNUC_MINOR__ < 4)
#define PRIME_CONSTEXPR const
#else
#define PRIME_CONSTEXPR constexpr
#endif
```

With GCC 12 as the floor the whole block can go and `Prime` (peer-mse.cc:75) becomes plain `constexpr`. That also fixes a live bug: clang reports `__GNUC__ == 4` (verified, `clang++ -E` expands it to `4 2`) and MSVC does not define it at all, so the `#if` is true on both and the 768-bit `wi::key_t Prime` is dynamically initialised rather than constant-folded on every clang and MSVC build -- all the macOS, Windows, BSD and clang-tidy jobs.

libtransmission/clients.cc
```diff
@@ -627,7 +627,7 @@ void tr_clientForId(char* buf, size_t buflen, tr_peer_id_t peer_id)
     // both "-WT" and "-WT-" is formatted by the "-WT-" entry
     if (auto const it = std::ranges::find_if(
             Clients,
-            [key](Client const& client) { return tr_strv_starts_with(key, client.begins_with); });
+            [key](Client const& client) { return key.starts_with(client.begins_with); });
```
ckerr, line 630: We can remove  `libtransmission/string-utils.h` at line 22 of this file now. Same with these other files:

- `gtk/Session.cc:18`
- `libtransmission/clients.cc:22`
- `libtransmission/crypto-utils.cc:25`
- `libtransmission/session.cc:53`
- `libtransmission/variant-benc.cc:26`
- `libtransmission/webseed.cc:25`

libtransmission/bitfield.h
```diff
@@ -131,7 +131,7 @@ class tr_bitfield
     tr_bitfield& operator|=(tr_bitfield const& that);
 
 private:
-    [[nodiscard]] TR_CONSTEXPR_VEC size_t count_flags() const noexcept
+    [[nodiscard]] constexpr size_t count_flags() const noexcept
```
ckerr, line 134: We can remove `#include "libtransmission/macros.h"` from these places now:

- `libtransmission/bitfield.h:18`
- `libtransmission/torrent-queue.h:16`
- `libtransmission/torrent-metainfo.h:17`
- `libtransmission/torrent.h:40`
- `libtransmission/peer-mgr-wishlist.h:22`

(But `libtransmission/torrent.cc` will need to add it)

libtransmission/torrent.cc
```diff
@@ -2359,7 +2359,7 @@ auto renamePath(tr_torrent const* tor, std::string_view oldpath, std::string_vie
 
     if (tr_sys_path_exists(src)) {
         auto const parent = tr_sys_path_dirname(src);
-        auto const tgt = tr_strv_ends_with(src, tr_torrent_files::PartialFileSuffix) ?
+        auto const tgt = src.ends_with(tr_torrent_files::PartialFileSuffix) ?
```
ckerr, line 2362: (opus 5 extra)

Note this `.ends_with` does not resolve to `std::string_view`'s -- `src` is a `tr_pathbuf`, so it binds to `tr_strbuf::ends_with` at `tr-strbuf.h:148`.

That class carries six hand-rolled `starts_with`/`ends_with` overloads reimplementing the exact `size() >= x_len && sv().substr(...) == x` logic this PR just deleted from `string-utils.h`, on the grounds that the standard now provides it. Same argument applies here: the range and char overloads each collapse to a one-line `return sv().ends_with(x);`, leaving only the `Char const*` pair for their extra nullptr guard.

Out of scope for this PR if you would rather keep it focused, but it is the last copy of the code you are removing.

qt/DetailsDialog.cc
```diff
@@ -228,8 +229,7 @@ class PeerItem : public QTreeWidgetItem
                 QByteArray tmp(16, '\0');
 
                 for (int i = 0; i < 16; ++i) {
-                    // NOLINTNEXTLINE(bugprone-narrowing-conversions): TODO(c++20): use std::bit_cast after gcc 11.1
-                    tmp[i] = ipv6_address[i];
+                    tmp[i] = std::bit_cast<char>(ipv6_address[i]);
                 }
```
ckerr, line 233: (opus 5 extra)

`Q_IPV6ADDR` is a `quint8 c[16]` aggregate, so the whole address converts in one `bit_cast` -- the prefill and the loop both go, and the literal `16` stops appearing three times.

```suggestion
                auto const octets = std::bit_cast<std::array<char, sizeof(Q_IPV6ADDR)>>(ipv6_address);
                auto const tmp = QByteArray{ std::data(octets), std::ssize(octets) };
```

Needs `#include <array>` alongside the new `<bit>`. `std::ssize` returns `ptrdiff_t`, which is Qt 6's `qsizetype`, so the braced init does not narrow.

gtk/FilterBar.cc
```diff
@@ -196,12 +196,12 @@ bool FilterBar::Impl::tracker_filter_model_update()
...
         std::string announce_url;
 
-        TR_CONSTEXPR_STR auto operator<=>(site_info const& that) const noexcept
+        constexpr auto operator<=>(site_info const& that) const noexcept
         {
             return sitename <=> that.sitename;
         }
 
-        TR_CONSTEXPR_STR auto operator==(site_info const& that) const
+        constexpr auto operator==(site_info const& that) const
```
ckerr, line 204:
```suggestion
        constexpr auto operator==(site_info const& that) const noexcept
```

ckerr:
Oops, I was too slow :crying_cat_face:

.github/workflows/actions.yml
tearfur: `oldstable` will remain Debian 12 (which is on GCC 12) until the next Debian Stable release, which is most likely next year.

qt/DetailsDialog.cc
tearfur: Here's where the C++ standard is tricky -- It doesn't guarantee `sizeof(std::array<char, 16>) == 16` and `sizeof(Q_IPV6ADDR) == 16`, but these asserts will turn out to be true in any sane implementation.

I'd rather keep the current guaranteed-to-be-correct code rather than relying on UB.

tests/libtransmission-app/converter-tests.cc
tearfur: #291

libtransmission/macros.h
tearfur: #291 Ooh nice catch.

libtransmission/clients.cc
tearfur: #291

libtransmission/bitfield.h
tearfur: #291

libtransmission/torrent.cc
tearfur: #291

gtk/FilterBar.cc
tearfur: #291

Issue #240, opened 2026-08-20
Network Interface Binding


392781:
### What is the issue?

Not sure which project is going to dominate in the future, but I want to throw in that it would be great to implement the following feature which is on the cusp of being approved in the original transmission repo: https://github.com/transmission/transmission/pull/8778

### Which application of Transmission?

macOS app

### Which version of Transmission?

4.1.3

robd003:
@392781 / @ckerr / @tearfur I just rebased the original PR to retransmission https://github.com/retransmission/retransmission/pull/276

PR #276, opened 2026-08-28
feat: network interface binding

robd003:
## Description

Issue: https://github.com/retransmission/retransmission/issues/240

Rebased James Healy's work from https://github.com/jameshealyio/transmission/tree/codex/network-interface-binding to `main` branch.

- add a macOS `bind_interface` setting that binds app-wide traffic to a selected network interface
- add per-torrent `bind_interface` overrides with fail-closed handling for shared UDP paths
- expose the setting through config/docs, daemon settings, RPC, resume data, and macOS UI controls

Notes: Added macOS support for binding Transmission traffic app-wide or per torrent to a selected network interface.

## Testing
- `./code_style.sh`
- `git diff --check`
- `cmake --build build -j8`
- `ctest --test-dir build -R 'LT\.(RpcTest\.(sessionBindInterface|torrentBindInterface|torrentAddBindInterface)|SessionTest\.(propertiesApi|torrentBindInterfaceResume)|SettingsTest\.(canLoadBindInterface|canSaveBindInterface))' --output-on-failure`
- `cmake --build build-mac --target transmission-mac/fast -j8` with `CODESIGN_EXECUTABLE=/usr/bin/true` due local provenance xattrs

## Checklist

- [X] I have manually written or manually audited all changes in this PR and vouch for them.
- [X] I have explained user-facing changes in a 'Notes:' paragraph for the release notes script.


ckerr:
@robd003 TAL at the clang-tidy warnings in https://github.com/retransmission/retransmission/actions/runs/33130976688/job/98720258803?pr=276 , they look valid + easy-to-fix

ckerr:
@robd003 since this is a 2000 LOC diff, are you OK with an AI code review to go over this PR before getting to the human review stage?

A human will go over every line that lands, but this is a huge PR :smile_cat:

robd003:
> @robd003 since this is a 2000 LOC diff, are you OK with an AI code review to go over this PR before getting to the human review stage?
> 
> A human will go over every line that lands, but this is a huge PR 😸

No worries, I'm ready for our robot overlords.

robd003:
> @robd003 TAL at the clang-tidy warnings in https://github.com/retransmission/retransmission/actions/runs/33130976688/job/98720258803?pr=276 , they look valid + easy-to-fix

@ckerr Fixed in commit https://github.com/retransmission/retransmission/pull/276/commits/2b9bbe58840c5c89dd93e57e31edd46682f97e1d

libtransmission/peer-io.cc
```diff
@@ -250,7 +255,7 @@ bool tr_peerIo::reconnect()
 
     close();
 
-    auto s = tr_peer_socket_tcp::create(*session_, socket_address(), client_is_seed());
+    auto s = tr_peer_socket_tcp::create(*session_, socket_address(), client_is_seed(), bind_interface_);
```
ckerr: **Peer sockets from `new_outgoing_utp()` reconnect without an interface binding.**

`tr_peerIo::new_outgoing_utp()` (peer-io.cc:166) is the one construction path that never populates `bind_interface_` — `new_outgoing()` assigns `io->bind_interface_` right after `create()` (peer-io.cc:155), and `create()` itself passes `{}`. So a peerIo built by the BEP-55 holepunch path (peer-mgr.cc:2851) carries an empty name into this line.

When the MSE handshake fails and handshake.cc:622/639 calls `io->reconnect()`, `tr_netSetSocketInterface` sees the empty name and returns `true` immediately, so the replacement TCP socket is opened on the system default route even though the session is bound to an interface. That is a plain traffic leak off the bound interface.

`tr_peerMgrConnectHolepunch` already refuses torrents whose bind interface differs from the session's, but it does not cover the *session-level* binding, which is exactly the case that leaks here.

No single-line suggestion: the fix belongs in `new_outgoing_utp()` — give it a `std::string_view bind_interface` parameter, set `io->bind_interface_` from it the way `new_outgoing()` does, and pass `tor->effective_bind_interface()` at the peer-mgr.cc:2851 call site.

libtransmission/announcer.cc
```diff
@@ -185,6 +185,21 @@ class tr_announcer_impl final : public tr_announcer
             tr_strv_starts_with(announce_sv, "http://"sv) || tr_strv_starts_with(announce_sv, "https://"sv)) {
             tr_tracker_http_announce(session, request, std::move(on_response));
         } else if (tr_strv_starts_with(announce_sv, "udp://"sv)) {
+            if (auto const* const tor = session->torrents().get(request.info_hash);
+                tor != nullptr && !tor->bind_interface_matches_session()) {
```
ckerr: **Scrapes bypass the bind interface entirely — both UDP and HTTP.**

This guard covers `announce()`, and `tr_tracker_http_announce` gets the per-torrent override at announcer-http.cc:260. `scrape()` (announcer.cc:166) has neither:

- it calls `announcer_udp_.scrape()` unconditionally, over the shared UDP socket bound to the *session's* interface;
- `tr_tracker_http_scrape` (announcer-http.cc:498) never sets `options.bind_interface`, because `tr_scrape_request` has no such field.

With a session default route and a torrent bound to `utun4`, every scrape still sends that torrent's info_hash to the tracker over the default route. docs/Editing-Configuration-Files.md promises "If a configured interface is unavailable, Transmission fails closed instead of falling back to the default route" — scrapes do the opposite.

No suggestion here: `tr_scrape_request` needs a `bind_interface` field (mirroring the one added to `tr_announce_request`), populated where the request is built, consumed in `tr_tracker_http_scrape`, and a matching skip in the `udp://` branch of `scrape()`. Note multiscrape batches several info_hashes into one request, so the batching in `tr_announcer_impl::scrapeAndAnnounceMore` has to group by effective interface (or exclude interface-bound torrents from batching) for the UDP skip to be correct.

libtransmission/port-forwarding.cc
```diff
@@ -169,8 +171,11 @@ class tr_port_forwarding_impl final : public tr_port_forwarding
...
     {
         auto const is_enabled = is_enabled_ && !is_shutting_down_;
+        auto const bind_interface = mediator_.bind_interface();
+        auto const is_interface_bound = !tr_net_interface_is_default(bind_interface);
 
-        if (!natpmp_) {
+        // NAT-PMP has no per-interface control, so skip it when bound to an interface
+        if (!is_interface_bound && !natpmp_) {
             natpmp_ = std::make_unique<tr_natpmp>();
         }
```
ckerr: **NAT-PMP is only skipped if it was never constructed; enabling the binding at runtime leaves it running.**

The comment says NAT-PMP is skipped when bound to an interface, but the check only guards *construction*. In the common case the process starts with no bind interface, so `natpmp_` is created here; the user then picks an interface in Preferences, and on the next pulse `natpmp_` is still non-null, so line 188's branch keeps calling `natpmp_->pulse(...)` and sending NAT-PMP UDP to the gateway over the default route — the leak this comment says is prevented. `natpmp_` is never reset, so it persists until restart.

```suggestion
        // NAT-PMP has no per-interface control, so it can't run while bound to an interface
        if (is_interface_bound) {
            natpmp_.reset();
        } else if (!natpmp_) {
            natpmp_ = std::make_unique<tr_natpmp>();
        }
```

With that, `is_enabled && !is_interface_bound` on line 189 becomes redundant and can go back to plain `is_enabled`. Trade-off worth calling out in the PR description: resetting rather than unmapping leaves any existing NAT-PMP mapping to expire on its own lease, which is the right call for a fail-closed feature but does mean the mapping outlives the setting change.

ckerr (commented):
(claude opus 5, prompt: "/review extra --comment https://github.com/retransmission/retransmission/pull/276")

Review of the network-interface binding feature (`2b9bbe588`). 15 findings, all inline; every one is caused by this PR rather than pre-existing, so there is no separate follow-up group.

The dominant theme is **gaps in the fail-closed guarantee**: five paths still send traffic over the default route when a bind interface is configured — the `reconnect()` fallback for uTP-created peerIos, tracker scrapes (UDP and HTTP), NAT-PMP after the setting is toggled at runtime, LPD multicast egress, and in-flight handshakes that survive an interface change.

Two others stand out: `tr_torrent::set_bind_interface` tears down peer connections on the caller's thread, which the macOS inspector calls from the main thread; and the synthesized UDP-skip announce response reports "Could not connect to tracker" instead of its own message.

The rest are an IPv6 bind-address fallback regression, a per-socket warning flood on non-Apple platforms, the `tr-vpn-none` sentinel design, the missing rpc-spec changelog entries, two macOS layout issues, a doubled SystemConfiguration scan, and missing tests for the predicates the whole design rests on.

libtransmission/tr-lpd.cc
```diff
@@ -291,6 +291,10 @@ class tr_lpd_impl final : public tr_lpd
             return false;
         }
 
+        if (!tr_netSetSocketInterface(sock, ip_protocol, mediator_.bind_interface())) {
```
ckerr, line 294: **`IP_BOUND_IF` here doesn't reach LPD's multicast traffic — the explicit multicast-interface options still select the system default.**

Further down, this same function pins the multicast interface explicitly:

- IPv4, line 329: `mcast_req.imr_interface = mediator_.bind_address(ip_protocol).addr.addr4;` then `IP_MULTICAST_IF`
- IPv6, line 366: `mcast_req.ipv6mr_interface = mediator_.bind_address(ip_protocol).to_interface_index().value_or(0);` then `IPV6_JOIN_GROUP` and `IPV6_MULTICAST_IF`

`tr_session::bind_address(TR_AF_INET6)` now returns `tr_address::any(TR_AF_INET6)` for *any* non-default bind interface (session.cc:431), and the IPv4 side is `0.0.0.0` unless the user set `bind_address_ipv4`. So both resolve to "system default": index 0 / `INADDR_ANY`. An explicit `IP_MULTICAST_IF`/`IPV6_MULTICAST_IF` takes precedence over `IP_BOUND_IF` for multicast destinations, so LPD announces still egress on the default interface — with the peer port and info_hash prefix in them.

No suggestion: this needs the interface index plumbed through, e.g. add `bind_interface_index()` to `tr_lpd::Mediator` (resolving via `if_nametoindex`) and use it for `imr_interface`/`ipv6mr_interface` when it is non-zero, falling back to the current `bind_address`-derived value otherwise.

libtransmission/torrent.cc
```diff
@@ -1849,6 +1856,41 @@ void tr_torrent::set_bandwidth_group(std::string_view group_name)
...
+
+    if (bind_interface_ == bind_interface) {
+        return;
+    }
+
+    bind_interface_ = bind_interface;
+    set_dirty();
+    mark_changed();
+
+    session->closeTorrentPeerConnections(this);
```
ckerr: **This tears down peer connections on the caller's thread, which on macOS is the AppKit main thread.**

`tr_peerMgrCloseTorrentConnections` calls `swarm->remove_peer()` for every peer, destroying `tr_peerMsgs`/`tr_peerIo` objects and their libevent events. The macOS UI reaches here synchronously off the main thread: InfoOptionsViewController.mm `setBindInterface:` → `Torrent.setBindInterface:` (Torrent.mm:484) → `tr_torrentSetBindInterface` (torrent.cc:1935) → here. The session thread can be reading those same objects concurrently, so this is a data race and a use-after-free, not just a lock-ordering nit — `unique_lock()` on line 1863 guards the session mutex, not libevent's per-socket state.

Every other peer-touching torrent operation marshals first: `tr_torrent::start` (torrent.cc:613), `tr_torrentStop` (721), `tr_torrentVerify` (1532). `tr_sessionSetBindInterface` added in this same PR (session.cc:1615) does too — this is the one setter that doesn't.

```suggestion
    session->run_in_session_thread([session = this->session, tor_id = id()]() {
        if (auto* const tor = session->torrents().get(tor_id); tor != nullptr) {
            session->closeTorrentPeerConnections(tor);
        }
    });
```

The id round-trip (rather than capturing `this`) also covers the torrent being removed between the setter and the closure running.

libtransmission/session.cc
```diff
@@ -804,10 +837,16 @@ void tr_session::setSettings(tr_session::Settings&& settings_in, bool force)
...
         dht_.reset();
-    } else if (force || !dht_ || port_changed || addr_changed || new_settings.dht_enabled != old_settings.dht_enabled) {
+    } else if (
+        force || !dht_ || port_changed || addr_changed || interface_changed ||
+        new_settings.dht_enabled != old_settings.dht_enabled) {
         dht_ = tr_dht::create(dht_mediator_, advertisedPeerPort(), udp_core_->socket4(), udp_core_->socket6());
     }
 
+    if (interface_changed) {
+        tr_peerMgrCloseConnections(peer_mgr_.get());
```
ckerr, line 894: **Changing the interface closes established peers but leaves in-flight handshakes on the old sockets.**

`tr_peerMgr::close_connections` iterates `tor->swarm->peers` only (peer-mgr.cc:2238). Two other sets of sockets survive:

- `tr_peerMgr::incoming_handshakes`, accepted on the *previous* listening socket;
- the outgoing handshakes owned by each `tr_peer_info`, started at peer-mgr.cc:2740 on sockets opened with the previous interface.

When those complete, `on_handshake_done` promotes them to full peers, so BitTorrent traffic keeps flowing on the unbound interface after the user asked for a bound one — for as long as a handshake takes to finish, and then indefinitely for the resulting peers.

No suggestion: `close_connections()` needs to also clear `incoming_handshakes` and abort each peer_info's pending handshake (there is already an abort path used when a torrent stops). Worth doing in `close_torrent_connections()` too, so the per-torrent setter has the same guarantee.

libtransmission/announcer.cc
```diff
@@ -185,6 +185,21 @@ class tr_announcer_impl final : public tr_announcer
...
+                tr_logAddWarn(
+                    fmt::format(
+                        fmt::runtime(_(
+                            "Skipping UDP tracker announce for '{torrent}' because its bind interface cannot be enforced on the shared UDP socket")),
+                        fmt::arg("torrent", request.log_name)));
+
+                auto response = tr_announce_response{};
```
ckerr: **This response reports "Could not connect to tracker", not the message on line 198.**

`tr_announce_response::did_connect` defaults to `false` (announcer-common.h:107), and `onAnnounceDone` tests it first:

```
if (!response.did_connect) {
    on_announce_error(tier, _("Could not connect to tracker"), event);
} else if (response.did_timeout) {
...
} else if (!std::empty(response.errmsg)) {
```

So `errmsg` is never reached and the carefully-worded explanation is dead code — the user sees a generic connection failure and has no way to tell an interface-policy skip from a dead tracker.

```suggestion
                auto response = tr_announce_response{};
                response.did_connect = true;
```

libtransmission/session.cc
```diff
@@ -414,11 +424,18 @@ tr_address tr_session::bind_address(tr_address_type type) const noexcept
 
     if (type == TR_AF_INET6) {
         // if user provided an address, use it.
+        if (!std::empty(settings_.bind_address_ipv6)) {
+            return tr_address::from_string(settings_.bind_address_ipv6).value_or(tr_address::any(TR_AF_INET6));
+        }
+        if (!tr_net_interface_is_default(settings_.bind_interface)) {
+            return tr_address::any(TR_AF_INET6);
+        }
```
ckerr: **Behaviour change: an invalid `bind_address_ipv6` now binds `::` instead of the global source address.**

Before this hunk the whole branch was `tr_address::from_string(settings_.bind_address_ipv6).value_or(default_addr)`, where `default_addr` is the source address computed on lines 436-438. Now a non-empty but unparseable value takes this early return and falls back to `tr_address::any(TR_AF_INET6)`, skipping the source-address path entirely.

That contradicts docs/Editing-Configuration-Files.md, which this PR leaves unchanged: "When no valid IPv6 address is provided, Transmission will try to bind to your default global IPv6 address. If that didn't work, then Transmission will bind to `::`."

Separately, the comment block is now split around the new branch: "if user provided an address, use it." sits above lines 427-429, but the "otherwise, if we can determine which one to use..." lines are separated from it by the interface check, so they no longer read as the continuation they were written as.

```suggestion
        // if user provided a valid address, use it.
        if (auto const addr = tr_address::from_string(settings_.bind_address_ipv6); addr) {
            return *addr;
        }

        // when bound to an interface, let that binding choose the source address
        if (!tr_net_interface_is_default(settings_.bind_interface)) {
            return tr_address::any(TR_AF_INET6);
        }
```

This keeps the new interface behaviour, restores the documented fallback for a malformed address, and puts each comment on the branch it describes.

libtransmission/net.cc
```diff
@@ -106,6 +106,71 @@ int tr_make_listen_socket_ipv6only(tr_socket_t const sock)
...
+        return false;
+    }
+
+    tr_logAddTrace(fmt::format("Bound socket {} to interface '{}'", sock, name));
+    return true;
+#else
+    if (!suppress_msgs) {
+        tr_logAddWarn(fmt::format("Interface binding is not supported on this platform: '{}'", name));
+    }
+    return false;
```
ckerr, line 222: **Two problems on the unsupported-platform path: unbounded warning rate, and `sockerrno` left stale.**

1. Every caller hits this on Linux/Windows once a `bind_interface` is set — each outgoing peer connect (peer-socket-tcp.cc:90), each listen bind (net.cc:236), both UDP sockets (tr-udp.cc:158, 205), and each ip-cache source-address probe. The log fills at connection-attempt rate with the same line.
2. The `__APPLE__` branch sets `ENXIO` via `set_sockerrno` (line 136) before returning false; this branch sets nothing. Callers then record a leftover errno: `tr_netBindTCPImpl` does `*err_out = sockerrno` (net.cc:237) and `get_source_address` does `err_out = sockerrno` (ip-cache.cc:85), so the reported failure reason is unrelated to what actually happened.

```suggestion
#else
    set_sockerrno(ENXIO);
    if (!suppress_msgs) {
        tr_logAddDebug(fmt::format("Interface binding is not supported on this platform: '{}'", name));
    }
    return false;
```

The user-visible warning is better emitted once from `tr_session::setSettings`, when `bind_interface` becomes non-default on a platform that can't honour it — one actionable line instead of a stream. Note `set_sockerrno` is a no-op on Windows (net.h:53), so the errno half only helps on POSIX; that's still strictly better than propagating a stale value.

macosx/Utils.mm
```diff
@@ -3,9 +3,33 @@
...
+
+#import <AppKit/AppKit.h>
+#import <SystemConfiguration/SystemConfiguration.h>
 
 #import "Utils.h"
 
+NSString* const TRBindInterfaceModeLiteral = @"literal";
+NSString* const TRBindInterfaceModeActiveVPN = @"active-vpn";
+NSString* const TRBindInterfaceActiveVPNMenuValue = @"__transmission_active_vpn__";
+NSString* const TRNoActiveVPNBindInterfaceName = @"tr-vpn-none";
```
ckerr: **The fail-closed state is encoded as a fake interface name, which turns a policy decision into a per-socket error.**

`tr-vpn-none` is handed to libtransmission as if it were a real interface (Controller.mm, in both the init path and `refreshActiveVPNBindInterfacePreference`). Every socket then reaches `if_nametoindex("tr-vpn-none")` == 0 in net.cc:135 and logs `Couldn't bind socket to interface 'tr-vpn-none': interface not found` — once per peer connect, per listen bind, per UDP socket, per ip-cache poll. With Active-VPN mode selected and no VPN up, that is a continuous stream of warnings that reads as a misconfiguration rather than as "traffic is intentionally blocked". It also means a real interface named `tr-vpn-none` would be silently reinterpreted.

No suggestion: this wants a first-class blocked state in libtransmission — e.g. a `bind_interface` value the core recognises as "refuse all sockets", which logs once, is reportable over RPC, and lets the UI show why nothing is connecting. That also gives the GTK and Qt clients something to render if this feature grows beyond macOS.

docs/rpc-spec.md
```diff
@@ -583,6 +586,7 @@ Response parameters: `path`, `name`, and `id`, holding the torrent ID integer
 | `alt_speed_time_end` | number | when to turn off alt speeds (units: same)
 | `alt_speed_up` | number | max global upload speed (kB/s)
 | `anti_brute_force_enabled` | boolean | true means to enable a basic brute force protection for RPC server
+| `bind_interface` | string | session network interface binding; empty means normal OS routing
```
ckerr: **These new args aren't in the version changelog.**

The PR adds `bind_interface` to five methods — `session_get` (this line), `session_set`, `torrent_get` (line 244), `torrent_set` (line 170) and `torrent_add` (line 517) — but the "Transmission 4.2.0 (`rpc_version_semver` 6.1.0, `rpc_version`: 20)" table at line 1144 is untouched. Every other arg added in that release (`is_folder`, `webseeds_ex`, `blocklist_updates_enabled`, `torrent_complete_verify_enabled`, ...) has an entry there, and third-party clients read that table to discover new capabilities.

Line 1144 is outside this PR's diff so this can't be a suggestion; the rows to add to that table are:

```
| `session_get` | new arg `bind_interface`
| `session_set` | new arg `bind_interface`
| `torrent_add` | new arg `bind_interface`
| `torrent_get` | new arg `bind_interface`
| `torrent_set` | new arg `bind_interface`
```

macosx/PrefsController.mm
```diff
@@ -417,6 +447,208 @@ - (void)setNat:(id)sender
...
+- (void)setupBindInterfaceStatusControls
+{
+    if (self.fBindInterfaceStatusField != nil || self.fBindInterfacePopUp.superview == nil) {
+        return;
+    }
+
+    NSView* container = self.fBindInterfacePopUp.superview;
+    NSButton* peerCommunicationButton = nil;
+    for (NSView* view in container.subviews) {
+        if ([view isKindOfClass:NSButton.class] && [(NSButton*)view action] == @selector(setUTP:)) {
```
ckerr: **This locates a control by selector comparison and then undoes a constraint the same PR adds to the XIB.**

PrefsWindow.xib gains `v05-Z8-iS7` (`2077.top == BIf-Pu-001.bottom + 28`) in this PR; a few lines later this method searches `container.constraints` for anything pairing that button with `fBindInterfacePopUp` (line 467), deactivates it, and re-adds an equivalent anchored to a programmatically created status label. The button itself is identified by `[(NSButton*)view action] == @selector(setUTP:)`.

Nothing here fails at build time. Renaming `setUTP:`, reordering the Network pane, or adjusting that constraint in Interface Builder silently produces a broken or overlapping layout, and the constraint in the XIB now exists only to be removed at launch.

No suggestion: the status label and info button belong in PrefsWindow.xib next to `BIf-Pu-001`, wired through `IBOutlet`s the way `fBindInterfacePopUp` already is. Then the XIB owns the whole Network pane layout and this method reduces to setting the label's string.

macosx/AddWindowController.mm
```diff
@@ -200,6 +203,39 @@ - (void)setDestination:(id)sender
     }];
 }
 
+- (void)setupBindInterfaceMenu
```
ckerr: **Duplicated with AddMagnetWindowController.mm:133, and both position views with literal coordinates.**

The two copies differ only in how the label's x-origin is derived (`148.0` here, `NSMaxX(self.fGroupPopUp.frame) + 20.0` there). Both leave `translatesAutoresizingMaskIntoConstraints` on and hardcode `NSMakeRect(148.0, y, 84.0, 17.0)` / `NSMakeRect(236.0, y - 5.0, 108.0, 26.0)` into windows whose other controls come from XIBs with autolayout.

Two consequences: an 84pt label clips any localisation of "Connection:" longer than the English string, and any future XIB change to the group popup's position silently misplaces this row — in two files that have to be fixed separately.

No suggestion: put the label and popup in AddWindow.xib and AddMagnetWindow.xib with `IBOutlet`s, as `fGroupPopUp` and `fPriorityPopUp` already are. If they must stay programmatic, hoist the shared construction next to `TRPopulateBindInterfacePopUp` in Utils.mm and use layout anchors against `fGroupPopUp` instead of absolute rects.

macosx/Utils.mm
```diff
@@ -17,3 +41,547 @@ bool isRatioEqual(CGFloat old_ratio, CGFloat new_ratio)
...
+                }
+            }
+        }
+
+        CFRelease(prefs);
+    }
+
+    [names addEntriesFromDictionary:TRUserDefinedNamesByInterfaceName()];
+
+    NSDictionary<NSString*, NSDictionary<NSString*, NSString*>*>* vpnServices = TRVPNServiceInfoByInterfaceName();
```
ckerr: **`TRUserDefinedNamesByInterfaceName()` runs twice per call — once here, once inside `TRVPNServiceInfoByInterfaceName()`.**

That function creates an `SCDynamicStore`, does a `SCDynamicStoreCopyKeyList` over `State:/Network/Service/.*/Interface`, and then a `SCDynamicStoreCopyValue` per key. Line 253's `TRVPNServiceInfoByInterfaceName()` repeats the whole thing at line 131, on top of its own `SCPreferencesCreate` + `SCNetworkServiceCopyAll` + a `SCNetworkConnectionCopyExtendedStatus` per VPN service.

This is all synchronous IPC to configd on the main thread, and `PrefsController.updateBindInterfaceMenu` calls into it from `activeVPNBindInterfaceDidChange:` — i.e. on every SCDynamicStore change notification, which arrives in bursts during network churn (waking from sleep, VPN connect/disconnect, Wi-Fi handoff).

```suggestion
    NSDictionary<NSString*, NSString*>* userDefinedNames = TRUserDefinedNamesByInterfaceName();
    [names addEntriesFromDictionary:userDefinedNames];

    NSDictionary<NSString*, NSDictionary<NSString*, NSString*>*>* vpnServices = TRVPNServiceInfoByInterfaceName(userDefinedNames);
```

That needs `TRVPNServiceInfoByInterfaceName` to take the dictionary as a parameter and drop its own call on line 131 — it already assigns the result to a local there.

libtransmission/net.cc
```diff
@@ -106,6 +106,71 @@ int tr_make_listen_socket_ipv6only(tr_socket_t const sock)
...
+{
+    auto const name = tr_strv_strip(bind_interface);
+    return std::empty(name) || name == "default"sv;
+}
+
+std::string_view tr_net_effective_bind_interface(std::string_view const bind_interface)
+{
+    auto const name = tr_strv_strip(bind_interface);
+    return tr_net_interface_is_default(name) ? std::string_view{} : name;
+}
```
ckerr, line 126: **The predicates the whole fail-closed design rests on have no direct tests.**

`tr_net_interface_is_default`, `tr_net_effective_bind_interface` (line 115), `tr_netCurlInterfaceString` (line 164) and `tr_torrent::bind_interface_matches_session` (torrent.cc:1885) are what every guard in this PR reads — the UDP-announce skip, the DHT/LPD filters in session.cc, the uTP and holepunch checks in peer-mgr.cc, and `allows_incoming_peer` in the handshake. The tests added here cover settings load/save, RPC round-trips and resume persistence, but none of them pin the semantics those guards depend on.

Worth adding to tests/libtransmission/net-test.cc and torrent-test.cc:

- `""`, `"  "` and `"default"` all read as the default route; `" utun4 "` strips to `utun4`.
- `tr_netCurlInterfaceString` returns `nullopt` for all three, and `"if!utun4"` otherwise.
- `bind_interface_matches_session`: torrent empty inherits the session; torrent `"default"` does **not** match a session bound to `utun4`; torrent `"utun4"` does. That last group is what decides whether a torrent keeps UDP trackers, DHT, LPD, uTP, holepunch and incoming peers.

Without them, a later change to the `"default"` sentinel or to the empty-means-inherit rule flips torrents from fail-closed to leaking with the suite still green.

robd003:
@ckerr Thanks for the review! I'll try to get these resolved by the end of the weekend.

I really appreciate you starting the Retransmission fork and giving the project velocity again.

ckerr:
> @ckerr Thanks for the review! I'll try to get these resolved by the end of the weekend.

No hurry, get to it when you can. We're a long way from any release :slightly_smiling_face: 

Sounds good though!

reardonia:
The majority of the complexity here seems to be due to per-torrent bind_address, and having to route that throughout libtransmission.  Is that a real conceivable user scenario? 

Reducing this to allow for only (1) dynamic binding (rebind) during runtime and (2) device-name binding, would be significantly simpler (like, maybe 80% smaller?) and might yield the best usability vs complexity tradeoff

Add a tunnel-setup script invoker that triggers (just prior to?) bind and configures devices? Not strictly necessary but could help for many quirky configs.

robd003:
@ckerr could you re-run your `/review` skill? I believe I covered everything it caught last time.

libtransmission/peer-io.cc
ckerr: Fixed: `new_outgoing_utp()` and `create()` now take a `bind_interface` and thread it into the constructor, and `tr_peerMgrConnectHolepunch` passes `tor->effective_bind_interface()`. The post-hoc assignment in `new_outgoing()` folded into the same `create()` call.

libtransmission/announcer.cc
ckerr: Fixed: `tr_scrape_request` gained a `bind_interface`, `tr_tracker_http_scrape` sets `options.bind_interface`, `multiscrape` groups by it, and the `udp://` branch refuses mismatched requests with a response that fills every row's info_hash so each tier's `isScraping` clears. One follow-up on the grouping comparison is in the new review.

libtransmission/port-forwarding.cc
ckerr: Fixed: `natPulse` calls `natpmp_.reset()` when the interface is bound, `pulse()` is back to plain `is_enabled`, and UPnP is gated on `!tr_net_interface_is_blocked(...)`. The comment now states the lease-expiry trade-off.

libtransmission/tr-lpd.cc
ckerr: Addressed: `tr_net_interface_index()` now feeds `ipv6mr_interface` and the IPv4 `ip_mreqn`. The IPv6 half is right; the IPv4 group join has a follow-up in the new review, since `ip_mreqn` for `IP_ADD_MEMBERSHIP` is a Linux-only extension.

libtransmission/torrent.cc
ckerr: Fixed: the teardown is wrapped in `run_in_session_thread` with an id-based lookup, matching `tr_torrentVerify`. A follow-up on the field write itself is in the new review.

libtransmission/session.cc
ckerr: Fixed: `close_connections()` clears `incoming_handshakes`, `close_torrent_connections()` erases the ones matching this torrent's hash, and `close_torrent_connections_impl` destroys each peer_info's outgoing handshake. `~tr_handshake` doesn't fire the done callback, so none of these re-enter `on_handshake_done`.

libtransmission/announcer.cc
ckerr: Fixed: `response.did_connect = true`, with a comment naming why. The new scrape-side skip sets it too.

libtransmission/session.cc
ckerr: Fixed: the branch returns `*addr` only when `from_string` succeeds, so an empty or malformed value falls through to the interface check and then the source-address path. Comments now sit on the branches they describe.

libtransmission/net.cc
ckerr: Fixed: the branch sets `ENXIO` and logs at debug, and `tr_session::setSettings` emits one warning when the setting becomes non-default on an unsupported platform.

macosx/Utils.mm
ckerr: Fixed: the reserved values now live in libtransmission as `TrBindInterfaceDefault`/`TrBindInterfaceBlocked`, `tr_net_interface_is_blocked` refuses sockets with `ENETDOWN` + a trace, and the session logs the policy once. A follow-up on the curl path is in the new review.

docs/rpc-spec.md
ckerr: Fixed: five `bind_interface` rows added to the 4.2.0 table, and the reference-table descriptions now document `blocked` alongside `default`.

macosx/PrefsController.mm
ckerr: Fixed: the method is gone; the status field and info button are `IBOutlet`s in PrefsWindow.xib.

macosx/AddWindowController.mm
ckerr: Fixed: both copies deleted; AddWindow.xib and AddMagnetWindow.xib define the label and popup, and `awakeFromNib` calls `TRPopulateBindInterfacePopUp` directly.

macosx/Utils.mm
ckerr: Fixed: `TRVPNServiceInfoByInterfaceName` takes the dictionary as a parameter and both call sites pass it in.

libtransmission/net.cc
ckerr: Fixed: net-test.cc covers all four predicates table-driven (including whitespace and empty-inherits-session), plus a platform-guarded `tr_net_interface_index` test; torrent-test.cc adds `bindInterfaceMatchesSession`.

libtransmission/torrent.cc
```diff
@@ -1854,6 +1861,41 @@ void tr_torrent::set_bandwidth_group(std::string_view group_name)
...
+    set_dirty();
+    mark_changed();
+
+    // peer teardown touches libevent state owned by the session thread.
+    // Look the torrent up by id in case it was removed before this runs.
+    session->run_in_session_thread([session = this->session, tor_id = id()]() {
+        if (auto* const tor = session->torrents().get(tor_id); tor != nullptr) {
+            session->closeTorrentPeerConnections(tor);
+        }
+    });
```
ckerr: **The teardown moved to the session thread, but the write to `bind_interface_` didn't.**

`unique_lock()` on line 1868 guards this write, but the session-thread readers don't take that lock:

- `create_announce_request` (announcer.cc:911) copies `tor->bind_interface()`; it runs from the announcer upkeep timer, which never takes the session lock — announcer.cc has exactly one `unique_lock()` and it isn't on that path.
- `HandshakeMediator::torrent` (peer-mgr.cc:84-85) calls `bind_interface_matches_session()` from a peerIo read callback.

The macOS UI still calls `tr_torrentSetBindInterface` from the AppKit main thread, so reassigning this `std::string` — which can reallocate — races with both. `effective_bind_interface()` (line 1889) returns a `string_view` into that same buffer, so a reallocation mid-read leaves a dangling view rather than just a torn one.

```suggestion
    session->run_in_session_thread([session = this->session, tor_id = id(), name = std::string{ tr_strv_strip(bind_interface) }]() {
        auto* const tor = session->torrents().get(tor_id);
        if (tor == nullptr || tor->bind_interface_ == name) {
            return;
        }

        tor->bind_interface_ = name;
        tor->set_dirty();
        tor->mark_changed();

        // peer teardown touches libevent state owned by the session thread
        session->closeTorrentPeerConnections(tor);
    });
```

That makes the whole setter async, which the RPC and macOS callers already tolerate — `tr_sessionSetBindInterface` behaves the same way, and `SessionTest.propertiesApi` waits on it. `TorrentTest.bindInterfaceMatchesSession` reads back synchronously today, so it would need the same `waitFor` the session test uses.

libtransmission/tr-lpd.cc
```diff
@@ -316,13 +320,31 @@ class tr_lpd_impl final : public tr_lpd
...
+            // ip_mreqn selects by index when one is set, falling back to the address
+            struct ip_mreqn mcast_req = {};
+            mcast_req.imr_multiaddr = mcast_addr_.sin_addr;
+            mcast_req.imr_address = mediator_.bind_address(ip_protocol).addr.addr4;
+            mcast_req.imr_ifindex = static_cast<int>(if_index);
+            auto const* const mcast_if = reinterpret_cast<char const*>(&mcast_req);
+            auto const mcast_if_len = sizeof(mcast_req);
+#endif
 
             if (setsockopt(sock, IPPROTO_IP, IP_ADD_MEMBERSHIP, reinterpret_cast<char const*>(&mcast_req), sizeof(mcast_req)) ==
```
ckerr, line 347: **`ip_mreqn` for `IP_ADD_MEMBERSHIP` is a Linux extension — Darwin discards `imr_ifindex` here, silently.**

Only Linux's `ip_mc_join_group` accepts `struct ip_mreqn` for `IP_ADD_MEMBERSHIP`. Darwin (and the BSDs it inherited this code from) copy in with `sooptcopyin(sopt, &mreqs, sizeof(struct ip_mreq), sizeof(struct ip_mreq))`, which **truncates** when `optlen` is larger instead of returning EINVAL, then performs the ip_mreq-to-ip_mreq_source "switcharoo" that reinterprets bytes 4-7 as `imr_interface`.

Bytes 4-7 of `ip_mreqn` are `imr_address`, set on line 343 to `bind_address(ip_protocol).addr.addr4` — `0.0.0.0` whenever a bind interface is configured. So on macOS, the one platform where interface binding actually works, this join succeeds and uses the default interface: LPD keeps receiving announces off the tunnel and receives none on it. `IP_MULTICAST_IF` on line 366 *does* accept `ip_mreqn` on Darwin, so egress is genuinely fixed and only the join is wrong — which makes it easy to miss in testing, since outbound announces look correct.

Separately, `struct ip_mreqn` does not exist on NetBSD or OpenBSD, so the `#else` branch won't compile there. It isn't in this workflow's CI matrix (ubuntu/macos/windows only), so nothing would catch it.

No suggestion, because it needs a decision: either use `struct ip_mreq` for the join everywhere and rely on `IP_BOUND_IF` + `IP_MULTICAST_IF` for the interface, or resolve the bound interface's IPv4 address and put *that* in `imr_interface` (which is what the join actually reads on Darwin). The latter is the only way to pin the join to a specific interface portably.

libtransmission/net.cc
```diff
@@ -106,6 +107,102 @@ int tr_make_listen_socket_ipv6only(tr_socket_t const sock)
...
+
+std::optional<std::string> tr_netCurlInterfaceString(std::string_view const bind_interface)
+{
+    auto const name = tr_net_effective_bind_interface(bind_interface);
+    if (std::empty(name)) {
+        return {};
+    }
+
+    return fmt::format("if!{}", name);
+}
```
ckerr, line 236: **`blocked` escapes as an interface name here, which is the pattern `TrBindInterfaceBlocked` was introduced to remove.**

`tr_netSetSocketInterface` handles it explicitly on lines 154-159: `ENETDOWN`, a trace, refuse. This function has no such branch, so `tr_net_effective_bind_interface("blocked")` returns `"blocked"` and line 203 hands `if!blocked` to `CURLOPT_INTERFACE` (web.cc:342). The new test in net-test.cc documents the consequence directly — *"the blocked value names an interface that can't exist, so curl fails closed"*.

That's the same "fail by naming something that doesn't exist" trick that `tr-vpn-none` used, just relocated into libtransmission. Two costs: a host with an interface literally named `blocked` sends every tracker and webseed request through it, and in the normal case the user gets an opaque libcurl bind failure instead of the stated refusal that the socket path produces.

```suggestion
std::optional<std::string> tr_netCurlInterfaceString(std::string_view const bind_interface)
{
    auto const name = tr_net_effective_bind_interface(bind_interface);
    if (std::empty(name) || tr_net_interface_is_blocked(name)) {
        return {};
    }

    return fmt::format("if!{}", name);
}
```

That alone would make blocked requests go out *unbound*, so it needs a companion check where the fetch is issued: `tr_web::Impl` should refuse the task outright (and the net-test expectation flips to `EXPECT_FALSE`).

libtransmission/announcer.cc
```diff
@@ -1313,6 +1351,11 @@ void multiscrape(tr_announcer_impl* announcer, std::vector<tr_tier*> const& tier
                 continue;
             }
 
+            // a request goes out on one interface, so it can only carry torrents bound to that interface
+            if (req->bind_interface != tier->tor->bind_interface()) {
+                continue;
+            }
```
ckerr: **This compares raw values, so torrents on the same effective interface get split across requests.**

`tr_net_bind_interface_matches` (net.cc:127) does the normalization — empty inherits the session, `default` and `""` are the same route, whitespace stripped — but this is a plain `!=` on the raw per-torrent string. With the session bound to `utun4`, a torrent left at `""` and one explicitly set to `utun4` resolve to the same interface yet land in different requests; likewise `""` vs `default` on the default route, and `utun4` vs `" utun4 "`.

That isn't just wasted requests: `request_count` is capped at `MaxScrapesPerUpkeep` on line 1367, so the surplus requests displace other tiers from the round entirely. Once the cap is saturated, some torrents silently stop scraping.

```suggestion
            // a request goes out on one interface, so it can only carry torrents bound to that interface
            if (!tr_net_bind_interface_matches(tier->tor->bind_interface(), req->bind_interface)) {
                continue;
            }
```

`tr_net_bind_interface_matches(torrent, session)` treats an empty first argument as inheriting the second, which is exactly the relation wanted here as long as line 1371 keeps seeding `req->bind_interface` from the first torrent in the batch.

docs/Editing-Configuration-Files.md
```diff
@@ -118,6 +118,7 @@ Here is a sample of the three basic types: respectively Boolean, Number and Stri
 #### Peers
  * **bind_address_ipv4:** String (default = "") Where to listen for peer connections. When no valid IPv4 address is provided, Transmission will bind to "0.0.0.0".
  * **bind_address_ipv6:** String (default = "") Where to listen for peer connections. When no valid IPv6 address is provided, Transmission will try to bind to your default global IPv6 address. If that didn't work, then Transmission will bind to "::".
+ * **bind_interface:** String (default = "") Network interface name to route peer, tracker, DHT, LPD, and port-mapping traffic through on supported platforms (currently macOS), such as `utun4`. Empty or `default` means normal OS routing; `blocked` refuses all traffic. If a configured interface is unavailable, Transmission fails closed instead of falling back to the default route.
```
ckerr: **"fails closed instead of falling back to the default route" isn't true for DNS.**

`CURLOPT_INTERFACE` binds the connection socket only. libcurl's default threaded resolver calls `getaddrinfo()`, which the option does not affect — only c-ares builds can bind lookups. So with `bind_interface` set to a VPN tunnel, `tr_tracker_http_announce`, `tr_tracker_http_scrape` and webseed fetches resolve the tracker/webseed hostname over the default route *before* the bound connection is attempted.

For the VPN use case this setting exists to serve, that DNS query — carrying the tracker hostname — is precisely the metadata the user expects to be inside the tunnel. The UDP and peer paths are unaffected, since they work from addresses rather than names.

```suggestion
 * **bind_interface:** String (default = "") Network interface name to route peer, tracker, DHT, LPD, and port-mapping traffic through on supported platforms (currently macOS), such as `utun4`. Empty or `default` means normal OS routing; `blocked` refuses all traffic. If a configured interface is unavailable, Transmission fails closed instead of falling back to the default route. Note that DNS lookups for HTTP trackers and webseeds are made by the system resolver and are not bound to the interface.
```

Worth mirroring in the macOS Preferences tooltip, which currently says only that traffic is blocked when the VPN is unavailable.

ckerr (commented):
Second pass, on `1ba2c7095`. **All 15 findings from the first review are fixed** — I've replied on each thread with what changed and resolved them.

Several went further than the minimum: the `tr-vpn-none` sentinel became first-class `TrBindInterfaceDefault`/`TrBindInterfaceBlocked` values in libtransmission rather than a macOS-only hack; the scrape fix added a `bind_interface` to `tr_scrape_request` *and* interface-aware multiscrape batching; the handshake teardown covers both the incoming map and each peer_info's outgoing handshake; and the new net-test.cc / torrent-test.cc coverage is table-driven and hits the whitespace and empty-inherits-session cases. The macOS layout work moved wholesale into the XIBs.

5 new findings, all in the fixes themselves rather than the original feature:

- **`bind_interface_` is still written on the caller's thread** — the teardown moved to the session thread, but the field write didn't, and two session-thread readers take no lock.
- **`ip_mreqn` for `IP_ADD_MEMBERSHIP` is Linux-only** — Darwin truncates the option and reads `imr_address` as the interface, so the IPv4 LPD group join still uses the default interface. The `IP_MULTICAST_IF` half of that change is correct.
- **`blocked` still escapes into curl as `if!blocked`**, so HTTP is blocked only by no interface having that name.
- **multiscrape batches on raw strings**, splitting torrents that resolve to the same interface and burning slots against `MaxScrapesPerUpkeep`.
- **DNS isn't covered by the binding** — a docs claim, not a code bug, but material for the VPN use case.

Nothing here is a regression against `main`; they're all consequences of the new code.

libtransmission/torrent.cc
robd003: Done. The setter now runs entirely on the session thread and looks the torrent up by id; the torrent test waits on each value like the session test does.

libtransmission/tr-lpd.cc
robd003: Went with resolving the interface's IPv4 address: ip_mreq everywhere, filled from a new tr_net_interface_address() helper, and IPv4 LPD refuses to start when the bound interface has no IPv4 address.

libtransmission/net.cc
robd003: Done. tr_netCurlInterfaceString returns nothing for blocked, and tr_web::Impl::fetch completes the request as a failed connection before curl is involved. Two WebTest cases confirm the server never sees a request.

libtransmission/announcer.cc
robd003: Fixed by comparing effective interfaces on both sides via a two-argument tr_net_effective_bind_interface()

docs/Editing-Configuration-Files.md
robd003: Added the DNS caveat to the config doc, the rpc-spec session row, and both macOS tooltip strings in PrefsWindow.

reardonia:
for those of us in the gallery: what is the use-case of per-torrent network bindings?  just because?

robd003:
> for those of us in the gallery: what is the use-case of per-torrent network bindings? just because?

so this was a part of the original PR

For an actual use case I can see a scenario where you have an internal company VPN and you're exchanging large blobs of proprietary data. (In my scenario it's sharing machine learning models and getting them from my home machine to a cluster of servers)

However if I'm downloading some public torrent `debian-13.6.0-amd64-DVD-1.iso.torrent` then I'd rather just use my default connection and not clog up the limited bandwidth on my VPN.

robd003:
@ckerr Can you please kick off your AI review on this branch one more time?

ckerr:
@robd003 I can't find any other torrent clients that have per-torrent network configuration like this -- looks like it's a global setting everywhere. And the use case for per-torrent configuration seems pretty minor to me, to be honest. Could you run a simplification pass to make this a global setting instead of a per-torrent one?

@reardonia good catch.

robd003:
@ckerr No problem, just give me a little bit to rip out all of per-torrent settings.

libtransmission/session.cc
```diff
@@ -812,10 +863,16 @@ void tr_session::setSettings(tr_session::Settings&& settings_in, bool force)
...
         dht_.reset();
-    } else if (force || !dht_ || port_changed || addr_changed || new_settings.dht_enabled != old_settings.dht_enabled) {
+    } else if (
+        force || !dht_ || port_changed || addr_changed || interface_changed ||
+        new_settings.dht_enabled != old_settings.dht_enabled) {
         dht_ = tr_dht::create(dht_mediator_, advertisedPeerPort(), udp_core_->socket4(), udp_core_->socket6());
     }
 
+    if (interface_changed) {
+        tr_peerMgrCloseConnections(peer_mgr_.get());
```
ckerr, line 894: **[P1] Cancel active HTTP transfers when the binding changes.**

Start a webseed download on the default route, then select a VPN or `blocked`. This closes peers and handshakes, but leaves `tr_web::running_tasks_` untouched. Their curl handles retain the original sockets, so the download continues outside the selected interface. Cancel or restart outstanding HTTP work as part of the binding transition, including tracker requests.

libtransmission/web.cc
```diff
@@ -335,8 +352,18 @@ class tr_web::Impl
...
+        [[nodiscard]] auto curl_interface() const
         {
+            if (options_.bind_interface) {
+                return tr_netCurlInterfaceString(*options_.bind_interface);
+            }
+
+            if (auto const bind_interface = impl.mediator.bind_interface(); bind_interface) {
+                if (auto const curl_interface = tr_netCurlInterfaceString(*bind_interface); curl_interface) {
+                    return curl_interface;
+                }
```
ckerr, line 398: **[P1] Refuse queued requests when the session becomes blocked.**

A tracker request can pass `fetch()`'s guard, then wait in `queued_tasks_` while the session changes to `blocked`. When `curlThreadFunc()` calls `initEasy()`, this reads the new policy, converts `blocked` to `nullopt`, and falls through to ordinary address binding. With wildcard addresses, the request goes out unbound. Revalidate the policy at worker admission and cancel invalidated queued work; the enqueue-time check is insufficient.

libtransmission/session.cc
```diff
@@ -746,10 +765,28 @@ void tr_session::setSettings(tr_session::Settings&& settings_in, bool force)
...
+        }
+#endif
+    }
+
+    if (auto const& val = new_settings.bind_address_ipv4; force || interface_changed || val != old_settings.bind_address_ipv4) {
         ip_cache_->update_addr(TR_AF_INET);
     }
-    if (auto const& val = new_settings.bind_address_ipv6; force || val != old_settings.bind_address_ipv6) {
+    if (auto const& val = new_settings.bind_address_ipv6; force || interface_changed || val != old_settings.bind_address_ipv6) {
         ip_cache_->update_addr(TR_AF_INET6);
```
ckerr: **[P1] Invalidate cached public IPs when switching interfaces.**

Switch to a VPN with the ISP address already cached, then announce before the new IP probe completes. `update_addr()` leaves the old global address readable, and `announcer-http.cc` includes it in the announce's `ipv4`/`ipv6` parameters, exposing the ISP IP through the VPN. An outstanding pre-switch probe can also suppress this refresh and subsequently cache its old result. Clear the cached addresses and reject responses from earlier binding generations.

libtransmission/session.cc
```diff
@@ -309,6 +309,15 @@ std::optional<std::string> tr_session::WebMediator::bind_address_V6() const
     return std::nullopt;
 }
 
+std::optional<std::string> tr_session::WebMediator::bind_interface() const
+{
+    if (auto const& bind_interface = session_->bind_interface(); !tr_net_interface_is_default(bind_interface)) {
+        return bind_interface;
```
ckerr: **[P2] Synchronize reads of the binding string.**

`Task::curl_interface()` calls this on the curl worker while `setSettings()` swaps `settings_` on the session thread. Reading/copying the same `std::string` without synchronization is undefined behavior; changing the binding during HTTP startup can corrupt the interface value or crash. `tr_sessionGetBindInterface()` has the same issue for AppKit callers. Provide a synchronized snapshot, accounting for the worker already holding `tasks_mutex_`.

libtransmission/port-forwarding.cc
```diff
@@ -180,24 +189,33 @@ class tr_port_forwarding_impl final : public tr_port_forwarding
...
+        // A blocked binding has no interface to give it, and miniupnpc would fall
+        // back to the default route rather than fail, so UPnP stays off in that case.
+        auto const upnp_bindaddr = is_interface_bound ? std::string{ bind_interface } :
+                                                        mediator_.incoming_peer_address().display_name();
         upnp_state_ = tr_upnpPulse(
             upnp_,
             mediator_.advertised_peer_port(),
             mediator_.local_peer_port(),
-            is_enabled,
+            is_enabled && !tr_net_interface_is_blocked(bind_interface),
```
ckerr, line 234: **[P2] Blocking must discard UPnP state without sending an unmap.**

With an existing UPnP mapping, change the binding to `blocked`. Passing `false` here makes `tr_upnpPulse()` enter `WillUnmap` and send two `UPNP_DeletePortMapping()` requests to the cached gateway. Those HTTP requests are unbound; `upnp_bindaddr` only affects discovery. Thus selecting `blocked` itself sends traffic over the old route. Discard the mapping state without network cleanup, as done for NAT-PMP, and prevent reuse of cached gateways across binding changes.

libtransmission/session.cc
```diff
@@ -792,17 +843,17 @@ void tr_session::setSettings(tr_session::Settings&& settings_in, bool force)
...
     }
 
-    if (port_changed) {
+    if (port_changed || interface_changed) {
         port_forwarding_->local_port_changed();
     }
 
-    if (!udp_core_ || force || addr_changed || port_changed || utp_changed) {
+    if (!udp_core_ || force || addr_changed || interface_changed || port_changed || utp_changed) {
         udp_core_ = std::make_unique<tr_session::tr_udp_core>(*this, udpPort());
```
ckerr, line 872: **[P2] Retry socket creation when a literal interface becomes available.**

Start with a saved literal `utun4` binding while that VPN is disconnected. Interface lookup fails, leaving TCP listeners and UDP sockets invalid. When `utun4` later appears, the setting is unchanged, so these sockets are never recreated; the macOS monitor only reapplies bindings in Active VPN mode. UDP trackers, DHT/uTP, and incoming TCP remain unavailable until another settings change or restart. Retry failed bindings on interface availability changes, including literal mode.

ckerr (changes requested):
Third review pass, GPT 6 Astra xhigh with prompt:

> /code-review pull request #276. Add code comments / suggestions as draft inline review comments. Do not publish; I will do that after reviewing your findings. Reduce verbosity by 50%. Ignore pre-existing issues unless this PR makes them worse. Ignore the branch conflicts in net-test.cc for now.

robd003:
@ckerr Thanks for the review! Working on it now, will try to have everything resolved later today.

tearfur (changes requested):
I genuinely don't know much about how people uses this feature, so please let me know.

Is there a demand to have IPv4 and IPv6 use different interfaces?

tearfur:
Ack I didn't mean to press "Request changes". Can't dismiss my review for some reason too.

ckerr:
> Is there a demand to have IPv4 and IPv6 use different interfaces?

We don't need a per-family interface setting. One bind_interface for both is what qBittorrent and the other libtorrent-based apps do.

ckerr:
> [lolgear](https://github.com/lolgear) and others added 18 commits [1 hour ago](https://github.com/retransmission/retransmission/pull/276#commits-pushed-ff37388)

rebasing gone wrong?

robd003:
@ckerr yep, fixing later tonight. Sorry about the trouble.

ckerr:
No worries, we're a long way from release and nothing's time-sensitive rn :)

jameshealyio:
😍

I'm glad to see it's used somewhere other than my own damn Macbook! 😂

PR #277, opened 2026-08-28
perf: use strongly-typed TRGroup model in GroupsController

lolgear:
### Description
This PR refactors the legacy macOS group management logic by moving away from raw, untyped `NSMutableDictionary` stores toward a robust, strongly-typed `TRGroup` model. 

### Key Changes
* **Type Safety:** Added the `TRGroup` class implementing `NSSecureCoding`, replacing brittle string-keyed dictionaries with compiler-verified properties.
* **Backward Compatibility:** Implemented a seamless migration path in `init` that reads old legacy `GroupDicts` from `NSUserDefaults`, transforms them into `TRGroup` objects, and saves them in the modern format.
* **Performance Optimization:** Introduced `fIndexMap` (`NSMutableDictionary<NSNumber*, NSNumber*>`) to cache Group ID -> Array Index relationships, dropping the time complexity of row index resolution from $O(N)$ linear scans to $O(1)$ lookups.
* **Code Cleanup:** Offloaded icon cache management directly into a runtime-only property inside the model, removing boilerplate dictionary modifications.

### DIFF
--- a/macosx/GroupsController.h
+++ b/macosx/GroupsController.h
@@ -15,32 +15,32 @@
 - (NSInteger)rowValueForIndex:(NSInteger)index;
 - (NSInteger)indexForRow:(NSInteger)row;
 
-- (NSString*)nameForIndex:(NSInteger)index;
+- (nullable NSString*)nameForIndex:(NSInteger)index;
 - (void)setName:(NSString*)name forIndex:(NSInteger)index;
 
-- (NSImage*)imageForIndex:(NSInteger)index;
+- (nonnull NSImage*)imageForIndex:(NSInteger)index;
 
-- (NSColor*)colorForIndex:(NSInteger)index;
-- (void)setColor:(NSColor*)color forIndex:(NSInteger)index;
+- (nullable NSColor*)colorForIndex:(NSInteger)index;
+- (void)setColor:(nullable NSColor*)color forIndex:(NSInteger)index;
 
 - (BOOL)usesCustomDownloadLocationForIndex:(NSInteger)index;
 - (void)setUsesCustomDownloadLocation:(BOOL)useCustomLocation forIndex:(NSInteger)index;
 
-- (NSString*)customDownloadLocationForIndex:(NSInteger)index;
-- (void)setCustomDownloadLocation:(NSString*)location forIndex:(NSInteger)index;
+- (nullable NSString*)customDownloadLocationForIndex:(NSInteger)index;
+- (void)setCustomDownloadLocation:(nullable NSString*)location forIndex:(NSInteger)index;
 
 - (BOOL)usesAutoAssignRulesForIndex:(NSInteger)index;
 - (void)setUsesAutoAssignRules:(BOOL)useAutoAssignRules forIndex:(NSInteger)index;
 
-- (NSPredicate*)autoAssignRulesForIndex:(NSInteger)index;
-- (void)setAutoAssignRules:(NSPredicate*)predicate forIndex:(NSInteger)index;
+- (nullable NSPredicate*)autoAssignRulesForIndex:(NSInteger)index;
+- (void)setAutoAssignRules:(nullable NSPredicate*)predicate forIndex:(NSInteger)index;
 
 - (void)addNewGroup;
 - (void)removeGroupWithRowIndex:(NSInteger)row;
 
 - (void)moveGroupAtRow:(NSInteger)oldRow toRow:(NSInteger)newRow;
 
-- (NSMenu*)groupMenuWithTarget:(id)target action:(SEL)action isSmall:(BOOL)small;
+- (nonnull NSMenu*)groupMenuWithTarget:(id)target action:(SEL)action isSmall:(BOOL)small;
 
 - (NSInteger)groupIndexForTorrent:(Torrent*)torrent;
 @end
--- a/macosx/GroupsController.h
+++ b/macosx/GroupsController.h
@@ -6,9 +6,11 @@
 
 @class Torrent;
 
+NS_ASSUME_NONNULL_BEGIN
+
 @interface GroupsController : NSObject
 
-@property(nonatomic, class, readonly) GroupsController* groups;
+@property(nonatomic, class, readonly, nonnull) GroupsController* groups;
 
 @property(nonatomic, readonly) NSInteger numberOfGroups;
 
@@ -16,7 +18,7 @@
 - (NSInteger)indexForRow:(NSInteger)row;
 
 - (nullable NSString*)nameForIndex:(NSInteger)index;
-- (void)setName:(NSString*)name forIndex:(NSInteger)index;
+- (void)setName:(nullable NSString*)name forIndex:(NSInteger)index;
 
 - (nonnull NSImage*)imageForIndex:(NSInteger)index;
 
@@ -40,7 +42,9 @@
 
 - (void)moveGroupAtRow:(NSInteger)oldRow toRow:(NSInteger)newRow;
 
-- (nonnull NSMenu*)groupMenuWithTarget:(id)target action:(SEL)action isSmall:(BOOL)small;
+- (nonnull NSMenu*)groupMenuWithTarget:(nullable id)target action:(nullable SEL)action isSmall:(BOOL)small;
 
 - (NSInteger)groupIndexForTorrent:(Torrent*)torrent;
 @end
+
+NS_ASSUME_NONNULL_END
--- a/macosx/GroupsController.mm
+++ b/macosx/GroupsController.mm
@@ -4,7 +4,6 @@
 
 #import "GroupsController.h"
 #import "NSImageAdditions.h"
-#import "NSKeyedUnarchiverAdditions.h"
 #import "NSMutableArrayAdditions.h"
 
 static CGFloat const kIconWidth = 16.0;
@@ -86,10 +85,16 @@ static CGFloat const kIconWidthSmall = 12.0;
 /** Fast lookup map (Group ID -> Array Index) providing O(1) row queries. */
 @property(nonatomic, strong) NSMutableDictionary<NSNumber*, NSNumber*>* fIndexMap;
 
-- (NSImage*)imageForGroupNone;
-- (NSImage*)imageForGroup:(TRGroup*)group;
-- (BOOL)torrent:(Torrent*)torrent doesMatchRulesForGroupAtIndex:(NSInteger)index;
+- (nullable TRGroup*)groupForIndex:(NSInteger)index;
 - (void)rebuildIndexMap;
+
+- (void)saveGroups;
+- (void)saveGroupsAndNotify;
+
+- (BOOL)torrent:(Torrent*)torrent doesMatchRulesForGroup:(TRGroup*)group;
+
+- (nonnull NSImage*)imageForGroupNone;
+- (nonnull NSImage*)imageForGroup:(TRGroup*)group;
 
 @end
 
@@ -129,34 +134,39 @@ static CGFloat const kIconWidthSmall = 12.0;
                                                                                                    nil]
                                                                     fromData:data
                                                                        error:nil];
-            _fGroups = [[NSMutableArray alloc] init];
-            for (NSDictionary* dict in oldDicts) {
-                TRGroup* g = [[TRGroup alloc] initWithIndex:[dict[@"Index"] integerValue] name:dict[@"Name"] color:dict[@"Color"]];
-                g.usesCustomDownloadLocation = [dict[@"UsesCustomDownloadLocation"] boolValue];
-                g.customDownloadLocation = dict[@"CustomDownloadLocation"];
-                g.usesAutoGroupRules = [dict[@"UsesAutoGroupRules"] boolValue];
-                g.autoGroupRules = dict[@"AutoGroupRules"];
-                [_fGroups addObject:g];
+            if (oldDicts != nil) {
+                _fGroups = [[NSMutableArray alloc] init];
+                for (NSDictionary* dict in oldDicts) {
+                    TRGroup* g = [[TRGroup alloc] initWithIndex:[dict[@"Index"] integerValue] name:dict[@"Name"]
+                                                          color:dict[@"Color"]];
+                    g.usesCustomDownloadLocation = [dict[@"UsesCustomDownloadLocation"] boolValue];
+                    g.customDownloadLocation = dict[@"CustomDownloadLocation"];
+                    g.usesAutoGroupRules = [dict[@"UsesAutoGroupRules"] boolValue];
+                    g.autoGroupRules = dict[@"AutoGroupRules"];
+                    [_fGroups addObject:g];
+                }
+                [self saveGroups];
             }
-            [self saveGroups];
+
+            [NSUserDefaults.standardUserDefaults removeObjectForKey:@"GroupDicts"];
         }
 
         // 3. Fallback: Initialize with default color-coded groups if no saved data exists
         if (_fGroups == nil) {
             _fGroups = [[NSMutableArray alloc]
-                initWithObjects:[[TRGroup alloc] initWithIndex:0 name:NSLocalizedString(@"Red", @"Groups -> Name")
+                initWithObjects:[[TRGroup alloc] initWithIndex:0 name:NSLocalizedString(@"Red", "Groups -> Name")
                                                          color:NSColor.systemRedColor],
-                                [[TRGroup alloc] initWithIndex:1 name:NSLocalizedString(@"Orange", @"Groups -> Name")
+                                [[TRGroup alloc] initWithIndex:1 name:NSLocalizedString(@"Orange", "Groups -> Name")
                                                          color:NSColor.systemOrangeColor],
-                                [[TRGroup alloc] initWithIndex:2 name:NSLocalizedString(@"Yellow", @"Groups -> Name")
+                                [[TRGroup alloc] initWithIndex:2 name:NSLocalizedString(@"Yellow", "Groups -> Name")
                                                          color:NSColor.systemYellowColor],
-                                [[TRGroup alloc] initWithIndex:3 name:NSLocalizedString(@"Green", @"Groups -> Name")
+                                [[TRGroup alloc] initWithIndex:3 name:NSLocalizedString(@"Green", "Groups -> Name")
                                                          color:NSColor.systemGreenColor],
-                                [[TRGroup alloc] initWithIndex:4 name:NSLocalizedString(@"Blue", @"Groups -> Name")
+                                [[TRGroup alloc] initWithIndex:4 name:NSLocalizedString(@"Blue", "Groups -> Name")
                                                          color:NSColor.systemBlueColor],
-                                [[TRGroup alloc] initWithIndex:5 name:NSLocalizedString(@"Purple", @"Groups -> Name")
+                                [[TRGroup alloc] initWithIndex:5 name:NSLocalizedString(@"Purple", "Groups -> Name")
                                                          color:NSColor.systemPurpleColor],
-                                [[TRGroup alloc] initWithIndex:6 name:NSLocalizedString(@"Gray", @"Groups -> Name")
+                                [[TRGroup alloc] initWithIndex:6 name:NSLocalizedString(@"Gray", "Groups -> Name")
                                                          color:NSColor.systemGrayColor],
                                 nil];
             [self saveGroups];
@@ -188,6 +198,12 @@ static CGFloat const kIconWidthSmall = 12.0;
     }
 }
 
+- (void)saveGroupsAndNotify
+{
+    [NSNotificationCenter.defaultCenter postNotificationName:@"UpdateGroups" object:self];
+    [self saveGroups];
+}
+
 #pragma mark - Data Accessors
 
 - (NSInteger)numberOfGroups
@@ -209,38 +225,45 @@ static CGFloat const kIconWidthSmall = 12.0;
     return self.fGroups[row].groupIndex;
 }
 
-- (NSString*)nameForIndex:(NSInteger)index
+- (nullable TRGroup*)groupForIndex:(NSInteger)index
 {
     NSInteger orderIndex = [self rowValueForIndex:index];
-    return orderIndex != -1 ? self.fGroups[orderIndex].name : nil;
+    return orderIndex != -1 ? self.fGroups[orderIndex] : nil;
+}
+
+- (NSString*)nameForIndex:(NSInteger)index
+{
+    return [self groupForIndex:index].name;
 }
 
 - (void)setName:(NSString*)name forIndex:(NSInteger)index
 {
-    NSInteger orderIndex = [self rowValueForIndex:index];
-    if (orderIndex != -1) {
-        self.fGroups[orderIndex].name = name;
-        [self saveGroups];
-        [NSNotificationCenter.defaultCenter postNotificationName:@"UpdateGroups" object:self];
+    TRGroup* group = [self groupForIndex:index];
+
+    if (group == nil) {
+        return;
     }
+
+    group.name = name;
+    [self saveGroupsAndNotify];
 }
 
 - (NSColor*)colorForIndex:(NSInteger)index
 {
-    NSInteger orderIndex = [self rowValueForIndex:index];
-    return orderIndex != -1 ? self.fGroups[orderIndex].color : nil;
+    return [self groupForIndex:index].color;
 }
 
 - (void)setColor:(NSColor*)color forIndex:(NSInteger)index
 {
-    NSInteger orderIndex = [self rowValueForIndex:index];
-    if (orderIndex != -1) {
-        TRGroup* group = self.fGroups[orderIndex];
-        group.icon = nil; // Invalidate cached icon to force re-render with the new color
-        group.color = color;
-        [self saveGroups];
-        [NSNotificationCenter.defaultCenter postNotificationName:@"UpdateGroups" object:self];
+    TRGroup* group = [self groupForIndex:index];
+
+    if (group == nil) {
+        return;
     }
+
+    group.icon = nil; // Invalidate cached icon to force re-render with the new color
+    group.color = color;
+    [self saveGroupsAndNotify];
 }
 
 #pragma mark - Download Location Management
@@ -249,69 +272,77 @@ static CGFloat const kIconWidthSmall = 12.0;
 {
     if (![self customDownloadLocationForIndex:index])
         return NO;
-    NSInteger orderIndex = [self rowValueForIndex:index];
-    return orderIndex != -1 ? self.fGroups[orderIndex].usesCustomDownloadLocation : NO;
+
+    return [self groupForIndex:index].usesCustomDownloadLocation;
 }
 
 - (void)setUsesCustomDownloadLocation:(BOOL)useCustomLocation forIndex:(NSInteger)index
 {
-    NSInteger orderIndex = [self rowValueForIndex:index];
-    if (orderIndex != -1) {
-        self.fGroups[orderIndex].usesCustomDownloadLocation = useCustomLocation;
-        [self saveGroups];
+    TRGroup* group = [self groupForIndex:index];
+
+    if (group == nil) {
+        return;
     }
+
+    group.usesCustomDownloadLocation = useCustomLocation;
+    [self saveGroups];
 }
 
 - (NSString*)customDownloadLocationForIndex:(NSInteger)index
 {
-    NSInteger orderIndex = [self rowValueForIndex:index];
-    return orderIndex != -1 ? self.fGroups[orderIndex].customDownloadLocation : nil;
+    return [self groupForIndex:index].customDownloadLocation;
 }
 
 - (void)setCustomDownloadLocation:(NSString*)location forIndex:(NSInteger)index
 {
-    NSInteger orderIndex = [self rowValueForIndex:index];
-    if (orderIndex != -1) {
-        self.fGroups[orderIndex].customDownloadLocation = location;
-        [self saveGroups];
+    TRGroup* group = [self groupForIndex:index];
+
+    if (group == nil) {
+        return;
     }
+
+    group.customDownloadLocation = location;
+    [self saveGroups];
 }
 
 #pragma mark - Smart Rules (NSPredicate Matching)
 
 - (BOOL)usesAutoAssignRulesForIndex:(NSInteger)index
 {
-    NSInteger orderIndex = [self rowValueForIndex:index];
-    return orderIndex != -1 ? self.fGroups[orderIndex].usesAutoGroupRules : NO;
+    return [self groupForIndex:index].usesAutoGroupRules;
 }
 
 - (void)setUsesAutoAssignRules:(BOOL)useAutoAssignRules forIndex:(NSInteger)index
 {
-    NSInteger orderIndex = [self rowValueForIndex:index];
-    if (orderIndex != -1) {
-        self.fGroups[orderIndex].usesAutoGroupRules = useAutoAssignRules;
-        [self saveGroups];
+    TRGroup* group = [self groupForIndex:index];
+
+    if (group == nil) {
+        return;
     }
+
+    group.usesAutoGroupRules = useAutoAssignRules;
+    [self saveGroups];
 }
 
 - (NSPredicate*)autoAssignRulesForIndex:(NSInteger)index
 {
-    NSInteger orderIndex = [self rowValueForIndex:index];
-    return orderIndex != -1 ? self.fGroups[orderIndex].autoGroupRules : nil;
+    return [self groupForIndex:index].autoGroupRules;
 }
 
 - (void)setAutoAssignRules:(NSPredicate*)predicate forIndex:(NSInteger)index
 {
-    NSInteger orderIndex = [self rowValueForIndex:index];
-    if (orderIndex != -1) {
-        TRGroup* group = self.fGroups[orderIndex];
-        if (predicate) {
-            group.autoGroupRules = predicate;
-            [self saveGroups];
-        } else {
-            group.autoGroupRules = nil;
-            [self setUsesAutoAssignRules:NO forIndex:index];
-        }
+    TRGroup* group = [self groupForIndex:index];
+
+    if (group == nil) {
+        return;
+    }
+
+    if (predicate != nil) {
+        group.autoGroupRules = predicate;
+        [self saveGroups];
+    } else {
+        group.autoGroupRules = nil;
+        [self setUsesAutoAssignRules:NO forIndex:index];
     }
 }
 
@@ -332,37 +363,41 @@ static CGFloat const kIconWidthSmall = 12.0;
     [self.fGroups addObject:newGroup];
 
     [self rebuildIndexMap];
-    [NSNotificationCenter.defaultCenter postNotificationName:@"UpdateGroups" object:self];
-    [self saveGroups];
+
+    [self saveGroupsAndNotify];
 }
 
 - (void)removeGroupWithRowIndex:(NSInteger)row
 {
-    if (row < 0 || row >= self.fGroups.count)
+    if (row < 0 || row >= self.fGroups.count) {
         return;
+    }
 
     NSInteger index = self.fGroups[row].groupIndex;
     [self.fGroups removeObjectAtIndex:row];
     [self rebuildIndexMap];
 
     [NSNotificationCenter.defaultCenter postNotificationName:@"GroupValueRemoved" object:self
-
                                                     userInfo:@{ @"Index" : @(index) }];
+
     // Reset UI group filter if the currently filtered group was deleted
     if (index == [NSUserDefaults.standardUserDefaults integerForKey:@"FilterGroup"]) {
         [NSUserDefaults.standardUserDefaults setInteger:-2 forKey:@"FilterGroup"];
     }
-    [NSNotificationCenter.defaultCenter postNotificationName:@"UpdateGroups" object:self];
-    [self saveGroups];
+
+    [self saveGroupsAndNotify];
 }
 
 - (void)moveGroupAtRow:(NSInteger)oldRow toRow:(NSInteger)newRow
 {
-    if (oldRow < 0 || oldRow >= self.fGroups.count || newRow < 0 || newRow >= self.fGroups.count)
+    if (oldRow < 0 || oldRow >= self.fGroups.count || newRow < 0 || newRow >= self.fGroups.count) {
         return;
+    }
+
     [self.fGroups moveObjectAtIndex:oldRow toIndex:newRow];
-    [self rebuildIndexMap]; // Critical: Map must be updated as table row positions shifted[self saveGroups];
-    [NSNotificationCenter.defaultCenter postNotificationName:@"UpdateGroups" object:self];
+    [self rebuildIndexMap];
+
+    [self saveGroupsAndNotify];
 }
 
 #pragma mark - UI Menu & Torrent Matching
@@ -394,19 +429,20 @@ static CGFloat const kIconWidthSmall = 12.0;
 {
     // Evaluate smart-rules linearly to find the first matching category
     for (TRGroup* group in self.fGroups) {
-        if ([self torrent:torrent doesMatchRulesForGroupAtIndex:group.groupIndex]) {
+        if ([self torrent:torrent doesMatchRulesForGroup:group]) {
             return group.groupIndex;
         }
     }
     return -1;
 }
 
-- (BOOL)torrent:(Torrent*)torrent doesMatchRulesForGroupAtIndex:(NSInteger)index
+- (BOOL)torrent:(Torrent*)torrent doesMatchRulesForGroup:(TRGroup*)group
 {
-    if (![self usesAutoAssignRulesForIndex:index]) {
+    if (group.usesAutoGroupRules == NO) {
         return NO;
     }
-    NSPredicate* predicate = [self autoAssignRulesForIndex:index];
+
+    NSPredicate* predicate = group.autoGroupRules;
     [predicate allowEvaluation];
     BOOL eval = NO;
     @try {
@@ -420,7 +456,7 @@ static CGFloat const kIconWidthSmall = 12.0;
 
 #pragma mark - Icon Graphics Rendering
 
-- (NSImage*)imageForGroupNone
+- (nonnull NSImage*)imageForGroupNone
 {
     static NSImage* icon;
     if (icon)
@@ -437,19 +473,27 @@ static CGFloat const kIconWidthSmall = 12.0;
     return icon;
 }
 
-- (NSImage*)imageForGroup:(TRGroup*)group
+- (nonnull NSImage*)imageForGroup:(TRGroup*)group
 {
-    if (group.icon)
-        return group.icon; // Render the crisp round badge matching the group color and cache it inside the model instance
+    if (group.icon) {
+        return group.icon;
+    }
+
+    // Render the crisp round badge matching the group color and cache it inside the model instance
     NSImage* icon = [NSImage discIconWithColor:group.color insetFactor:0];
     group.icon = icon;
+
     return icon;
 }
 
-- (NSImage*)imageForIndex:(NSInteger)index
+- (nonnull NSImage*)imageForIndex:(NSInteger)index
 {
-    NSInteger orderIndex = [self rowValueForIndex:index];
-    TRGroup* group = orderIndex != -1 ? self.fGroups[orderIndex] : nil;
+    TRGroup* group = [self groupForIndex:index];
+
+    if (group == nil) {
+        return [self imageForGroupNone];
+    }
+
     return [self imageForGroup:group];
 }
 
--- a/macosx/GroupsController.mm
+++ b/macosx/GroupsController.mm
@@ -11,9 +11,85 @@ static CGFloat const kIconWidth = 16.0;
 static CGFloat const kBorderWidth = 1.25;
 static CGFloat const kIconWidthSmall = 12.0;
 
+@interface TRGroup : NSObject<NSSecureCoding>
+
+@property(nonatomic, assign) NSInteger groupIndex;
+@property(nonatomic, copy) NSString* name;
+@property(nonatomic, strong) NSColor* color;
+
+/** Runtime-only cache for the rendered group icon. Ignored during serialization. */
+@property(nonatomic, strong, nullable) NSImage* icon;
+
+@property(nonatomic, assign) BOOL usesCustomDownloadLocation;
+@property(nonatomic, copy, nullable) NSString* customDownloadLocation;
+
+@property(nonatomic, assign) BOOL usesAutoGroupRules;
+@property(nonatomic, strong, nullable) NSPredicate* autoGroupRules;
+
+- (instancetype)initWithIndex:(NSInteger)index name:(NSString*)name color:(NSColor*)color;
+
+@end
+
+@implementation TRGroup
+
++ (BOOL)supportsSecureCoding
+{
+    return YES;
+}
+
+- (instancetype)initWithIndex:(NSInteger)index name:(NSString*)name color:(NSColor*)color
+{
+    self = [super init];
+    if (self) {
+        _groupIndex = index;
+        _name = [name copy];
+        _color = color;
+    }
+    return self;
+}
+
+#pragma mark - NSSecureCoding
+
+- (void)encodeWithCoder:(NSCoder*)coder
+{
+    [coder encodeInteger:self.groupIndex forKey:@"Index"];
+    [coder encodeObject:self.name forKey:@"Name"];
+    [coder encodeObject:self.color forKey:@"Color"];
+    [coder encodeBool:self.usesCustomDownloadLocation forKey:@"UsesCustomDownloadLocation"];
+    [coder encodeObject:self.customDownloadLocation forKey:@"CustomDownloadLocation"];
+    [coder encodeBool:self.usesAutoGroupRules forKey:@"UsesAutoGroupRules"];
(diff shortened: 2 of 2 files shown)


macosx/GroupsController.mm
```diff
@@ -26,362 +102,355 @@ + (GroupsController*)groups
...
 }
 
 - (void)moveGroupAtRow:(NSInteger)oldRow toRow:(NSInteger)newRow
 {
+    if (oldRow < 0 || oldRow >= self.fGroups.count || newRow < 0 || newRow >= self.fGroups.count)
+        return;
     [self.fGroups moveObjectAtIndex:oldRow toIndex:newRow];
-
-    [self saveGroups];
+    [self rebuildIndexMap]; // Critical: Map must be updated as table row positions shifted[self saveGroups];
```
ckerr: **Blocking.** `[self saveGroups];` ended up inside the `//` comment, so it never runs — reordering groups is no longer persisted.

Drag a group to a new position in Prefs → Groups: the table updates, but `GroupObjects` still holds the old order and the change is gone on next launch.

```suggestion
    // Row positions shifted, so the id -> row map is stale.
    [self rebuildIndexMap];
    [self saveGroups];
```

macosx/GroupsController.mm
```diff
@@ -26,362 +102,355 @@ + (GroupsController*)groups
...
-    @try {
-        eval = [predicate evaluateWithObject:torrent];
-    } @catch (NSException* exception) {
-        NSLog(@"Error when evaluating predicate (%@) - %@", predicate, exception);
-    } @finally {
-        return eval;
-    }
+    NSInteger orderIndex = [self rowValueForIndex:index];
+    TRGroup* group = orderIndex != -1 ? self.fGroups[orderIndex] : nil;
+    return [self imageForGroup:group];
```
ckerr: **Blocking.** When `orderIndex` is -1 this now passes `nil` into `imageForGroup:`, dropping the old `imageForGroupNone` branch.

`FilterBarController.updateGroupsButton` (FilterBarController.mm:351) calls `imageForIndex:` with `-1` for the "None" group filter. `group.color` on nil gives nil, and `discIconWithColor:nil` draws nothing — the filter-bar button gets a blank swatch instead of the dashed-outline None icon. `group.icon = icon` is also a no-op on nil, so a fresh blank `NSImage` is allocated on every `UpdateGroups` notification.

```suggestion
    return orderIndex != -1 ? [self imageForGroup:self.fGroups[orderIndex]] : [self imageForGroupNone];
```

macosx/GroupsController.mm
```diff
@@ -26,362 +102,355 @@ + (GroupsController*)groups
...
+            NSArray* oldDicts = [NSKeyedUnarchiver unarchivedObjectOfClasses:[NSSet setWithObjects:NSArray.class,
+                                                                                                   NSDictionary.class,
+                                                                                                   NSNumber.class,
+                                                                                                   NSColor.class,
+                                                                                                   NSString.class,
+                                                                                                   NSPredicate.class,
+                                                                                                   nil]
+                                                                    fromData:data
+                                                                       error:nil];
+            _fGroups = [[NSMutableArray alloc] init];
```
ckerr: **Blocking — data loss.** `_fGroups` is assigned a non-nil empty array *before* `oldDicts` is known to be good, so the `if (_fGroups == nil)` defaults fallback below can no longer fire.

If `GroupDicts` exists but fails to decode (the `error:` is discarded), the loop body never runs, line 141 persists the empty array to `GroupObjects`, and the user is left with **zero groups permanently** — every later launch loads the empty `GroupObjects`. Before this PR the same corruption left `_fGroups` nil and restored the seven stock groups.

Build into a local and only assign on success:

```objc
NSMutableArray<TRGroup*>* migrated = [[NSMutableArray alloc] init];
for (NSDictionary* dict in oldDicts) { ... }
if (migrated.count > 0) {
    _fGroups = migrated;
    [self saveGroups];
}
```

macosx/GroupsController.mm
```diff
@@ -26,362 +102,355 @@ + (GroupsController*)groups
...
-                                                                                          nil]
-                                                           fromData:data
-                                                              error:nil];
-        } else if ((data = [NSUserDefaults.standardUserDefaults dataForKey:@"Groups"])) //handle old groups
-        {
-            _fGroups = [NSKeyedUnarchiver deprecatedUnarchiveObjectWithData:data];
-            [NSUserDefaults.standardUserDefaults removeObjectForKey:@"Groups"];
+
+        // 1. Try to load modern strongly-typed TRGroup objects
+        if ((data = [NSUserDefaults.standardUserDefaults dataForKey:@"GroupObjects"])) {
```
ckerr, line 129: **Blocking — data loss.** The `GroupObjects` load and the `GroupDicts` migration are chained with `else if`, so a failed modern decode never falls back to the legacy data the migration deliberately leaves behind.

After migration both keys exist (line 122 never removes `GroupDicts`). If `GroupObjects` later fails to decode, `_fGroups` is nil, the `else if` is skipped, and line 145 installs the stock defaults — then `saveGroups` overwrites `GroupObjects`. The still-valid `GroupDicts` copy is never consulted.

Also worth noting: the allowed-class set here has `NSMutableArray` but not `NSArray`, so anything that archives the root as an immutable array decodes to nil and hits exactly this path.

macosx/GroupsController.mm
```diff
@@ -26,362 +102,355 @@ + (GroupsController*)groups
...
+            _fGroups = [NSKeyedUnarchiver
+                unarchivedObjectOfClasses:[NSSet setWithObjects:NSMutableArray.class, TRGroup.class, NSColor.class, NSPredicate.class, nil]
+                                 fromData:data
+                                    error:nil];
+        }
+        // 2. Backward compatibility: Migrate old untyped dictionaries to TRGroup objects
+        else if ((data = [NSUserDefaults.standardUserDefaults dataForKey:@"GroupDicts"])) {
```
ckerr, line 136: **Blocking.** This drops the legacy `@"Groups"` branch (`deprecatedUnarchiveObjectWithData`) and its `removeObjectForKey:@"Groups"` cleanup, with no replacement.

A user upgrading from a Transmission old enough to only have `Groups` — never having run a version that wrote `GroupDicts` — now falls through to the defaults branch and loses their custom groups, colors, auto-assign rules and download locations, with `saveGroups` immediately making it permanent. The stale `Groups` key is also never cleaned up.

Was the removal intentional? If so it needs a note in the PR description, since it's a user-visible data migration being dropped.

Knock-on: `#import "NSKeyedUnarchiverAdditions.h"` (line 7) is now dead in this file — `deprecatedUnarchiveObjectWithData:` was its only use here.

macosx/GroupsController.mm
```diff
@@ -26,362 +102,355 @@ + (GroupsController*)groups
...
+                                                                                                   NSNumber.class,
+                                                                                                   NSColor.class,
+                                                                                                   NSString.class,
+                                                                                                   NSPredicate.class,
+                                                                                                   nil]
+                                                                    fromData:data
+                                                                       error:nil];
+            _fGroups = [[NSMutableArray alloc] init];
+            for (NSDictionary* dict in oldDicts) {
+                TRGroup* g = [[TRGroup alloc] initWithIndex:[dict[@"Index"] integerValue] name:dict[@"Name"] color:dict[@"Color"]];
```
ckerr: The two ways of constructing a `TRGroup` don't agree on defaults. `initWithCoder:` substitutes `@""` for a missing name and `systemGrayColor` for a missing color (lines 69-70); this migration path passes `dict[@"Name"]` / `dict[@"Color"]` straight through with no fallback.

A legacy entry missing `Color` gives `group.color == nil` → `colorForIndex:` returns nil and `imageForGroup:` renders a blank disc. Missing `Name` gives nil, which `groupMenuWithTarget:` hands to `[[NSMenuItem alloc] initWithTitle:nil ...]`.

Cleanest fix is to put the fallbacks in `initWithIndex:name:color:` so both paths inherit them.

ckerr (commented):
(claude opus 5, prompt: "/review extra --comment https://github.com/retransmission/retransmission/pull/277")

Review of the refactor — mostly mechanical translation to `TRGroup`, which reads well, but there are a few things to fix before this can land.

**Two that will bite users immediately:**
- `moveGroupAtRow:toRow:` — `[self saveGroups];` got absorbed into a trailing `//` comment, so group reordering is never persisted.
- `imageForIndex:` — now passes `nil` to `imageForGroup:` for index `-1` instead of returning the "None" icon; the filter-bar button renders a blank swatch.

**Three data-loss paths in the new two-key migration:** a failed `GroupDicts` decode now persists zero groups instead of restoring defaults; a failed `GroupObjects` decode never falls back to the `GroupDicts` copy the migration deliberately leaves in place; and the legacy `@"Groups"` path was removed outright.

**CI:** `./code_style.sh --force` will fail on the stray blank line at 349 and the over-long trailing comments at 385-386 and 443.

The rest are non-blocking notes, marked as such inline. Happy to talk through the `fIndexMap` design question (line 203) — for an array of seven, the invariant it introduces may cost more than the scan it removes, and it's what the line 364 bug fell out of.

Thanks for taking this on — the `TRGroup` model itself is a clear improvement over the string-keyed dictionaries.

macosx/GroupsController.mm
```diff
@@ -26,362 +102,355 @@ + (GroupsController*)groups
...
-    NSInteger index = [self.fGroups[row][@"Index"] integerValue];
+    if (row < 0 || row >= self.fGroups.count)
+        return;
+
+    NSInteger index = self.fGroups[row].groupIndex;
     [self.fGroups removeObjectAtIndex:row];
+    [self rebuildIndexMap];
 
     [NSNotificationCenter.defaultCenter postNotificationName:@"GroupValueRemoved" object:self
-                                                    userInfo:@{ @"Index" : @(index) }];
```
ckerr: **Blocking — CI.** Stray blank line inside a single message send, between `object:self` and `userInfo:`.

CI runs `./code_style.sh --force` (.github/workflows/actions.yml:120) and fails on the resulting diff. Same job will also catch the trailing comments at lines 385-386 and 443, which run past the 128-column `ColumnLimit`. Running the script locally fixes all of them.

macosx/GroupsController.mm
```diff
@@ -26,362 +102,355 @@ + (GroupsController*)groups
...
-        for (NSUInteger i = 0; i < self.fGroups.count; i++) {
-            if (index == [self.fGroups[i][@"Index"] integerValue]) {
-                return i;
-            }
-        }
-    }
-    return -1;
+    if (index == -1 || index == NSNotFound)
+        return -1;
+    NSNumber* row = self.fIndexMap[@(index)];
```
ckerr, line 233: Design question, non-blocking: is `fIndexMap` worth its invariant?

It adds a "rebuild after every mutation and reorder" requirement across four call sites to replace a linear scan over an array that holds seven groups by default. That invariant is the direct cause of the bug at line 364 — the `rebuildIndexMap` call got prioritized and `saveGroups` was lost to the comment explaining it.

Meanwhile `rowValueForIndex:` now boxes an `NSNumber` on every call, which at N≈7 is plausibly *slower* than the integer scan it replaced, and `rebuildIndexMap` is itself O(N) per mutation.

One behavioral change to be aware of either way: with duplicate `groupIndex` values (reachable from a legacy dict missing its `Index` key, which `integerValue` turns into 0) the map keeps the **last** match where the scan returned the **first**.

macosx/GroupsController.mm
```diff
@@ -26,362 +102,355 @@ + (GroupsController*)groups
...
 
 - (NSInteger)groupIndexForTorrent:(Torrent*)torrent
 {
-    for (NSDictionary* group in self.fGroups) {
-        NSInteger row = [group[@"Index"] integerValue];
-        if ([self torrent:torrent doesMatchRulesForGroupAtIndex:row]) {
-            return row;
+    // Evaluate smart-rules linearly to find the first matching category
+    for (TRGroup* group in self.fGroups) {
+        if ([self torrent:torrent doesMatchRulesForGroupAtIndex:group.groupIndex]) {
```
ckerr: Non-blocking, but this is the one path that actually iterates, and it didn't get the optimization.

`group` is right here, yet `group.groupIndex` is passed to `torrent:doesMatchRulesForGroupAtIndex:`, which round-trips back through `rowValueForIndex:` twice (once in `usesAutoAssignRulesForIndex:`, once in `autoAssignRulesForIndex:`) to recover the object we already hold — two `NSNumber` boxes and two dictionary lookups per group, per torrent.

A `- (BOOL)torrent:(Torrent*)torrent matchesGroup:(TRGroup*)group` reading `group.usesAutoGroupRules` / `group.autoGroupRules` directly removes all of it. (The round-trip predates this PR; flagging it because the PR is about lookup cost.)

macosx/GroupsController.mm
```diff
@@ -11,9 +11,85 @@
...
 static CGFloat const kIconWidthSmall = 12.0;
 
+@interface TRGroup : NSObject<NSSecureCoding>
+
+@property(nonatomic, assign) NSInteger groupIndex;
+@property(nonatomic, copy) NSString* name;
+@property(nonatomic, strong) NSColor* color;
+
+/** Runtime-only cache for the rendered group icon. Ignored during serialization. */
+@property(nonatomic, strong, nullable) NSImage* icon;
```
ckerr, line 20: Non-blocking: `nullable` appears here and on lines 24 and 27, but the file has no `NS_ASSUME_NONNULL_BEGIN`/`_END` pair and no other annotations.

`-Wnullability-completeness` fires file-wide once any nullability qualifier is present, so `name` and `color` (17-18), `initWithIndex:name:color:` (29), and the four extension declarations (84-91) will all warn "pointer is missing a nullability type specifier".

The annotations also miss the case that matters: `imageForGroup:` (line 90) is declared with an unannotated `TRGroup*` and is called with nil at line 453.

macosx/GroupsController.mm
```diff
@@ -11,9 +11,85 @@
...
 @interface GroupsController ()
 
-@property(nonatomic, readonly) NSMutableArray<NSMutableDictionary*>* fGroups;
+/** Backing array storing strongly-typed TRGroup model objects. */
+@property(nonatomic, strong) NSMutableArray<TRGroup*>* fGroups;
+
+/** Fast lookup map (Group ID -> Array Index) providing O(1) row queries. */
+@property(nonatomic, strong) NSMutableDictionary<NSNumber*, NSNumber*>* fIndexMap;
+
+- (NSImage*)imageForGroupNone;
```
ckerr: Non-blocking: these four forward declarations aren't needed — methods defined later in the same `@implementation` are callable without them. `saveGroups` in this same file proves it: used at line 141, defined at line 179, never declared.

They're duplication that has to be kept in sync with the definitions, and they replaced the `#pragma mark - Private` grouping that conveyed the same intent for free.

macosx/GroupsController.mm
```diff
@@ -26,362 +102,355 @@ + (GroupsController*)groups
...
         [NSUserDefaults.standardUserDefaults setInteger:-2 forKey:@"FilterGroup"];
     }
-
     [NSNotificationCenter.defaultCenter postNotificationName:@"UpdateGroups" object:self];
     [self saveGroups];
 }
 
 - (void)moveGroupAtRow:(NSInteger)oldRow toRow:(NSInteger)newRow
 {
+    if (oldRow < 0 || oldRow >= self.fGroups.count || newRow < 0 || newRow >= self.fGroups.count)
```
ckerr: Non-blocking: the new bounds guards here and in `removeGroupWithRowIndex:` silently `return`, while `indexForRow:` (line 209) still indexes `self.fGroups[row]` unguarded.

A bad row passed to `removeGroupWithRowIndex:` is now a no-op, so `GroupsPrefsController` removes the table row and the model doesn't — the desync surfaces later as an NSTableView row-count exception that's much harder to trace than the original throw. The same bad row through `indexForRow:` still throws.

Worth picking one: validate at the boundary consistently, or let all three throw.

macosx/GroupsController.mm
```diff
@@ -26,362 +102,355 @@ + (GroupsController*)groups
...
-
-            NSMutableDictionary* blue = [NSMutableDictionary
-                dictionaryWithObjectsAndKeys:NSColor.systemBlueColor, @"Color", NSLocalizedString(@"Blue", "Groups -> Name"), @"Name", @4, @"Index", nil];
-
-            NSMutableDictionary* purple = [NSMutableDictionary
-                dictionaryWithObjectsAndKeys:NSColor.systemPurpleColor, @"Color", NSLocalizedString(@"Purple", "Groups -> Name"), @"Name", @5, @"Index", nil];
+        // 3. Fallback: Initialize with default color-coded groups if no saved data exists
+        if (_fGroups == nil) {
+            _fGroups = [[NSMutableArray alloc]
+                initWithObjects:[[TRGroup alloc] initWithIndex:0 name:NSLocalizedString(@"Red", @"Groups -> Name")
```
ckerr: Nit: the comment argument changed from a C string to an NSString literal. Every one of the ~445 other `NSLocalizedString` call sites under `macosx/` uses the bare form — including line 386 in this same diff.

```suggestion
                initWithObjects:[[TRGroup alloc] initWithIndex:0 name:NSLocalizedString(@"Red", "Groups -> Name")
```

macosx/GroupsController.mm
```diff
@@ -26,362 +102,355 @@ + (GroupsController*)groups
...
-    if ((icon = dict[@"Icon"])) {
-        return icon;
-    }
-
-    icon = [NSImage discIconWithColor:dict[@"Color"] insetFactor:0];
-
-    dict[@"Icon"] = icon;
-
+    if (group.icon)
+        return group.icon; // Render the crisp round badge matching the group color and cache it inside the model instance
```
ckerr: Nit: the comment describes lines 444-445, but sits on the cached-value early return.

Line 386 has the same problem — it ends with `// Populate menu items using the clean object properties`, which describes the loop below, not the "None" item being added on that line.

```suggestion
    if (group.icon)
        return group.icon;
    // Cache the rendered badge on the model so it survives until the color changes.
```

ckerr (changes requested):

macosx/GroupsController.mm
```diff
@@ -11,9 +11,85 @@
...
+@property(nonatomic, copy, nullable) NSString* customDownloadLocation;
+
+@property(nonatomic, assign) BOOL usesAutoGroupRules;
+@property(nonatomic, strong, nullable) NSPredicate* autoGroupRules;
+
+- (instancetype)initWithIndex:(NSInteger)index name:(NSString*)name color:(NSColor*)color;
+
+@end
+
+@implementation TRGroup
```
Coeur, line 32: Could you move it to its own TRGroup.mm file?
I know, it will then require a TRGroup.h file... but long files are not easy to work with: we have to scroll so much to find the essential.

Note: the TRGroup.h files can be with visibility "project" instead of public.

macosx/GroupsController.mm
ckerr: re-review after 6ce6c9f: these forward declarations are still unnecessary

macosx/GroupsController.mm
ckerr: re-review after 6ce6c9f: the legacy "@Groups" branch was not restored

macosx/GroupsController.mm
ckerr: re-review after 6ce6c9f: `initWithIndex:name:color:` still applies no fallback, while `initWithCoder:` still has `?: @"" / ?: systemGrayColor`.

macosx/GroupsController.mm
ckerr: re-review after 6ce6c9f: the stray blank line is gone, but the two over-long lines remain (new 122 at 139 chars, new 421 at 161), so ./code_style.sh --force should still produce a diff.

macosx/GroupsController.mm
ckerr: re-review after 6ce6c9f: not fixed, and now broader: `TRGroup` still mixes nullable with unannotated name/color/init, and the header gained annotations with incomplete coverage (the groups class property, setName:'s param, groupIndexForTorrent:'s param). -Wnullability-completeness would now fire in every TU importing `GroupsController.h`, not just this file. `NS_ASSUME_NONNULL_BEGIN/_END` is still the fix.

macosx/GroupsController.mm
ckerr: re-review after 6ce6c9f: `indexForRow:` is still unguarded at new line 225 while its two siblings return silently.

macosx/GroupsController.mm
ckerr: re-review after 6ce6c9f: that one comment moved above the code it describes, but the identical problem at old 385-386 (now 420-421) is untouched; my comment named both.

ckerr (commented):
Updated code review comments inline, plus one new bug the latest commits introduced:

At new line ~152, `removeObjectForKey:@"GroupDicts"` sits outside the `if (oldDicts != nil)` block, so a failed decode now deletes the legacy data and installs defaults — the user's groups are gone with nothing left to recover from. Same hazard if `saveGroups` fails (it only logs). The removal needs to be conditional on a successful migration and a successful save.

ckerr (commented):
(claude opus 5, extra, prompt: "op has pushed up new commits, please update the analysis and add suggested replacements for all the findings as review comments to the PR")

Re-reviewed against the two new commits. Five of the earlier threads are properly fixed — the swallowed `saveGroups`, the nil-group `imageForIndex:`, the empty-array migration, the `groupIndexForTorrent:` round-trip, and the `NSLocalizedString` comment style. `groupForIndex:` and `saveGroupsAndNotify` tightened up the accessors nicely, and I'm happy to leave `fIndexMap` and the extension declarations as you've made them.

**One new problem arrived with the fixes:** `removeObjectForKey:@"GroupDicts"` runs outside the `if (oldDicts != nil)` guard, so a failed legacy decode now *deletes* the data before falling back to defaults. That's a worse outcome than the bug it was added alongside — details inline.

The rest are the items still outstanding from the first pass, each with a suggestion you can commit directly:

- `init` still discards the decode error, so first-run and corruption are indistinguishable (two threads, they work together)
- `initWithIndex:name:color:` still applies no defaults where `initWithCoder:` does
- `indexForRow:` still unguarded while its two siblings return silently
- the header's nullability coverage is incomplete, which now warns in every importer
- two trailing comments still exceed the column limit and will fail the style job
- whether the legacy `@"Groups"` upgrade path was meant to go away

Only the `GroupDicts` deletion and the style job block merging as far as I'm concerned; the rest is up to you.

macosx/GroupsController.mm
```diff
@@ -26,362 +107,394 @@ + (GroupsController*)groups
...
 
-            NSMutableDictionary* yellow = [NSMutableDictionary
-                dictionaryWithObjectsAndKeys:NSColor.systemYellowColor, @"Color", NSLocalizedString(@"Yellow", "Groups -> Name"), @"Name", @2, @"Index", nil];
+        // 1. Try to load modern strongly-typed TRGroup objects
+        if ((data = [NSUserDefaults.standardUserDefaults dataForKey:@"GroupObjects"])) {
+            _fGroups = [NSKeyedUnarchiver
+                unarchivedObjectOfClasses:[NSSet setWithObjects:NSMutableArray.class, TRGroup.class, NSColor.class, NSPredicate.class, nil]
+                                 fromData:data
+                                    error:nil];
+        }
```
ckerr, line 134: **Still open** — the decode error is discarded, so "key absent" (first run) and "key present but corrupt" are handled identically. They want opposite responses: the first should write defaults, the second must not.

Capturing the error also lets `mutableCopy` fix the class-set gap properly. `_fGroups` is declared `NSMutableArray*` but currently holds whatever the unarchiver returns — if that ever comes back immutable, the failure isn't here, it's later inside `addNewGroup` as an unrecognized-selector throw far from the cause.

`NSColor` and `NSPredicate` come out of the set because nothing decodes them at this level — `TRGroup initWithCoder:` passes its own class to `decodeObjectOfClass:` for each. As a bonus, the shorter line drops back under the 128-column limit.

```suggestion
        NSData* data;
        BOOL loadFailed = NO;

        // 1. Try to load modern strongly-typed TRGroup objects
        if ((data = [NSUserDefaults.standardUserDefaults dataForKey:@"GroupObjects"])) {
            NSError* error = nil;
            NSArray<TRGroup*>* const groups = [NSKeyedUnarchiver
                unarchivedObjectOfClasses:[NSSet setWithObjects:NSArray.class, TRGroup.class, nil]
                                 fromData:data
                                    error:&error];
            if (groups != nil) {
                _fGroups = [groups mutableCopy];
            } else {
                NSLog(@"unable to read saved groups: %@", error.localizedDescription);
                loadFailed = YES;
            }
        }
```

Pairs with the `loadFailed` check further down at line 172.

macosx/GroupsController.mm
```diff
@@ -26,362 +107,394 @@ + (GroupsController*)groups
...
+                                [[TRGroup alloc] initWithIndex:3 name:NSLocalizedString(@"Green", "Groups -> Name")
+                                                         color:NSColor.systemGreenColor],
+                                [[TRGroup alloc] initWithIndex:4 name:NSLocalizedString(@"Blue", "Groups -> Name")
+                                                         color:NSColor.systemBlueColor],
+                                [[TRGroup alloc] initWithIndex:5 name:NSLocalizedString(@"Purple", "Groups -> Name")
+                                                         color:NSColor.systemPurpleColor],
+                                [[TRGroup alloc] initWithIndex:6 name:NSLocalizedString(@"Gray", "Groups -> Name")
+                                                         color:NSColor.systemGrayColor],
+                                nil];
+            [self saveGroups];
```
ckerr, line 186: Second half of the line-117 change: with `loadFailed` available, don't overwrite groups you failed to read.

The app still runs with defaults for the session, but the original blob stays on disk — so a decode bug that ships in one release is recoverable in the next instead of having silently erased everyone's groups on first launch.

```suggestion
            // Groups that exist but didn't decode stay on disk: a later release may
            // read them, and overwriting turns one bad launch into permanent loss.
            if (!loadFailed) {
                [self saveGroups];
            }
```

macosx/GroupsController.mm
```diff
@@ -26,362 +107,394 @@ + (GroupsController*)groups
...
+                    g.usesAutoGroupRules = [dict[@"UsesAutoGroupRules"] boolValue];
+                    g.autoGroupRules = dict[@"AutoGroupRules"];
+                    [_fGroups addObject:g];
+                }
+                [self saveGroups];
+            }
 
-            NSMutableDictionary* green = [NSMutableDictionary
-                dictionaryWithObjectsAndKeys:NSColor.systemGreenColor, @"Color", NSLocalizedString(@"Green", "Groups -> Name"), @"Name", @3, @"Index", nil];
+            [NSUserDefaults.standardUserDefaults removeObjectForKey:@"GroupDicts"];
```
ckerr: **New in this push, and the most serious thing here.** `removeObjectForKey:@"GroupDicts"` sits outside the `if (oldDicts != nil)` block, so it runs even when the decode failed — the legacy data is deleted, `_fGroups` stays nil, defaults get installed, and the user's groups are gone with nothing left to recover from.

The `if (oldDicts != nil)` guard added in this push fixed the empty-array bug, but this line reintroduces the loss through the other door.

```suggestion
                [self saveGroups];

                // Drop the legacy copy only once its replacement is written.
                [NSUserDefaults.standardUserDefaults removeObjectForKey:@"GroupDicts"];
            }
```

Worth going one step further: `saveGroups` currently reports failure only through `NSLog`, so even inside the guard a serialization error deletes `GroupDicts` while leaving nothing valid in `GroupObjects`. Having `saveGroups` return `BOOL` and gating the removal on it would close that too.

macosx/GroupsController.mm
```diff
@@ -4,16 +4,97 @@
...
+- (instancetype)initWithIndex:(NSInteger)index name:(NSString*)name color:(NSColor*)color
+{
+    self = [super init];
+    if (self) {
+        _groupIndex = index;
+        _name = [name copy];
+        _color = color;
+    }
+    return self;
+}
```
ckerr: **Still open.** The two ways of building a `TRGroup` don't agree on defaults: `initWithCoder:` substitutes `@""` and `systemGrayColor` (lines 69-70), this one passes nil straight through.

The migration at line 140 feeds it `dict[@"Name"]` / `dict[@"Color"]` unchecked, so a legacy entry missing `Color` yields `group.color == nil` — `colorForIndex:` returns nil and `imageForGroup:` renders a blank disc. A missing `Name` reaches `[[NSMenuItem alloc] initWithTitle:nil ...]`.

Putting the fallbacks here means both paths inherit them:

```suggestion
- (instancetype)initWithIndex:(NSInteger)index name:(nullable NSString*)name color:(nullable NSColor*)color
{
    self = [super init];
    if (self) {
        _groupIndex = index;
        _name = [name copy] ?: @"";
        _color = color ?: NSColor.systemGrayColor;
    }
    return self;
}
```

With this in place, `initWithCoder:` could route through it and drop its own `?:` pair, so the defaults live in exactly one spot.

macosx/GroupsController.mm
```diff
@@ -26,362 +107,394 @@ + (GroupsController*)groups
...
+        return -1;
+    NSNumber* row = self.fIndexMap[@(index)];
+    return row ? [row integerValue] : -1;
 }
 
 - (NSInteger)indexForRow:(NSInteger)row
 {
-    return [self.fGroups[row][@"Index"] integerValue];
+    return self.fGroups[row].groupIndex;
 }
```
ckerr, line 240: **Still open.** `removeGroupWithRowIndex:` and `moveGroupAtRow:` both validate their row and return silently; this one still indexes straight into the array.

It's reached from `GroupsPrefsController.mm:62` on every cell draw and from a dozen `self.fTableView.selectedRow` call sites, where an empty selection is `-1`. Returning the same `-1` sentinel the other accessors use keeps the three consistent:

```suggestion
- (NSInteger)indexForRow:(NSInteger)row
{
    if (row < 0 || row >= self.fGroups.count) {
        return -1;
    }

    return self.fGroups[row].groupIndex;
}
```

Either direction is defensible — the one thing worth avoiding is the current split, where two of the three swallow bad input and the third throws.

macosx/GroupsController.mm
```diff
@@ -26,362 +107,394 @@ + (GroupsController*)groups
...
-
         [menu addItem:item];
-    };
-
-    addItemWithTitleTagIcon(NSLocalizedString(@"None", "Groups -> Menu"), -1, [self imageForGroupNone]);
-
-    for (NSMutableDictionary* dict in self.fGroups) {
-        addItemWithTitleTagIcon(dict[@"Name"], [dict[@"Index"] integerValue], [self imageForGroup:dict]);
+    }; // Add the default placeholder item "None" (-1)
+    addItemWithTitleTagIcon(NSLocalizedString(@"None", "Groups -> Menu"), -1, [self imageForGroupNone]); // Populate menu items using the clean object properties
```
ckerr: **Still open** — these two trailing comments are the last thing keeping `./code_style.sh --force` from being a no-op. Line 421 is 161 characters against a 128 limit, and both comments describe code other than the line they sit on: "Add the default placeholder item" is on the block's closing brace, and "Populate menu items" is on the line that adds the *None* item, not the loop below it.

```suggestion
    };

    // The "None" item is tagged -1, the same sentinel the accessors return for "no group".
    addItemWithTitleTagIcon(NSLocalizedString(@"None", "Groups -> Menu"), -1, [self imageForGroupNone]);

```

macosx/GroupsController.h
```diff
@@ -15,32 +15,32 @@
...
 - (void)addNewGroup;
 - (void)removeGroupWithRowIndex:(NSInteger)row;
 
 - (void)moveGroupAtRow:(NSInteger)oldRow toRow:(NSInteger)newRow;
 
-- (NSMenu*)groupMenuWithTarget:(id)target action:(SEL)action isSmall:(BOOL)small;
+- (nonnull NSMenu*)groupMenuWithTarget:(id)target action:(SEL)action isSmall:(BOOL)small;
 
 - (NSInteger)groupIndexForTorrent:(Torrent*)torrent;
 @end
```
ckerr, line 48: **Still open, and now wider than before.** Adding annotations to the header without full coverage means `-Wnullability-completeness` fires in every translation unit that imports `GroupsController.h`, not just in the .mm. Line 11's `groups` property, `setName:`'s parameter and `groupIndexForTorrent:`'s parameter are all still bare.

An assume-nonnull region is the usual fix — it covers the whole file, so the explicit `nonnull` spellings go away and only the exceptions stay marked:

```suggestion
- (NSInteger)rowValueForIndex:(NSInteger)index;
- (NSInteger)indexForRow:(NSInteger)row;

- (nullable NSString*)nameForIndex:(NSInteger)index;
- (void)setName:(NSString*)name forIndex:(NSInteger)index;

- (NSImage*)imageForIndex:(NSInteger)index;

- (nullable NSColor*)colorForIndex:(NSInteger)index;
- (void)setColor:(nullable NSColor*)color forIndex:(NSInteger)index;

- (BOOL)usesCustomDownloadLocationForIndex:(NSInteger)index;
- (void)setUsesCustomDownloadLocation:(BOOL)useCustomLocation forIndex:(NSInteger)index;

- (nullable NSString*)customDownloadLocationForIndex:(NSInteger)index;
- (void)setCustomDownloadLocation:(nullable NSString*)location forIndex:(NSInteger)index;

- (BOOL)usesAutoAssignRulesForIndex:(NSInteger)index;
- (void)setUsesAutoAssignRules:(BOOL)useAutoAssignRules forIndex:(NSInteger)index;

- (nullable NSPredicate*)autoAssignRulesForIndex:(NSInteger)index;
- (void)setAutoAssignRules:(nullable NSPredicate*)predicate forIndex:(NSInteger)index;

- (void)addNewGroup;
- (void)removeGroupWithRowIndex:(NSInteger)row;

- (void)moveGroupAtRow:(NSInteger)oldRow toRow:(NSInteger)newRow;

- (NSMenu*)groupMenuWithTarget:(id)target action:(SEL)action isSmall:(BOOL)small;

- (NSInteger)groupIndexForTorrent:(Torrent*)torrent;
@end

NS_ASSUME_NONNULL_END
```

This suggestion can't reach the matching opener, which needs to go above the `groups` property to cover it — please add `NS_ASSUME_NONNULL_BEGIN` on its own line after `@class Torrent;`.

Same treatment is worth applying in the .mm: `TRGroup` still mixes `nullable` on three properties with bare `name`, `color` and `initWithIndex:name:color:`.

macosx/GroupsController.mm
```diff
@@ -4,16 +4,97 @@
 
 #import "GroupsController.h"
 #import "NSImageAdditions.h"
-#import "NSKeyedUnarchiverAdditions.h"
 #import "NSMutableArrayAdditions.h"
```
ckerr, line 7: **Still open** — this is really a question rather than a defect.

Removing `NSKeyedUnarchiverAdditions.h` resolved the dead-import half of my earlier note, but it also confirms the legacy `@"Groups"` branch is gone for good. Anyone upgrading from a build old enough to have only that key — never having run a version that wrote `GroupDicts` — silently gets the seven stock groups, and `saveGroups` makes it permanent on first launch.

If dropping that path is deliberate, no code change needed, but it's a user-visible data migration being retired and belongs in the PR description (and probably wants a `Notes:` paragraph). If it isn't, the import needs to come back:

```suggestion
#import "GroupsController.h"
#import "NSImageAdditions.h"
#import "NSKeyedUnarchiverAdditions.h"
#import "NSMutableArrayAdditions.h"
```

along with a third branch after the `GroupDicts` one, mirroring what `main` does today:

```objc
else if ((data = [NSUserDefaults.standardUserDefaults dataForKey:@"Groups"])) {
    _fGroups = [[NSKeyedUnarchiver deprecatedUnarchiveObjectWithData:data] mutableCopy];
    [NSUserDefaults.standardUserDefaults removeObjectForKey:@"Groups"];
    [self saveGroups];
}
```

Note that one still hands back untyped dictionaries, so it needs the same dict-to-`TRGroup` conversion the `GroupDicts` branch does.

lolgear: Could you show anyone who can upgrade from that old version of ( Transmission )?

These groups were deprecated at this commit ( 111ce452d7940f6cbb648be6ca86fa6634473013 ).
Wed, 24 Dec, 2008, 20:41

I don't understand who may be affected?

macosx/GroupsController.mm
lolgear: I use `clang-format version 22.1.8`.
I don't understand why it doesn't apply rules properly.

macosx/GroupsController.mm
lolgear:
```objective-c
        else if ((data = [NSUserDefaults.standardUserDefaults dataForKey:@"Groups"])) //handle old groups
        {
            _fGroups = [NSKeyedUnarchiver deprecatedUnarchiveObjectWithData:data];
            [NSUserDefaults.standardUserDefaults removeObjectForKey:@"Groups"];
            [self saveGroups];
        }
```

It only checks presence of data at specific key.

lolgear:
@ckerr 

I don't understand this migration mechanism.
Could you tell what you're  trying to achieve in this migration?

Because

"Groups" as I said is very old key/data representation.
It **doesn't matter** if parser succeed here, it gives result and that is all. No error handling.
After that data at key "Groups" is **lost**.

Ok, what about "GroupDicts"?

Data doesn't exist at key "GroupDicts".
So, user doesn't have this data and here we have a fallback to default set/array of groups.

Data exists BUT parser/archiver cannot parse.
Actually, that means that user somehow corrupts this data.
No matter who did this. 
User ( human ) or his A.I. or system after upgrade or even Physics ( using heat or electric or magnetic forces ).

What can we do here?
We can use default set of groups and later save it in userDefaults at key "GroupObjects".

And what about this "corrupted" data at key "GroupDicts"?
It is lost.
Migration is over.

ckerr:
> I don't understand this migration mechanism. Could you tell what you're trying to achieve in this migration?

The goal is that every existing user's groups survive the upgrade, and that we can still fix it in a point release if some don't.

`Groups` is (probably) safe to delete on failure because almost nobody has it. Everybody has `GroupDicts`, and this migration runs once per profile — once the key is gone there's no second attempt.

The thing I'd change: `oldDicts == nil` doesn't mean the data is corrupt, only that we couldn't read it, and the likeliest cause is our own class list. The `GroupObjects` set a few lines above lists `NSMutableArray` without `NSArray`, so an immutable root decodes to nil on intact data. We can't tell that apart from real corruption here.

You don't need the error handling if you'd rather skip it. Moving the delete inside if `(oldDicts != nil)` is enough: don't delete what you couldn't read.

lolgear:
@ckerr @Coeur 

Well-well-well, I had a marvelous conversation with Gemini.
I asked it about NSKeyedArchiver and proper classes set and it insists on modern solutions as JSON representation in user defaults without binary data.

It makes sense and also it gives "discoverability" of data  ( plain json is better to read and edit by hand if needed ).
Another good point is zero cost in Objective-C++ to Swift migration.

### Hidden thing

One hidden thing is color serialization.
Actually, I cannot imagine how deep and complex this problem is.
Color could be dynamic ( light/dark theme ) and also they depend on devices or displays and the same color has various representations in different color spaces.

That means that "NSColor.systemRedColor" is not the same after first encoding/decoding and it differs from this getter "NSColor.systemRedColor".
At least it is better to use hex representation to encode colors from default set.

### And?..

I suggest to follow Gemini's path and to use plain text ( JSON representation ) instead of custom structures.
And don't use NSKeyedArchiver.

### Summary from Gemini

#### 1. The `NSKeyedArchiver` and `NSColor` Issue:

`NSColor` and `NSPredicate` are implemented as "Class Clusters." Under the hood, they hide various system subclasses (such as `NSCalibratedRGBColor` or `NSComparisonPredicate`). When `NSSecureCoding` is enabled, the unarchiver often throws exceptions or returns nil because the actual subclass type inside the binary payload fails the strict type validation against the base class.

#### 2. The System (Dynamic) Colors Trap:

Archiving a dynamic color like `NSColor.systemRedColor` via `NSKeyedArchiver` takes a static "snapshot" of its physical RGB components at that exact moment. The object completely loses its dynamic system capability. As a result, it stops responding to Light/Dark mode transitions or accessibility contrast adjustments, creating a visual mismatch against fresh system colors down the line.

#### The Clean UserDefaults Solution:

The most robust approach is to completely discard `NSKeyedArchiver` and binary blobs (`NSData`). Instead, objects are serialized into a clean, text-based `NSDictionary` and persisted in `UserDefaults` as a standard property-list array:

* System Colors are saved as string identifiers (e.g., "system:systemRedColor"), ensuring they are re-instantiated fresh and fully dynamic upon loading.
* Custom Colors (e.g., picked from an NSColorWell) are cleanly stored as basic hex strings ("#RRGGBBAA").
* Predicates are effortlessly preserved via their native string format using [predicate predicateFormat].

ckerr:
(claude opus 5, extra, prompt: "OP has pushed new commits and also had a new proposal at https://github.com/retransmission/retransmission/pull/277#issuecomment-5468939907 . Please update the review after analyzing the new comment and new code.")

The class-cluster claim doesn't hold. `decodeObjectOfClass:` accepts subclasses, so `NSColor.class` in the allowed set covers `NSCalibratedRGBColor`. `main` has archived `NSColor` and `NSPredicate` with `requiringSecureCoding:YES` since `GroupDicts` was added — if strict validation rejected cluster subclasses, group colors would have been broken for every user for years.

The dynamic-color point is real and worth its own issue. `main` archives the same `NSColor.systemRedColor` objects the same way, so if they flatten, that's a bug we already ship rather than something this PR introduces. Your fix for it works just as well inside `encodeWithCoder:` — encode the catalog name as a string instead of the color object. It doesn't require dropping `NSKeyedArchiver`.

Predicates via `predicateFormat` I'd rule out. `predicateWithFormat:` parses text that we then evaluate against Torrent objects, and `FUNCTION()` can invoke arbitrary selectors. The `allowEvaluation` call in `torrent:doesMatchRulesForGroup:` is there because unarchived predicates are untrusted by default. Making that text hand-editable in defaults is the part I don't want.

If you want the plist format, please do it as a separate PR. This one already carries a format change and a migration, and that migration has produced two data-loss bugs in three rounds.

Coeur:
There is always a lot to read with AI generated discussions.
I don't use AI, I do manual reviews. Kindly split the @interface and @implementation that you added into their own files, and I'll review afterward.

macosx/GroupsController.mm
```diff
@@ -26,362 +116,403 @@ + (GroupsController*)groups
...
-
-            NSMutableDictionary* yellow = [NSMutableDictionary
-                dictionaryWithObjectsAndKeys:NSColor.systemYellowColor, @"Color", NSLocalizedString(@"Yellow", "Groups -> Name"), @"Name", @2, @"Index", nil];
+        // 1. Try to load modern strongly-typed TRGroup objects
+        if ((data = [NSUserDefaults.standardUserDefaults dataForKey:@"GroupObjects"])) {
+            _fGroups = [NSKeyedUnarchiver
+                unarchivedObjectOfClasses:[NSSet setWithObjects:NSMutableArray.class, TRGroup.class, NSColor.class, NSPredicate.class, nil]
+                                 fromData:data
+                                    error:nil];
+        }
```
ckerr: The error observation you added to the `GroupDicts` branch didn't reach this one. `GroupObjects` is the format every user ends up on, so it needs it more: `error:nil` here makes "no key yet" and "key present but unreadable" indistinguishable, and they want opposite responses.

Two other things on this line. The set lists `NSMutableArray` without `NSArray`, and secure coding matches by `isKindOfClass:` — an immutable root fails against a mutable-only entry. The legacy set can lean on `main` having shipped it; this format has no such history. `[groups mutableCopy]` also makes the declared type of `_fGroups` true, rather than trusting the archive to hand back something mutable.

`NSColor` and `NSPredicate` come out because nothing decodes them at this level — `TRGroup initWithCoder:` passes its own class per field. Line 131 drops under 128 columns as a result, which is the last thing failing the style job.

```suggestion
        NSData* data;
        BOOL loadFailed = NO;

        // 1. Try to load modern strongly-typed TRGroup objects
        if ((data = [NSUserDefaults.standardUserDefaults dataForKey:@"GroupObjects"])) {
            NSError* error = nil;
            NSArray<TRGroup*>* const groups = [NSKeyedUnarchiver
                unarchivedObjectOfClasses:[NSSet setWithObjects:NSArray.class, TRGroup.class, nil]
                                 fromData:data
                                    error:&error];
            if (groups != nil) {
                _fGroups = [groups mutableCopy];
            } else {
                NSLog(@"unable to read saved groups: %@", error.localizedDescription);
                loadFailed = YES;
            }
        }
```

macosx/GroupsController.mm
```diff
@@ -26,362 +116,403 @@ + (GroupsController*)groups
...
+                                [[TRGroup alloc] initWithIndex:3 name:NSLocalizedString(@"Green", "Groups -> Name")
+                                                         color:NSColor.systemGreenColor],
+                                [[TRGroup alloc] initWithIndex:4 name:NSLocalizedString(@"Blue", "Groups -> Name")
+                                                         color:NSColor.systemBlueColor],
+                                [[TRGroup alloc] initWithIndex:5 name:NSLocalizedString(@"Purple", "Groups -> Name")
+                                                         color:NSColor.systemPurpleColor],
+                                [[TRGroup alloc] initWithIndex:6 name:NSLocalizedString(@"Gray", "Groups -> Name")
+                                                         color:NSColor.systemGrayColor],
+                                nil];
+            [self saveGroups];
```
ckerr: Pairs with the change at line 126. You applied this rule to `GroupDicts` by moving the delete inside the guard; it applies here too.

Writing defaults over groups you failed to read makes a decoding bug permanent on first launch. Skipping the write costs nothing: the app still runs with defaults, and the next successful edit saves.

```suggestion
            // Groups that exist but didn't decode stay on disk: a later release may
            // read them, and overwriting turns one bad launch into permanent loss.
            if (!loadFailed) {
                [self saveGroups];
            }
```

ckerr (commented):
Re-reviewed at 49ecb7c. Four threads from the last round are fixed: the `GroupDicts` delete now sits inside the guard, `null_resettable` setters give both construction paths the same defaults (cleaner than what I suggested — routing `initWithCoder:` through the designated init puts them in one place), the trailing comments are off the code lines, and the header has an assume-nonnull region.

Four items remain, all in the inline comments. The `GroupObjects` decode is the one that matters: the error handling you added to the legacy branch needs to be there too, and that path has no shipped precedent to fall back on.

Minor, no comment threads for these:
- `if (error != nil)` keys the log off the out-param; Cocoa only guarantees it on a failure return. The decision correctly still uses `oldDicts != nil`, so this is cosmetic.
- `nonnull` on `groups`, `imageForIndex:` and `groupMenuWithTarget:`'s return is now redundant inside `NS_ASSUME_NONNULL_BEGIN`.

On the JSON proposal, replied separately.

macosx/GroupsController.mm
```diff
@@ -26,362 +116,403 @@ + (GroupsController*)groups
...
+        return -1;
+    NSNumber* row = self.fIndexMap[@(index)];
+    return row ? [row integerValue] : -1;
 }
 
 - (NSInteger)indexForRow:(NSInteger)row
 {
-    return [self.fGroups[row][@"Index"] integerValue];
+    return self.fGroups[row].groupIndex;
 }
```
ckerr: Still unguarded, while `removeGroupWithRowIndex:` and `moveGroupAtRow:` both validate and return.

`GroupsPrefsController` calls this on every cell draw and from a dozen `selectedRow` sites, where an empty selection is -1. Returning the -1 sentinel the other accessors use makes the three consistent.

```suggestion
- (NSInteger)indexForRow:(NSInteger)row
{
    if (row < 0 || row >= self.fGroups.count) {
        return -1;
    }

    return self.fGroups[row].groupIndex;
}
```

macosx/GroupsController.mm
```diff
@@ -4,16 +4,106 @@
 
 #import "GroupsController.h"
 #import "NSImageAdditions.h"
-#import "NSKeyedUnarchiverAdditions.h"
 #import "NSMutableArrayAdditions.h"
```
ckerr: Still open from the first round, and it needs an answer rather than a code change.

Dropping `NSKeyedUnarchiverAdditions.h` retires the `@"Groups"` upgrade path. Anyone coming from a build that only ever wrote that key gets the seven stock groups, and `saveGroups` makes it permanent on first launch.

If that's deliberate, say so in the PR description and add a `Notes:` paragraph, since it's user-visible. If not, the import comes back along with a third branch mirroring `main`, plus the same dict-to-`TRGroup` conversion the `GroupDicts` branch does.

ckerr:
The class-cluster claim doesn't hold. `decodeObjectOfClass:` accepts subclasses, so `NSColor.class` in the allowed set covers `NSCalibratedRGBColor`. `main` has archived `NSColor` and `NSPredicate` with `requiringSecureCoding:YES` since `GroupDicts` was added — if strict validation rejected cluster subclasses, group colors would have been broken for every user for years.

The dynamic-color point is real and worth its own issue. `main` archives the same `NSColor.systemRedColor` objects the same way, so if they flatten, that's a bug we already ship rather than something this PR introduces. Your fix for it works just as well inside `encodeWithCoder:` — encode the catalog name as a string instead of the color object. It doesn't require dropping NSKeyedArchiver.

Predicates via `predicateFormat` I'd rule out. `predicateWithFormat:` parses text that we then evaluate against Torrent objects, and `FUNCTION()` can invoke arbitrary selectors. The `allowEvaluation` call in `torrent:doesMatchRulesForGroup:` is there because unarchived predicates are untrusted by default. Making that text hand-editable in defaults is the part I don't want.

There's also no Swift in the tree today, so that benefit is hypothetical.

If you want the plist format, please do it as a separate PR. This one already carries a format change and a migration, and that migration has produced two data-loss bugs in three rounds.

PR #278, opened 2026-08-28
perf: move TorrentCell priority updates out of drawRect

lolgear:
### Description
Optimized `TorrentCell` by moving the torrent priority UI logic out of the `drawRect:` method.

### Changes
- Removed priority icon state management and image loading from `drawRect:`.
- Added `setObjectValue:` override to capture data updates.
- Introduced `setTorrentPriority:` helper method to update the priority icon and stack view visibility only when the data changes.

### Impact
Improves UI rendering performance and avoids unnecessary stack view layout passes during cell redraws.

### DIFF
--- a/macosx/TorrentCell.mm
+++ b/macosx/TorrentCell.mm
@@ -266,20 +266,26 @@ static CGFloat const kRevealButtonTrailingOffset = -8.0; // inverted for constra
         // draw progress bar
         NSRect barRect = self.fTorrentProgressBarView.frame;
         [ProgressBarView.sharedInstance drawBarInRect:barRect forTableView:self.fTorrentTableView withTorrent:torrent];
-
-        // set priority icon
-        if (torrent.priority != TR_PRI_NORMAL) {
-            NSImage* priorityImage = [NSImage imageNamed:(torrent.priority == TR_PRI_HIGH ? @"PriorityHighTemplate" : @"PriorityLowTemplate")];
-
-            self.fTorrentPriorityView.image = priorityImage;
-
-            [self.fStackView setVisibilityPriority:NSStackViewVisibilityPriorityMustHold forView:self.fTorrentPriorityView];
-        } else {
-            [self.fStackView setVisibilityPriority:NSStackViewVisibilityPriorityNotVisible forView:self.fTorrentPriorityView];
-        }
     }
 
     [super drawRect:dirtyRect];
+}
+
+- (void)setObjectValue:(id)objectValue {
+    [super setObjectValue:objectValue];
+    [self setTorrentPriority:[(Torrent*)objectValue priority]];
+}
+
+- (void)setTorrentPriority:(tr_priority_t)priority {    
+    if (priority != TR_PRI_NORMAL) {
+        NSImage* priorityImage = [NSImage imageNamed:(priority == TR_PRI_HIGH ? @"PriorityHighTemplate" : @"PriorityLowTemplate")];
+
+        self.fTorrentPriorityView.image = priorityImage;
+
+        [self.fStackView setVisibilityPriority:NSStackViewVisibilityPriorityMustHold forView:self.fTorrentPriorityView];
+    } else {
+        [self.fStackView setVisibilityPriority:NSStackViewVisibilityPriorityNotVisible forView:self.fTorrentPriorityView];
+    }
 }
 
 // otherwise progress bar is inverted
--- a/macosx/TorrentCell.mm
+++ b/macosx/TorrentCell.mm
@@ -279,15 +279,20 @@ static CGFloat const kRevealButtonTrailingOffset = -8.0; // inverted for constra
 
 - (void)setTorrentPriority:(tr_priority_t)priority
 {
-    if (priority != TR_PRI_NORMAL) {
-        NSImage* priorityImage = [NSImage imageNamed:(priority == TR_PRI_HIGH ? @"PriorityHighTemplate" : @"PriorityLowTemplate")];
+    BOOL const hasPriority = priority != TR_PRI_NORMAL;
 
-        self.fTorrentPriorityView.image = priorityImage;
+    NSImage* const image = hasPriority ?
+        [NSImage imageNamed:(priority == TR_PRI_HIGH ? @"PriorityHighTemplate" : @"PriorityLowTemplate")] :
+        nil;
 
-        [self.fStackView setVisibilityPriority:NSStackViewVisibilityPriorityMustHold forView:self.fTorrentPriorityView];
-    } else {
-        [self.fStackView setVisibilityPriority:NSStackViewVisibilityPriorityNotVisible forView:self.fTorrentPriorityView];
+    NSStackViewVisibilityPriority const visibility = hasPriority ? NSStackViewVisibilityPriorityMustHold : NSStackViewVisibilityPriorityNotVisible;
+
+    if (self.fTorrentPriorityView.image == image && [self.fStackView visibilityPriorityForView:self.fTorrentPriorityView] == visibility) {
+        return;
     }
+
+    self.fTorrentPriorityView.image = image;
+    [self.fStackView setVisibilityPriority:visibility forView:self.fTorrentPriorityView];
 }
 
 // otherwise progress bar is inverted
--- a/macosx/TorrentCell.mm
+++ b/macosx/TorrentCell.mm
@@ -271,12 +271,14 @@ static CGFloat const kRevealButtonTrailingOffset = -8.0; // inverted for constra
     [super drawRect:dirtyRect];
 }
 
-- (void)setObjectValue:(id)objectValue {
+- (void)setObjectValue:(id)objectValue
+{
     [super setObjectValue:objectValue];
     [self setTorrentPriority:[(Torrent*)objectValue priority]];
 }
 
-- (void)setTorrentPriority:(tr_priority_t)priority {    
+- (void)setTorrentPriority:(tr_priority_t)priority
+{
     if (priority != TR_PRI_NORMAL) {
         NSImage* priorityImage = [NSImage imageNamed:(priority == TR_PRI_HIGH ? @"PriorityHighTemplate" : @"PriorityLowTemplate")];
 


macosx/TorrentCell.mm
```diff
@@ -266,20 +266,28 @@ - (void)drawRect:(NSRect)dirtyRect
         // draw progress bar
         NSRect barRect = self.fTorrentProgressBarView.frame;
         [ProgressBarView.sharedInstance drawBarInRect:barRect forTableView:self.fTorrentTableView withTorrent:torrent];
+    }
+
+    [super drawRect:dirtyRect];
```
ckerr: **Refresh trigger narrowed.** Worth being explicit about what this move gives up: the priority icon used to be re-derived on *any* repaint of the cell, and now it is re-derived only when `objectValue` is assigned.

Two existing paths repaint without reloading:

- `TorrentTableView.refreshTorrentTable` (`TorrentTableView.mm:92-95`) sets `self.needsDisplay = YES` in response to the `RefreshTorrentTable` notification posted from `GlobalOptionsPopoverViewController.mm:49`.
- `flagsChanged:` (`TorrentTableView.mm:490`) calls `[self display]`.

Neither currently coincides with a priority change, so I don't think this regresses anything a user can see today — the 1 Hz `updateUI` → `reloadVisibleRows` covers it (and is skipped entirely while `NSApp.hidden`, `Controller.mm:2068`). But the invariant "a repaint leaves the cell correct" is gone and nothing re-establishes it. Fine to accept deliberately; just worth a line in the commit message so the next person knows it was a decision rather than an oversight.

macosx/TorrentCell.mm
```diff
@@ -266,20 +266,28 @@ - (void)drawRect:(NSRect)dirtyRect
...
+}
+
+- (void)setObjectValue:(id)objectValue
```
ckerr: **Two configuration paths for one cell.** Everything else this cell displays is set explicitly in `TorrentTableView.outlineView:viewForTableColumn:item:` (`TorrentTableView.mm:216-309`): icon, title, status string, progress string, group indicator, button actions. With this change, priority becomes the one field configured as a *side effect* of `torrentCell.objectValue = torrent`.

That splits the answer to "where does this cell get filled in?" across two mechanisms, and the implicit one is invisible to anyone reading `viewForTableColumn:`.

Two coherent options:

1. Set it alongside the other fields — `[torrentCell setTorrentPriority:torrent.priority];` right after line 287 — and skip the `setObjectValue:` override entirely. Smallest diff, keeps one configuration site, and still gets the layout work out of `drawRect:`.
2. Go the other way and move the whole per-row configuration behind `objectValue` on the cell.

Either is defensible; the current half-and-half is the fragile one. I'd take (1) for this PR.

macosx/TorrentCell.mm
```diff
@@ -266,20 +266,28 @@ - (void)drawRect:(NSRect)dirtyRect
...
+{
+    [super setObjectValue:objectValue];
+    [self setTorrentPriority:[(Torrent*)objectValue priority]];
```
ckerr: **Unchecked cast on an AppKit-owned setter.** `objectValue` is `id` on `NSTableCellView`, and this override sits on the public setter, so it accepts whatever anyone assigns — not just what `TorrentTableView.mm:287` passes today.

- `nil` is safe only by accident: nil-messaging returns `0`, which happens to equal `TR_PRI_NORMAL`.
- Anything that is not a `Torrent` is `unrecognized selector` at row-configuration time. That is not hypothetical: if the data source ever implements `outlineView:objectValueForTableColumn:byItem:` — which `TorrentTableView.mm:382` already assumes exists — AppKit auto-assigns its return value to the cell view's `objectValue`.

The code this replaced at least read `torrent.priority` behind `if (self.fTorrentTableView)` inside `drawRect:`. This path has no guard at all.

```suggestion
    Torrent* const torrent = [objectValue isKindOfClass:Torrent.class] ? objectValue : nil;
    [self setTorrentPriority:torrent ? torrent.priority : TR_PRI_NORMAL];
```

macosx/TorrentCell.mm
```diff
@@ -266,20 +266,28 @@ - (void)drawRect:(NSRect)dirtyRect
...
+}
 
-        // set priority icon
-        if (torrent.priority != TR_PRI_NORMAL) {
-            NSImage* priorityImage = [NSImage imageNamed:(torrent.priority == TR_PRI_HIGH ? @"PriorityHighTemplate" : @"PriorityLowTemplate")];
+- (void)setTorrentPriority:(tr_priority_t)priority
```
ckerr: **Naming and declaration.** Two things here:

*It is declared nowhere.* `TorrentCell.h:31-33` declares exactly the methods intended for subclass override (`configureViews`, `configurePriorities`, `setupConstraints`), and `SmallTorrentCell` overrides all three. This new method follows neither that pattern nor the alternative one for a file-private helper (a class extension at the top of the `.mm`). Pick one — a class extension is right here, since nothing outside this file calls it.

*The `set` prefix is misleading.* It advertises a `torrentPriority` property that does not exist, makes the cell KVC-settable for a key it does not store, and reads like an accessor when the method actually pushes state into subviews. `updatePriorityView:` says what it does.

```objc
@interface TorrentCell ()
- (void)updatePriorityView:(tr_priority_t)priority;
@end
```

macosx/TorrentCell.mm
```diff
@@ -266,20 +266,28 @@ - (void)drawRect:(NSRect)dirtyRect
...
+{
+    if (priority != TR_PRI_NORMAL) {
```
ckerr: **This is where the "only when the data changes" claim needs to become true.** As written there is no early-out, and `setObjectValue:` is called for every visible row on every `reloadVisibleRows` — once per second. With ~30 visible rows that is 30 `imageNamed:` lookups and 30 `setVisibilityPriority:forView:` writes per second, i.e. the same work the `drawRect:` version was doing.

Caching the last applied value on the cell makes the description accurate and the win real:

```objc
// in the class extension
@property(nonatomic) tr_priority_t fPriority;

- (void)updatePriorityView:(tr_priority_t)priority
{
    if (priority == self.fPriority) {
        return;
    }
    self.fPriority = priority;
    ...
}
```

Note this needs an initial value that cannot collide with a real priority, or an explicit reset, so a freshly created cell always runs the update once — `TR_PRI_NORMAL` is `0`, which is also the default for a new ivar, and a normal-priority cell starting hidden makes that work out. Worth a brief comment stating that rather than leaving it to be rediscovered.

macosx/TorrentCell.mm
```diff
@@ -266,20 +266,28 @@ - (void)drawRect:(NSRect)dirtyRect
...
+        self.fTorrentPriorityView.image = priorityImage;
 
-            [self.fStackView setVisibilityPriority:NSStackViewVisibilityPriorityMustHold forView:self.fTorrentPriorityView];
-        } else {
-            [self.fStackView setVisibilityPriority:NSStackViewVisibilityPriorityNotVisible forView:self.fTorrentPriorityView];
-        }
+        [self.fStackView setVisibilityPriority:NSStackViewVisibilityPriorityMustHold forView:self.fTorrentPriorityView];
+    } else {
+        [self.fStackView setVisibilityPriority:NSStackViewVisibilityPriorityNotVisible forView:self.fTorrentPriorityView];
     }
```
ckerr: **Stale image on recycled cells.** The hide branch never clears the image, so a cell that previously showed a `TR_PRI_HIGH` torrent keeps `PriorityHighTemplate` loaded when it is recycled (`makeViewWithIdentifier:`, `TorrentTableView.mm:238`) for a normal-priority one. Only `NSStackViewVisibilityPriorityNotVisible` keeps it off screen.

This predates your PR, but the original implementation did clear it — `1a86c65b4` had `self.fTorrentPriorityView.image = nil;` in the else branch, and it was lost in a later refactor. Since you are relocating this exact block, it is the natural place to put it back.

```suggestion
    } else {
        self.fTorrentPriorityView.image = nil;

        [self.fStackView setVisibilityPriority:NSStackViewVisibilityPriorityNotVisible forView:self.fTorrentPriorityView];
    }
```

lolgear:
@ckerr
Could you check this PR?

macosx/TorrentCell.mm
ckerr: (Fable 5.1) The description says this updates "only when the data changes", but nothing checks: `objectValue` is reassigned on every `reloadVisibleRows` (1 s timer, plus each hover), so the stack view is poked once per second per visible row, the same cadence as before. If the guard is wanted, read the current state back from the views rather than caching the priority in an ivar: one zero-initialised to `TR_PRI_NORMAL` would skip the first hide on a fresh cell, whose stack view defaults to `MustHold`. The same edit collapses the two branches and clears the stale image in the normal case.

```suggestion
    BOOL const hasPriority = priority != TR_PRI_NORMAL;
    NSImage* const image = hasPriority ?
        [NSImage imageNamed:(priority == TR_PRI_HIGH ? @"PriorityHighTemplate" : @"PriorityLowTemplate")] :
        nil;
    NSStackViewVisibilityPriority const visibility = hasPriority ? NSStackViewVisibilityPriorityMustHold :
                                                                    NSStackViewVisibilityPriorityNotVisible;

    if (self.fTorrentPriorityView.image == image && [self.fStackView visibilityPriorityForView:self.fTorrentPriorityView] == visibility) {
        return;
    }

    self.fTorrentPriorityView.image = image;
    [self.fStackView setVisibilityPriority:visibility forView:self.fTorrentPriorityView];
```

macosx/TorrentCell.mm
ckerr, line 280: (Fable 5.1) `setTorrentPriority:` reads as the setter of a `torrentPriority` property that doesn't exist, and beside `Torrent.priority` (a real setter writing through to libtransmission) it suggests the cell changes the torrent.

Consider a verb like the sibling cells' `updateDisplay` (FilePriorityCellView, FileCheckCellView), e.g. `updatePriorityView`, reading `((Torrent*)self.objectValue).priority` itself as `drawRect:` does, and declaring it in an `@interface TorrentCell ()` extension as SmallTorrentCell.mm does.

macosx/TorrentCell.mm
ckerr, line 274: (Fable 5.1) Optional: every other per-torrent field is set in `TorrentTableView -outlineView:viewForTableColumn:item:`. Calling the helper from there, next to `fGroupIndicatorView.image` where a typed `Torrent*` is already in scope, keeps the binding in one place and drops this override and its `id` cast. Either way, the comment above `torrentCell.objectValue = torrent;` (TorrentTableView.mm:286, "set this so that we can draw bar in torrentCell drawRect") now covers only half of what that assignment does.

ckerr (commented):
Reviewed by Fable 5.1, extra, with the prompt:

> /code-review PR 278. Focus on issues/suggestions that are new to this PR. Ignore pre-existing issues unless this PR makes them worse. Add as draft review comments/suggestions. Prepend comments you add with "(Fable 5.1)". Cut verbosity by 50%.

macosx/TorrentCell.mm
lolgear: Yes, it is intentional.
This PR doesn't "update" priority or do additional refactoring like "extract priority to model".

Only one thing:

Do not update torrent priority ( and stackView properties ) in Drawing cycle. ( drawRect ).

macosx/TorrentCell.mm
lolgear: Yes, it is intentional.

Nothing to do with other properties/data.
This PR doesn't refactor whole TorrentCell and encapsulate everything in setters/update methods.

PR #282, opened 2026-08-31
macOS. Migrate deprecated allowedFileTypes to UTType.

lolgear:
## Description

This PR migrates the macOS-specific file type handling from the deprecated `allowedFileTypes` and `typeOfFile` APIs to the modern `UTType` and `allowedContentTypes` infrastructure introduced in macOS 11+.

## Changes
- Replaced `allowedFileTypes` array strings with `UTType.torrent` and `UTTypeText` in panels.
- Refactored file type detection in `Controller`, `DragOverlayWindow`, and `ExpandedPathToIconTransformer` to leverage the new helper properties.
- Extended `UTType` and `NSURL` via categories in `DefaultAppHelper` to provide robust, uniform torrent file classification.

## Motivation and Context
The older `allowedFileTypes` methods are deprecated by Apple. Moving to `UTType` ensures forward compatibility with newer macOS versions, eliminates hardcoded uniform type identifier strings across multiple files, and improves type-safety.

## How Has This Been Tested?
- Verified that opening and saving `.torrent` files works via dialog sheets.
- Verified that drag-and-drop actions properly identify torrent files and directories.

### DIFF
--- a/macosx/Controller.mm
+++ b/macosx/Controller.mm
@@ -958,9 +958,11 @@ static auto getSettingsFromNSUserDefaults(NSUserDefaults* defaults)
     didReceiveResponse:(nonnull NSURLResponse*)response
      completionHandler:(nonnull void (^)(NSURLSessionResponseDisposition))completionHandler
 {
-    UTType* contentType = [UTType typeWithMIMEType:response.MIMEType];
+    NSString* mimeType = response.MIMEType;
+    UTType* contentType = mimeType.length > 0 ? [UTType typeWithMIMEType:mimeType] : nil;
     NSString* suggestedName = response.suggestedFilename;
-    UTType* fileType = [UTType typeWithFilenameExtension:suggestedName.pathExtension];
+    NSString* suggestedExtension = suggestedName.pathExtension;
+    UTType* fileType = suggestedExtension.length > 0 ? [UTType typeWithFilenameExtension:suggestedExtension] : nil;
     UTType* torrentType = UTType.torrent;
 
     BOOL isTorrent = [contentType conformsToType:torrentType] || [fileType conformsToType:torrentType];
@@ -986,7 +988,11 @@ static auto getSettingsFromNSUserDefaults(NSUserDefaults* defaults)
         if (mainWindow) {
             [alert beginSheetModalForWindow:mainWindow completionHandler:nil];
         } else {
-            [NSApp activateIgnoringOtherApps:YES];
+            if (@available(macOS 14.0, *)) {
+                [NSApp activate];
+            } else {
+                [NSApp activateIgnoringOtherApps:YES];
+            }
             [alert runModal];
         }
     });
--- a/macosx/Controller.mm
+++ b/macosx/Controller.mm
@@ -62,6 +62,7 @@
 #import "ExpandedPathToIconTransformer.h"
 #import "PowerManager.h"
 #import "Utils.h"
+#import "DefaultAppHelper.h"
 
 typedef NSString* ToolbarItemIdentifier NS_TYPED_EXTENSIBLE_ENUM;
 
@@ -1284,7 +1285,7 @@ static auto getSettingsFromNSUserDefaults(NSUserDefaults* defaults)
     panel.canChooseFiles = YES;
     panel.canChooseDirectories = NO;
 
-    panel.allowedFileTypes = @[ @"org.bittorrent.torrent", @"torrent" ];
+    panel.allowedContentTypes = @[ UTType.torrent ];
 
     [panel beginSheetModalForWindow:self.fWindow completionHandler:^(NSInteger result) {
         if (result == NSModalResponseOK) {
@@ -1875,7 +1876,7 @@ static auto getSettingsFromNSUserDefaults(NSUserDefaults* defaults)
 
     if (!torrent.magnet && [NSFileManager.defaultManager fileExistsAtPath:torrent.torrentLocation]) {
         NSSavePanel* panel = [NSSavePanel savePanel];
-        panel.allowedFileTypes = @[ @"org.bittorrent.torrent", @"torrent" ];
+        panel.allowedContentTypes = @[ UTType.torrent ];
         panel.extensionHidden = NO;
 
         panel.nameFieldStringValue = torrent.name;
@@ -3131,8 +3132,8 @@ static auto getSettingsFromNSUserDefaults(NSUserDefaults* defaults)
 
         NSString* fullFile = [path stringByAppendingPathComponent:file];
 
-        if (!([[NSWorkspace.sharedWorkspace typeOfFile:fullFile error:NULL] isEqualToString:@"org.bittorrent.torrent"] ||
-              [fullFile.pathExtension caseInsensitiveCompare:@"torrent"] == NSOrderedSame)) {
+        auto fileURL = [NSURL fileURLWithPath:fullFile];
+        if (!(fileURL.isTorrentFile)) {
             continue;
         }
 
@@ -3369,8 +3370,7 @@ static auto getSettingsFromNSUserDefaults(NSUserDefaults* defaults)
         NSArray<NSURL*>* files = [pasteboard readObjectsForClasses:@[ NSURL.class ]
                                                            options:@{ NSPasteboardURLReadingFileURLsOnlyKey : @YES }];
         for (NSURL* fileToParse in files) {
-            if ([[NSWorkspace.sharedWorkspace typeOfFile:fileToParse.path error:NULL] isEqualToString:@"org.bittorrent.torrent"] ||
-                [fileToParse.pathExtension caseInsensitiveCompare:@"torrent"] == NSOrderedSame) {
+            if (fileToParse.isTorrentFile) {
                 torrent = YES;
                 auto metainfo = tr_torrent_metainfo{};
                 if (metainfo.parse_torrent_file(fileToParse.path.UTF8String)) {
@@ -3431,8 +3431,7 @@ static auto getSettingsFromNSUserDefaults(NSUserDefaults* defaults)
                                                            options:@{ NSPasteboardURLReadingFileURLsOnlyKey : @YES }];
         NSMutableArray<NSString*>* filesToOpen = [NSMutableArray arrayWithCapacity:files.count];
         for (NSURL* file in files) {
-            if ([[NSWorkspace.sharedWorkspace typeOfFile:file.path error:NULL] isEqualToString:@"org.bittorrent.torrent"] ||
-                [file.pathExtension caseInsensitiveCompare:@"torrent"] == NSOrderedSame) {
+            if (file.isTorrentFile) {
                 torrent = YES;
                 auto metainfo = tr_torrent_metainfo{};
                 if (metainfo.parse_torrent_file(file.path.UTF8String)) {
--- a/macosx/Controller.mm
+++ b/macosx/Controller.mm
@@ -958,11 +958,18 @@ static auto getSettingsFromNSUserDefaults(NSUserDefaults* defaults)
     didReceiveResponse:(nonnull NSURLResponse*)response
      completionHandler:(nonnull void (^)(NSURLSessionResponseDisposition))completionHandler
 {
+    UTType* contentType = [UTType typeWithMIMEType:response.MIMEType];
     NSString* suggestedName = response.suggestedFilename;
-    if ([suggestedName.pathExtension caseInsensitiveCompare:@"torrent"] == NSOrderedSame) {
+    UTType* fileType = [UTType typeWithFilenameExtension:suggestedName.pathExtension];
+    UTType* torrentType = UTType.torrent;
+
+    BOOL isTorrent = [contentType conformsToType:torrentType] || [fileType conformsToType:torrentType];
+
+    if (isTorrent) {
         completionHandler(NSURLSessionResponseBecomeDownload);
         return;
     }
+
     completionHandler(NSURLSessionResponseCancel);
 
     NSString* message = [NSString
@@ -974,7 +981,14 @@ static auto getSettingsFromNSUserDefaults(NSUserDefaults* defaults)
         [alert addButtonWithTitle:NSLocalizedString(@"OK", "Download not a torrent -> button")];
         alert.messageText = NSLocalizedString(@"Torrent download failed", "Download not a torrent -> title");
         alert.informativeText = message;
-        [alert runModal];
+
+        NSWindow* mainWindow = NSApp.mainWindow ?: NSApp.keyWindow;
+        if (mainWindow) {
+            [alert beginSheetModalForWindow:mainWindow completionHandler:nil];
+        } else {
+            [NSApp activateIgnoringOtherApps:YES];
+            [alert runModal];
+        }
     });
 }
 
--- a/macosx/CreatorWindowController.mm
+++ b/macosx/CreatorWindowController.mm
@@ -18,6 +18,7 @@
 #import "CreatorWindowController.h"
 #import "Controller.h"
 #import "NSStringAdditions.h"
+#import "DefaultAppHelper.h"
 
 typedef NS_ENUM(NSUInteger, TrackerSegmentTag) {
     TrackerSegmentTagAdd = 0,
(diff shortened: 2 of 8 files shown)


lolgear:
@nevack 

This PR finishes macOS 12 adoption ( hope so ).

macosx/Controller.mm
```diff
@@ -62,6 +62,7 @@
 #import "ExpandedPathToIconTransformer.h"
 #import "PowerManager.h"
 #import "Utils.h"
+#import "DefaultAppHelper.h"
```
lolgear, line 65: @nevack @Coeur 

Put everything in this file because ( don't want to disturb xcodeproj ) this file contains functions to work with both NSWorkspace and custom file types/URLs.

macosx/Controller.mm
```diff
@@ -3431,8 +3431,7 @@ - (BOOL)performDragOperation:(id<NSDraggingInfo>)info
                                                            options:@{ NSPasteboardURLReadingFileURLsOnlyKey : @YES }];
         NSMutableArray<NSString*>* filesToOpen = [NSMutableArray arrayWithCapacity:files.count];
         for (NSURL* file in files) {
-            if ([[NSWorkspace.sharedWorkspace typeOfFile:file.path error:NULL] isEqualToString:@"org.bittorrent.torrent"] ||
-                [file.pathExtension caseInsensitiveCompare:@"torrent"] == NSOrderedSame) {
```
lolgear, line 3435: `UTType -conformToType` has this `pathExtension` check under the hood. ( with lowercase )

macosx/DefaultAppHelper.h
```diff
@@ -3,6 +3,18 @@
 // License text can be found in the licenses/ folder.
 
 #import <Foundation/Foundation.h>
+#import <UniformTypeIdentifiers/UniformTypeIdentifiers.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface UTType (Torrent)
+@property (class, readonly, strong, nonnull) UTType *torrent;
++ (UTType*)contentTypeForFilenameExtension:(NSString*)fileExtension isFolder:(BOOL)isFolder;
```
lolgear, line 12: Not sure about nullability of this `fileExtension` parameter.

macosx/Controller.mm
```diff
@@ -957,11 +958,18 @@ - (void)URLSession:(nonnull NSURLSession*)session
     didReceiveResponse:(nonnull NSURLResponse*)response
      completionHandler:(nonnull void (^)(NSURLSessionResponseDisposition))completionHandler
 {
+    UTType* contentType = [UTType typeWithMIMEType:response.MIMEType];
     NSString* suggestedName = response.suggestedFilename;
-    if ([suggestedName.pathExtension caseInsensitiveCompare:@"torrent"] == NSOrderedSame) {
+    UTType* fileType = [UTType typeWithFilenameExtension:suggestedName.pathExtension];
```
ckerr:
```suggestion
    NSString* mimeType = response.MIMEType;
    UTType* contentType = mimeType.length > 0 ? [UTType typeWithMIMEType:mimeType] : nil;
    NSString* suggestedName = response.suggestedFilename;
    NSString* suggestedExtension = suggestedName.pathExtension;
    UTType* fileType = suggestedExtension.length > 0 ? [UTType typeWithFilenameExtension:suggestedExtension] : nil;
```
 
- `response.MIMEType` and `response.suggestedFilename` are both nullable: `MIMEType` is nil when the server sends no `Content-Type`. But `+typeWithMIMEType:` and `+typeWithFilenameExtension:` declare their parameters `nonnull`.  We need to ensure they're non-null here.

- `+contentTypeForFilenameExtension:isFolder:` below guards its own input with `fileExtension.length > 0` but these two call sites don't.

- Also, the nil path did change behavior here regardless: `[nil caseInsensitiveCompare:@"torrent"]` returns `NSOrderedSame` (zero), so a response with no suggested filename used to be accepted as a torrent and now falls through to the alert. I *think* this is a fix but wanted toa sk about it since you didn't mention it in the description?

macosx/Controller.mm
```diff
@@ -957,11 +958,18 @@ - (void)URLSession:(nonnull NSURLSession*)session
...
+    UTType* torrentType = UTType.torrent;
+
+    BOOL isTorrent = [contentType conformsToType:torrentType] || [fileType conformsToType:torrentType];
```
ckerr: Torrent classification now has three shapes: this MIME-or-extension pair, `NSURL.isTorrentFile` (content type only), and the `UTType.torrent` conformance checks at the panel call sites. They already disagree — this one accepts an extension match and `isTorrentFile` does not.

Maybe fold the predicate into the new category (`+[UTType isTorrentType:]` or `-[UTType isTorrent]`) so this fix lands in one place?

macosx/Controller.mm
```diff
@@ -973,7 +981,14 @@ - (void)URLSession:(nonnull NSURLSession*)session
         [alert addButtonWithTitle:NSLocalizedString(@"OK", "Download not a torrent -> button")];
         alert.messageText = NSLocalizedString(@"Torrent download failed", "Download not a torrent -> title");
         alert.informativeText = message;
-        [alert runModal];
+
+        NSWindow* mainWindow = NSApp.mainWindow ?: NSApp.keyWindow;
+        if (mainWindow) {
+            [alert beginSheetModalForWindow:mainWindow completionHandler:nil];
+        } else {
+            [NSApp activateIgnoringOtherApps:YES];
```
ckerr: `-[NSApplication activateIgnoringOtherApps:]` is deprecated in macOS 14.

```suggestion
            if (@available(macOS 14.0, *)) {
                [NSApp activate];
            } else {
                [NSApp activateIgnoringOtherApps:YES];
            }
```

macosx/DefaultAppHelper.mm
```diff
@@ -10,6 +10,45 @@
...
+@end
+
+@implementation NSURL (Torrent)
+- (BOOL)isTorrentFile
+{
+    UTType* contentType = nil;
+    if ([self getResourceValue:&contentType forKey:NSURLContentTypeKey error:NULL] && contentType) {
+        return [contentType conformsToType:UTType.torrent];
+    }
+    return NO;
```
ckerr: Dropping the extension fallback makes torrent detection depend entirely on the LaunchServices database.

`macosx/Info.plist.in` declares `org.bittorrent.torrent` as conforming **to** `com.bittorrent.torrent` — conformance runs that direction only. So if another installed client owns the `.torrent` extension and LS resolves a file to `com.bittorrent.torrent`, or resolves it to a `dyn.*` UTI before Transmission is registered (first launch straight from the DMG), `conformsToType:` is `NO`.

Auto-import silently stops importing and drag-and-drop stops accepting `.torrent` files, with no error shown. The old code kept working in exactly those cases.

```suggestion
    UTType* contentType = nil;
    if ([self getResourceValue:&contentType forKey:NSURLContentTypeKey error:NULL] && contentType &&
        [contentType conformsToType:UTType.torrent]) {
        return YES;
    }

    // LaunchServices resolves .torrent to another declared type when a different
    // client owns it, and to a dynamic UTI when nothing is registered yet,
    // so the extension stays authoritative.
    return [self.pathExtension caseInsensitiveCompare:@"torrent"] == NSOrderedSame;
```

macosx/DefaultAppHelper.mm
```diff
@@ -10,6 +10,45 @@
...
+}
+@end
+
 UTType* GetTorrentFileType(void)
```
ckerr: `GetTorrentFileType()` is dead now — this PR replaced both call sites, and `git grep GetTorrentFileType` matches only this definition. It's non-static, so the compiler won't flag it.

Its body is also identical to `+[UTType torrent]` above, so a future change to the exported-type conformance has two places to edit and only one that's live. Please delete lines 52-62.

macosx/DefaultAppHelper.h
```diff
@@ -3,6 +3,18 @@
...
+NS_ASSUME_NONNULL_BEGIN
+
+@interface UTType (Torrent)
+@property(class, readonly, strong, nonnull) UTType* torrent;
++ (UTType*)contentTypeForFilenameExtension:(NSString*)fileExtension isFolder:(BOOL)isFolder;
+@end
+
+@interface NSURL (Torrent)
+@property(nonatomic, readonly) BOOL isTorrentFile;
+@end
```
ckerr, line 17: Suggest moving these two categories out of `DefaultAppHelper.h` and into `NS*Additions`-style files, following the pattern already used across `macosx/`.

The concrete problem: Controller.mm, CreatorWindowController.mm, DragOverlayWindow.mm, ExpandedPathToIconTransformer.mm and Torrent.mm now all import `DefaultAppHelper.h` — the header for a class about registering as the system's default handler — to reach an icon helper unrelated to that. Anyone looking for the icon helper has no reason to open that file.

The existing convention is one file per extended class, named `NS<Class>Additions.{h,mm}`, with the category named after the file (see NSStringAdditions, NSImageAdditions, NSApplicationAdditions). Applied here that's:

**`macosx/UTTypeAdditions.h` / `.mm`**
```objc
@interface UTType (UTTypeAdditions)

@property(class, readonly) UTType* torrent;
+ (UTType*)contentTypeForFilenameExtension:(NSString*)fileExtension isFolder:(BOOL)isFolder;

@end
```

**`macosx/NSURLAdditions.h` / `.mm`**
```objc
@interface NSURL (NSURLAdditions)

@property(nonatomic, readonly) BOOL isTorrentFile;

@end
```

Add the four files to the source list in `macosx/CMakeLists.txt`, alongside the other `*Additions` entries. `DefaultAppHelper.h` then declares only `DefaultAppHelper` again, and each call site imports the one addition it actually uses.

One design note while these are moving. `contentTypeForFilenameExtension:isFolder:` makes every caller derive the `BOOL` itself, and all four derive it differently:

| call site | source of `isFolder` |
| --- | --- |
| CreatorWindowController.mm:167 | `file_count() > 1 \|\| path contains '/'` |
| DragOverlayWindow.mm:102 | `fileCount <= 1` (inverted — see that comment) |
| ExpandedPathToIconTransformer.mm:34 | extension empty && path doesn't exist |
| Torrent.mm:622 | `self.folder` |

Two of those have a real path on disk and could use a `+contentTypeForPath:` overload that answers the folder question itself, which removes the chance of getting the flag backwards. The other two (DragOverlayWindow, Torrent) take folder-ness from torrent metadata for a file that may not exist yet, so they still need the explicit flag. Adding the overload alongside the flag version would cover both.

macosx/DragOverlayWindow.mm
```diff
@@ -98,8 +99,8 @@ - (void)setTorrents:(NSArray<NSString*>*)files
 
     NSImage* icon;
     if (count == 1) {
-        icon = [NSWorkspace.sharedWorkspace
-            iconForFileType:fileCount <= 1 ? name.pathExtension : NSFileTypeForHFSTypeCode(kGenericFolderIcon)];
+        auto fileType = [UTType contentTypeForFilenameExtension:name.pathExtension isFolder:fileCount <= 1];
```
ckerr: `isFolder:` is inverted. The ternary this replaced used the folder icon when `fileCount` was **greater** than 1.

Single-file torrent (`fileCount == 1`) now shows a folder icon; multi-file torrent takes the extension of `metainfo.name()`, so a torrent named `Ubuntu 24.04` yields `pathExtension` `04`, a dynamic UTI, and a generic document icon.

```suggestion
        auto fileType = [UTType contentTypeForFilenameExtension:name.pathExtension isFolder:fileCount > 1];
```

macosx/ExpandedPathToIconTransformer.mm
```diff
@@ -29,13 +30,10 @@ - (id)transformedValue:(id)value
...
-    NSImage* icon;
     //show a folder icon if the folder doesn't exist
-    if ([path.pathExtension isEqualToString:@""] && ![NSFileManager.defaultManager fileExistsAtPath:path]) {
-        icon = [NSWorkspace.sharedWorkspace iconForFileType:NSFileTypeForHFSTypeCode(kGenericFolderIcon)];
-    } else {
-        icon = [NSWorkspace.sharedWorkspace iconForFile:path];
-    }
+    auto isFolder = [path.pathExtension isEqualToString:@""] && ![NSFileManager.defaultManager fileExistsAtPath:path];
+    auto contentType = [UTType contentTypeForFilenameExtension:path isFolder:isFolder];
+    NSImage* icon = [NSWorkspace.sharedWorkspace iconForContentType:contentType];
```
ckerr: Two problems in this block.

**The whole path is passed where a bare extension is expected.** `+contentTypeForFilenameExtension:` calls `+[UTType typeWithFilenameExtension:]`, which returns nil for `/Users/me/Downloads/incomplete`, so the helper falls through to `UTTypeData` and every path renders as a blank generic document. This transformer drives the folder icons in PrefsWindow.xib and the destination popups in AddWindowController, AddMagnetWindowController and GroupsPrefsController.

**Existing folders lose their icon even after that fix.** `isFolder` is true only when the folder does *not* exist, so `~/Downloads` takes the extension path, finds no extension, and gets `UTTypeData`. Dropping `iconForFile:` also drops per-item icons — custom folder icons, volumes, packages, app bundles. `-[NSWorkspace iconForFile:]` isn't deprecated, so it can stay.

```suggestion
    NSImage* icon;
    if ([NSFileManager.defaultManager fileExistsAtPath:path]) {
        icon = [NSWorkspace.sharedWorkspace iconForFile:path];
    } else {
        //show a folder icon if the path has no extension
        auto isFolder = path.pathExtension.length == 0;
        auto contentType = [UTType contentTypeForFilenameExtension:path.pathExtension isFolder:isFolder];
        icon = [NSWorkspace.sharedWorkspace iconForContentType:contentType];
    }
```

macosx/CreatorWindowController.mm
```diff
@@ -163,8 +164,9 @@ - (void)awakeFromNib
 
     auto const is_folder = self.fBuilder->file_count() > 1 || tr_strv_contains(self.fBuilder->path(0), '/');
 
-    NSImage* icon = [NSWorkspace.sharedWorkspace
-        iconForFileType:is_folder ? NSFileTypeForHFSTypeCode(kGenericFolderIcon) : self.fPath.pathExtension];
+    auto fileType = [UTType contentTypeForFilenameExtension:self.fPath.pathExtension isFolder:is_folder];
+
+    NSImage* icon = [NSWorkspace.sharedWorkspace iconForContentType:fileType];
     icon.size = self.fIconView.frame.size;
```
ckerr, line 171: `-[NSWorkspace iconForContentType:]` returns an image from a per-type cache, and this PR funnels four call sites into the same handful of types.

After the change, this window, ExpandedPathToIconTransformer (16x16), DragOverlayWindow and Torrent.mm all ask for `UTTypeFolder` or `UTTypeData` and get the same instance back. Let the prefs pane transform a path while a Create-Torrent window is open and the creator's large icon shrinks to 16x16 with it. Previously these paths returned distinct objects (`imageNamed:NSImageNameFolder`, per-path `iconForFile:`).

`icon = [icon copy];` before setting `size` is enough. Same at ExpandedPathToIconTransformer.mm:38.

macosx/MessageWindowController.mm
```diff
@@ -433,7 +433,7 @@ - (void)clearLog:(id)sender
 - (void)writeToFile:(id)sender
 {
     NSSavePanel* panel = [NSSavePanel savePanel];
-    panel.allowedFileTypes = @[ @"txt" ];
+    panel.allowedContentTypes = @[ UTTypeText ];
```
ckerr: `UTTypeText` is `public.text`, an abstract type with no `public.filename-extension` tag. NSSavePanel takes the appended extension from the first allowed type's `preferredFilenameExtension`, so the log now saves as `untitled` with no extension, and `canSelectHiddenExtension` has nothing to hide. The panel also accepts .rtf, .html, .swift and anything else conforming to `public.text` for content that's always UTF-8 plain text.

```suggestion
    panel.allowedContentTypes = @[ UTTypePlainText ];
```

Also, need to import `<UniformTypeIdentifiers/UniformTypeIdentifiers.h>` in this file for the UTType.

ckerr (changes requested):

macosx/DragOverlayWindow.mm
lolgear: Oh, yes-yes, condition inversion.

macosx/Controller.mm
lolgear: Added.

macosx/Controller.mm
lolgear: Added.

macosx/DefaultAppHelper.mm
lolgear: Added.

macosx/DefaultAppHelper.mm
lolgear: Removed.

macosx/MessageWindowController.mm
lolgear: Added.

macosx/Controller.mm
lolgear: Actually, we could remove response.MIMEType?
Not sure if it is a necessary change here.

macosx/DefaultAppHelper.h
lolgear: Yes, I would like to add files, but right now we have two good PRs in Draft. ( one for Xcode 16 and another for conversion groups to folders ).
These PRs untangle complexity of files additions.

lolgear:
@ckerr 
Could you check PR?

PR #284, opened 2026-08-31
Move error badge rendering from table view to torrent cell.

lolgear:
## Description
This PR refactors the error badge rendering logic for torrent rows in the macOS client. 

Previously, `TorrentTableView` manually drew the caution badge directly onto the torrent's icon image using focus-locking drawing operations. This approach has been replaced by delegating the error state management to the cells via a new method `-setAnyErrorOrWarning:`. 

`TorrentCell` now manages an internal `NSImageView` (`_errorImageView`) overlayed using modern Auto Layout constraints. `SmallTorrentCell` implements a stub for this method since it handles error status representation via its main icon view.

## Changes
- **TorrentCell.h / .mm**: Added `-setAnyErrorOrWarning:` method and implemented dynamic initialization and layout constraints for `_errorImageView`.
- **SmallTorrentCell.mm**: Implemented an empty stub for `-setAnyErrorOrWarning:` to conform to the shared interface.
- **TorrentTableView.mm**: Simplified the row configuration by removing legacy image-drawing code and replacing it with the new cell method call.

## Type of Change
- [x] Refactoring (non-breaking change which improves code structure)

## Verification
- Verified that the error badge (`NSImageNameCaution`) appears correctly positioned over the torrent icon when a torrent encounters an error.
- Verified that the badge disappears when the error state is cleared.
- Checked that `SmallTorrentCell` does not crash or layout incorrectly during state updates.

### DIFF
--- a/macosx/SmallTorrentCell.mm
+++ b/macosx/SmallTorrentCell.mm
@@ -151,6 +151,11 @@ static CGFloat const kButtonsSpacing = 3.0;
     ]];
 }
 
+- (void)setAnyErrorOrWarning:(BOOL)errorOrWarning
+{
+    // nothing to do. we use iconView for error status.
+}
+
 // show fControlButton and fRevealButton
 - (void)mouseEntered:(NSEvent*)event
 {
--- a/macosx/TorrentCell.mm
+++ b/macosx/TorrentCell.mm
@@ -35,6 +35,13 @@ static CGFloat const kButtonSize = 14.0;
 static CGFloat const kButtonsSpacing = 3.0;
 static CGFloat const kRevealButtonTrailingOffset = -8.0; // inverted for constraints.
 
+// Error status
+static CGFloat const kErrorImageSize = 20.0;
+
+@interface TorrentCell ()
+@property(nonatomic, readonly) NSImageView* errorImageView;
+@end
+
 @implementation TorrentCell
 
 - (instancetype)initWithFrame:(NSRect)frameRect
@@ -307,6 +314,28 @@ static CGFloat const kRevealButtonTrailingOffset = -8.0; // inverted for constra
 
     NSColor* priorityColor = backgroundStyle == NSBackgroundStyleEmphasized ? NSColor.whiteColor : NSColor.labelColor;
     self.fTorrentPriorityView.contentTintColor = priorityColor;
+}
+
+- (void)setAnyErrorOrWarning:(BOOL)errorOrWarning
+{
+    if (errorOrWarning) {
+        if (_errorImageView == nil) {
+            _errorImageView = [[NSImageView alloc] init];
+            _errorImageView.imageScaling = NSImageScaleProportionallyDown;
+            _errorImageView.image = [NSImage imageNamed:NSImageNameCaution];
+            [self.fIconView addSubview:_errorImageView];
+            _errorImageView.translatesAutoresizingMaskIntoConstraints = NO;
+
+            [NSLayoutConstraint activateConstraints:@[
+                [_errorImageView.leadingAnchor constraintEqualToAnchor:self.fIconView.centerXAnchor],
+                [_errorImageView.topAnchor constraintEqualToAnchor:self.fIconView.centerYAnchor],
+                [_errorImageView.widthAnchor constraintEqualToConstant:kErrorImageSize],
+                [_errorImageView.heightAnchor constraintEqualToConstant:kErrorImageSize],
+            ]];
+        }
+    }
+
+    _errorImageView.hidden = !errorOrWarning;
 }
 
 @end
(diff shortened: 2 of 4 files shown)


macosx/TorrentCell.mm
```diff
@@ -309,4 +316,26 @@ - (void)setBackgroundStyle:(NSBackgroundStyle)backgroundStyle
...
+        if (_errorImageView == nil) {
+            _errorImageView = [[NSImageView alloc] init];
+            _errorImageView.imageScaling = NSImageScaleProportionallyDown;
+            _errorImageView.image = [NSImage imageNamed:NSImageNameCaution];
+            [self.fIconView addSubview:_errorImageView];
+            _errorImageView.translatesAutoresizingMaskIntoConstraints = NO;
+
+            [NSLayoutConstraint activateConstraints:@[
+                [_errorImageView.leadingAnchor constraintEqualToAnchor:self.fIconView.centerXAnchor],
+                [_errorImageView.topAnchor constraintEqualToAnchor:self.fIconView.centerYAnchor],
```
ckerr: (Opus 5.5 xhigh) These constraints push the badge 2pt past the icon view's right and bottom edges. Pin its trailing and bottom edges to the icon view instead. On each axis, the badge spans 18pt to 38pt of a 36pt icon view. macOS 12 and 13 clip subviews to their superview, so those systems clip the badge's last 2pt.

```suggestion
                [_errorImageView.trailingAnchor constraintEqualToAnchor:self.fIconView.trailingAnchor],
                [_errorImageView.bottomAnchor constraintEqualToAnchor:self.fIconView.bottomAnchor],
```

macosx/TorrentCell.mm
```diff
@@ -309,4 +316,26 @@ - (void)setBackgroundStyle:(NSBackgroundStyle)backgroundStyle
     self.fTorrentPriorityView.contentTintColor = priorityColor;
 }
 
+- (void)setAnyErrorOrWarning:(BOOL)errorOrWarning
+{
+    if (errorOrWarning) {
+        if (_errorImageView == nil) {
+            _errorImageView = [[NSImageView alloc] init];
```
ckerr: (Opus 5.5 xhigh) This setter builds the badge view on first use, apart from the rest of the cell's layout. Create it in `configureViews` and constrain it in `setupConstraints`, next to `fIconView`. That keeps the layout in one place and shrinks this setter to `self.fErrorImageView.hidden = !errorOrWarning;`.

macosx/SmallTorrentCell.mm
```diff
@@ -151,6 +151,11 @@ - (void)setupConstraints
     ]];
 }
 
+- (void)setAnyErrorOrWarning:(BOOL)errorOrWarning
+{
+    // nothing to do. we use iconView for error status.
+}
```
ckerr: (Opus 5.5 xhigh) This override never runs, because TorrentTableView.mm:248 calls `setAnyErrorOrWarning:` only for regular cells. Move the small cell's caution-icon swap from TorrentTableView.mm:216 into this override. Then call `setAnyErrorOrWarning:` from `-[TorrentCell setObjectValue:]`, next to the priority update. Together, these changes finish moving error rendering out of the table view.

macosx/TorrentTableView.mm
```diff
@@ -245,28 +245,8 @@ - (NSView*)outlineView:(NSOutlineView*)outlineView viewForTableColumn:(NSTableCo
...
-                NSRect frame = torrentCell.fIconView.frame;
-                NSImage* resultImage = [[NSImage alloc] initWithSize:frame.size];
-                [resultImage lockFocus];
-
-                // draw fileImage
-                [fileImage drawAtPoint:NSZeroPoint fromRect:NSZeroRect operation:NSCompositingOperationSourceOver fraction:1.0];
-
-                // overlay error badge
-                NSImage* errorImage = [NSImage imageNamed:NSImageNameCaution];
-                NSRect const errorRect = NSMakeRect(frame.origin.x, 0, kErrorImageSize, kErrorImageSize);
```
ckerr: (Opus 5.5 xhigh) You removed the last use of `kErrorImageSize` here. Delete its definition at line 27. The macOS CI jobs on this PR already warn that it is unused.

ckerr (changes requested):
(Opus 5.5 xhigh)

Please fix the badge overflow and the unused `kErrorImageSize` before merge. The other two comments suggest optional cleanups.

Add a `Notes:` paragraph to the description or the `notes:none` label. The badge moves only a few points, so `notes:none` fits.

macosx/SmallTorrentCell.mm
lolgear: SmallTorrentCell replaces icon with caution image.
TorrentCell shows two images.
It is impossible to do with current setter's signature and requires more changes which are out of scope of this PR.

ckerr: (Opus 5.5 xhigh) Agreed. We can move the small cell's icon swap in a follow-up. TorrentTableView.mm:240 calls `setAnyErrorOrWarning:` only on regular cells, so this override never runs. Please delete it.

PR #286, opened 2026-08-31
perf: Optimize group ratio toggle

lolgear:
## Description

This PR optimizes the `toggleGroupRowRatio` method in `TorrentTableView`. Instead of triggering a full `reloadVisibleRows` which refreshes all visible rows regardless of their type, the updated logic now targets only the visible `TorrentGroup` rows and updates their cells directly.

## Changes

- Replaced the heavy `[self reloadVisibleRows]` call with a selective loop over visible rows.
- Added a guard clause to return early if `SortByGroup` is disabled.
- Dynamically fetches and updates `GroupCell` instances using `setDisplayRatio:` for currently visible group items.

## Performance Impact

Reduces UI overhead when toggling the group row ratio, as non-group cells are no longer unnecessarily reloaded.

### DIFF
--- a/macosx/TorrentTableView.mm
+++ b/macosx/TorrentTableView.mm
@@ -609,9 +609,25 @@ static NSTimeInterval const kToggleProgressSeconds = 0.175;
 
 - (void)toggleGroupRowRatio
 {
-    BOOL displayGroupRowRatio = [self.fDefaults boolForKey:@"DisplayGroupRowRatio"];
-    [self.fDefaults setBool:!displayGroupRowRatio forKey:@"DisplayGroupRowRatio"];
-    [self reloadVisibleRows];
+    BOOL updatedDisplayGroupRowRatio = ![self.fDefaults boolForKey:@"DisplayGroupRowRatio"];
+    [self.fDefaults setBool:updatedDisplayGroupRowRatio forKey:@"DisplayGroupRowRatio"];
+
+    if (![self.fDefaults boolForKey:@"SortByGroup"]) {
+        return;
+    }
+
+    NSRange visibleRows = [self rowsInRect:self.visibleRect];
+
+    for (NSUInteger i = 0; i < visibleRows.length; ++i) {
+        NSUInteger row = visibleRows.location + i;
+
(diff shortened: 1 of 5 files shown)


macosx/TorrentTableView.mm
```diff
@@ -642,9 +609,25 @@ - (void)hoverEventEndedForView:(id)view
...
+    for (NSUInteger i = 0; i < visibleRows.length; ++i) {
+        NSUInteger row = visibleRows.location + i;
+
+        id item = [self itemAtRow:row];
+
+        if ([item isKindOfClass:[TorrentGroup class]]) {
+            GroupCell* groupCell = (GroupCell*)[self viewAtColumn:0 row:row makeIfNecessary:NO];
+            [groupCell setDisplayRatio:updatedDisplayGroupRowRatio];
+        }
+    }
```
ckerr: (Opus 5.5 xhigh) This loop saves one reload of the visible rows per click, and the PR gives no measurement of that saving. Please keep `[self reloadVisibleRows]` unless you can measure a cost users would notice. `-[Controller updateUI]` already runs the same reload every second at Controller.mm:2082. If one reload were slow enough to notice, the table would stutter every second.

The loop also flips `hidden` without setting any titles. It shows the right values only because `setDownloadSpeed:uploadSpeed:ratio:` also formats the hidden button's title. With the reload, that method can skip the hidden title, as the open thread on #264 asks.

```suggestion
    [self reloadVisibleRows];
```

ckerr (commented):
(Opus 5.5 xhigh) This PR's first commit, edcdfbbc9, duplicates #264's first commit. You have added three commits to #264 since then. Please rebase onto #264's latest head now, and onto main once #264 merges. If the loop in `toggleGroupRowRatio` stays, call `updateDisplayRatio:` there. #264 renamed `setDisplayRatio:`, so the old name won't compile after the rebase. This review covers only the perf commit, because #264's review covers the GroupCell changes.

Please add the `notes:none` label, since the perf commit changes nothing users see.

Please settle the rebase and the inline comment on `toggleGroupRowRatio` before this leaves draft.

PR #290, opened 2026-09-03
refactor: remove redundant makeKeyWindow calls on mouse down.

lolgear:
## Description
This PR removes the redundant `mouseDown:` method overrides from `FileOutlineView`, `TrackerTableView`, and `WebSeedTableView`. 

Previously, these classes explicitly called `[self.window makeKeyWindow]` before passing the event to `[super mouseDown:]`. This explicit activation is unnecessary because AppKit inherently handles window focusing and text/element selection correctly during mouse-down events on these standard controls. 

Removing these overrides cleans up unnecessary boilerplate code and prevents potential side effects related to window ordering and focus management.

## Changes
- **`macosx/FileOutlineView.mm`**: Removed custom `mouseDown:` implementation.
- **`macosx/TrackerTableView.mm`**: Removed custom `mouseDown:` implementation.
- **`macosx/WebSeedTableView.mm`**: Removed custom `mouseDown:` implementation.

## Testing Done
- [x] Verified that clicking on the elements inside File Outline, Tracker Table, and Web Seed Table still correctly focuses the main window.
- [x] Verified that standard mouse interaction (selection, context menus) functions normally.

### DIFF
--- a/macosx/FileOutlineView.mm
+++ b/macosx/FileOutlineView.mm
@@ -2,11 +2,8 @@
 // It may be used under the MIT (SPDX: MIT) license.
 // License text can be found in the licenses/ folder.
 
-#import "InfoWindowController.h"
-#import "FileListNode.h"
 #import "BaseFileNameCellView.h"
 #import "FileOutlineView.h"
-#import "Torrent.h"
 
 @interface FileOutlineView ()
 
--- a/macosx/FileOutlineView.mm
+++ b/macosx/FileOutlineView.mm
@@ -22,12 +22,6 @@
     self.indentationPerLevel = 14.0;
 }
 
-- (void)mouseDown:(NSEvent*)event
-{
-    [self.window makeKeyWindow];
-    [super mouseDown:event];
-}
-
 - (NSMenu*)menuForEvent:(NSEvent*)event
 {
     NSInteger const row = [self rowAtPoint:[self convertPoint:event.locationInWindow fromView:nil]];
--- a/macosx/InfoWindowController.mm
+++ b/macosx/InfoWindowController.mm
@@ -89,7 +89,9 @@ typedef NS_ENUM(NSUInteger, TabTag) {
     windowRect.size.height = windowHeight;
     [window setFrame:windowRect display:NO];
 
-    // Let inspector gain keyboard focus when clicked on non-interactive areas
+    // Any click makes the inspector the key window.
+    // Table views return NO from needsPanelToBecomeKey.
+    // Without this, the file, tracker and web seedlists would not focus it.
     window.becomesKeyOnlyIfNeeded = NO;
 
     //disable green maximise window button
--- a/macosx/InfoWindowController.mm
+++ b/macosx/InfoWindowController.mm
@@ -91,7 +91,7 @@ typedef NS_ENUM(NSUInteger, TabTag) {
 
     // Any click makes the inspector the key window.
     // Table views return NO from needsPanelToBecomeKey.
-    // Without this, the file, tracker and web seedlists would not focus it.
+    // Without this, the file, tracker and web seed lists would not focus it.
     window.becomesKeyOnlyIfNeeded = NO;
 
     //disable green maximise window button
--- a/macosx/TrackerTableView.mm
+++ b/macosx/TrackerTableView.mm
@@ -8,12 +8,6 @@
 
 @implementation TrackerTableView
 
-- (void)mouseDown:(NSEvent*)event
-{
-    [self.window makeKeyWindow];
-    [super mouseDown:event];
-}
-
 - (void)copy:(id)sender
 {
     NSMutableArray* addresses = [NSMutableArray arrayWithCapacity:self.trackers.count];
--- a/macosx/WebSeedTableView.mm
+++ b/macosx/WebSeedTableView.mm
@@ -6,12 +6,6 @@
 
 @implementation WebSeedTableView
 
-- (void)mouseDown:(NSEvent*)event
-{
-    [self.window makeKeyWindow];
-    [super mouseDown:event];
-}
-
 - (void)copy:(id)sender
 {
     NSIndexSet* indexes = self.selectedRowIndexes;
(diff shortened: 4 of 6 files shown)


macosx/FileOutlineView.mm
```diff
@@ -22,12 +22,6 @@ - (void)awakeFromNib
     self.indentationPerLevel = 14.0;
 }
 
-- (void)mouseDown:(NSEvent*)event
```
ckerr, line 25: After this deletion, this file references only `BaseFileNameCellView` and AppKit so  `InfoWindowController.h` (line 5), `FileListNode.h` (line 6) and `Torrent.h` (line 9) all unused and can be remove.d

The `@interface FileOutlineView ()` at lines 11-13 is empty too. (I think)

Lines 5-15 become:

```objc
#import "BaseFileNameCellView.h"
#import "FileOutlineView.h"

@implementation FileOutlineView
```

macosx/FileOutlineView.mm
```diff
@@ -22,12 +22,6 @@ - (void)awakeFromNib
...
-{
-    [self.window makeKeyWindow];
```
ckerr, line 27: These are redundant and removing them is right, but only as long as `becomesKeyOnlyIfNeeded = NO` is set.  See d498cc77d9a96da77cfdd42f71a3ede08c1c033d. Let's add a comment to `InfoWindowController.mm:92` to make this explicit:

```objc
// Any click makes the inspector the key window.
// Table views return NO from needsPanelToBecomeKey.
// Without this, the file, tracker and web seedlists would not focus it.
window.becomesKeyOnlyIfNeeded = NO;
```

macosx/TrackerTableView.mm
```diff
@@ -8,12 +8,6 @@
 
 @implementation TrackerTableView
 
-- (void)mouseDown:(NSEvent*)event
-{
-    [self.window makeKeyWindow];
```
ckerr, line 13: (opus 5 extra)

Also remove this redundant one in  `macosx/InfoTrackersViewController.mm:284`

```objc
- (void)addTrackers
{
    [self.view.window makeKeyWindow];
```

Redundant for exactly the reason this PR deletes the three overrides: `addTrackers` is reachable only from `addRemoveTracker:` (line 249), i.e. a click on the +/- segmented control inside the Inspector panel, which has already keyed it.

Removing it too keeps the sweep coherent. Otherwise a reader sees the identical idiom deleted in three files and kept in a fourth

macosx/WebSeedTableView.mm
```diff
@@ -6,12 +6,6 @@
 
 @implementation WebSeedTableView
 
-- (void)mouseDown:(NSEvent*)event
-{
-    [self.window makeKeyWindow];
```
ckerr, line 11: (opus 5 extra) 

**One more dead `makeKeyWindow`, if you want the sweep complete.**

`macosx/Controller.mm:656`
```objc
[nc addObserver:self.fWindow selector:@selector(makeKeyWindow) name:@"MakeWindowKey" object:nil];
```

`grep -rn MakeWindowKey --exclude-dir=.git .` returns exactly one hit: this line. Nothing posts the notification, so the observer can never fire. Same family as what this PR removes.

macosx/WebSeedTableView.mm
```diff
@@ -6,12 +6,6 @@
...
-    [super mouseDown:event];
```
ckerr, line 12: (opus 5 extra)

**Pre-existing, but this PR makes it the last one standing.**

After this change `PeerTableView` is the only `mouseDown:` override left among the Inspector tables, and it's actually broken:

`macosx/PeerTableView.mm:9`
```objc
- (void)mouseDown:(NSEvent*)event
{
    NSPoint point = [self convertPoint:event.locationInWindow fromView:nil];
    if ([self rowAtPoint:point] != -1 && ...) { ... }
}
```

No `[super mouseDown:event]` on any path, so `NSTableView` never sees the click: no peer row is selectable by mouse, even though `InfoPeersView.xib:92` leaves single selection enabled on that table (`multipleSelection="NO"`, no `selectionHighlightStyle="none"`).

Not this PR's bug — flagging it only so the omission isn't later read as "reviewed and deliberately kept".

ckerr (commented):
Comments beginning with `(opus 5)` were created with opus 5, extra effort, with the prompt

> /code-review PR 290. Focus on findings that are new to this PR -- ignore pre-existing bugs unless they're made worse by this PR

macosx/WebSeedTableView.mm
lolgear: Yes, I would like to open additional PR.
It will convert mouseDown actions to GestureRecognizers ( as Apple wants ).

macosx/WebSeedTableView.mm
lolgear: Wow, done.

macosx/TrackerTableView.mm
lolgear: Hm, ok, done.
Interesting. 
That is why we want to migrate to Swift as soon as possible ( private functions ).

macosx/FileOutlineView.mm
lolgear: Done.

macosx/FileOutlineView.mm
lolgear: Done.

macosx/InfoWindowController.mm
```diff
@@ -89,7 +89,9 @@ - (void)awakeFromNib
     windowRect.size.height = windowHeight;
     [window setFrame:windowRect display:NO];
 
-    // Let inspector gain keyboard focus when clicked on non-interactive areas
+    // Any click makes the inspector the key window.
+    // Table views return NO from needsPanelToBecomeKey.
+    // Without this, the file, tracker and web seedlists would not focus it.
```
ckerr: _(nit)_

```suggestion
    // Without this, the file, tracker and web seed lists would not focus it.
```

ckerr (commented):
LGTM other than a minor typo

macosx/InfoWindowController.mm
lolgear: Done.

ckerr (approved):

Issue #289, opened 2026-09-03
New considerations for macOS 12 Monterey


lolgear:
Since project is moving towards new interfaces (UI and code), there is a gap between current codebase and actual "Apple Design Standards" in both "Human Design" and "Code Design".

Let's keep an eye on these gaps.

### Why You Should NOT Override mouseDown: in Modern macOS

* Breaks Cross-Device Input (Sidecar & Apple Pencil)
mouseDown: is hardcoded for physical mice. If a user uses Sidecar to turn an iPad into a second screen and taps your view with an Apple Pencil, AppKit won't translate the touch phases correctly [TN3212]. Apple explicitly mandates Gesture Recognizers to abstract hardware differences [TN3212].
* Interferes with System Gestures (Force Touch, Scrolling, Drag-and-Drop)
macOS handles complex multi-phase interactions at the window manager level (Force Click lookups, smart zoom, drag-and-drop initiation). Intercepting raw mouseDown: events often "swallows" the input stream, confusing the system about whether the user intended to click, scroll, or start dragging.
* Violates Apple Human Interface Guidelines (HIG)
Standard Mac interactions trigger actions on release (mouseUp), not on press (mouseDown). This gives users a chance to cancel the action by dragging the cursor away before lifting their finger. Triggering immediate side-effects inside mouseDown: creates a jarring, un-Mac-like UX.
* Blocks the Main Thread (Legacy Tracking Loops)
Historically, overriding mouseDown: involved writing custom while loops to poll for mouse movement. In modern macOS, blocking the main run loop like this severely degrades performance, breaking features like responsive scrolling (NSScrollView) and fluid UI animations.

Some parts of project still contain these overrides of `mouseDown` events.
And it is save to remove them in current macOS Monterey.

### Massive View Controllers (The "God Object" Syndrome)

* Legacy Approach: Projects like Transmission historically relied on a monolithic controller (e.g., Controller.mm) spanning thousands of lines. It simultaneously handled app lifecycles, UI rendering, toolbar management, business logic, and network/RPC communications.
* Why it breaks: It violates the Single Responsibility Principle, creates high coupling, completely breaks unit-testing capabilities, and makes it impossible to refactor individual components or port views to SwiftUI.
* Modern Standard: Decompose into small, focused, and testable micro-services (e.g., separating the UI into dedicated ToolbarController, AutoImportController, and background operations into isolated business logic classes).

### Cell-Based Tables (NSTableView)

[...]

PR #296, opened 2026-09-04
refactor: remove custom TrackerTableView and move clipboard logic to controller

lolgear:
## Description

This PR addresses the "Code Design" modernization goals outlined in #289. It eliminates the legacy custom `TrackerTableView` subclass, reverting the table in `InfoTrackersView.xib` to a standard `NSTableView`. 

All clipboard operations (`copy:` / `paste:`) and menu validation logic have been refactored and inline-migrated directly into `InfoTrackersViewController` using modern AppKit approaches, adhering to the Single Responsibility Principle.

## Related Issues
Addresses parts of the codebase modernization discussed in #289.

## Changes
- **UI (`InfoTrackersView.xib`)**: Replaced the custom `TrackerTableView` class with a standard `NSTableView`.
- **Controller (`InfoTrackersViewController.mm`)**: 
  - Updated `fTrackerTable` property to use `NSTableView`.
  - Implemented the modern `-tableView:pasteboardWriterForRow:` protocol method for cleaner, safer item preparation for the pasteboard.
  - Migrated and adapted `-copy:`, `-paste:`, and `-validateMenuItem:` methods from the old view class.
  - Removed obsolete `.trackers` and `.torrent` property bindings.
- **Cleanup**: Deleted `TrackerTableView.h` and `TrackerTableView.mm` as they are no longer needed.


lolgear:
Yes, I would like to merge this PR, however, it is not possible right now, because I don't want to break another PR which updates Xcode version  to 16.3.

lolgear:
Ready for review.

macosx/InfoTrackersViewController.mm
```diff
@@ -131,6 +126,17 @@ - (NSInteger)numberOfRowsInTableView:(NSTableView*)tableView
...
+{
+    id item = self.fTrackers[row];
+
+    if ([item isKindOfClass:[NSDictionary class]]) {
+        return nil;
+    } else {
+        return [(TrackerNode*)item fullAnnounceAddress];
+    }
+}
+
```
ckerr: (Opus 5.5 xhigh) This data source method makes the tracker table a drag source. Users who click-drag down the list now start a drag instead of selecting a range. Please delete it and look up the address in `copy:`, as the suggestion there does.

```suggestion
```

macosx/InfoTrackersViewController.mm
```diff
@@ -405,11 +404,67 @@ - (void)removeTrackers
...
     [NSNotificationCenter.defaultCenter postNotificationName:@"UpdateUI" object:nil]; //in case sort by tracker
 }
 
+- (void)copy:(id)sender
+{
+    NSMutableArray* addresses = [NSMutableArray arrayWithCapacity:self.fTrackers.count];
+    NSIndexSet* indexes = self.fTrackerTable.selectedRowIndexes;
+
+    [indexes enumerateIndexesUsingBlock:^(NSUInteger idx, BOOL* _Nonnull stop) {
+        [addresses addObject:[self tableView:self.fTrackerTable pasteboardWriterForRow:idx]];
```
ckerr: (Opus 5.5 xhigh) `copy:` throws `NSInvalidArgumentException` when a tier row is selected, because `addObject:` receives nil. Please copy only `TrackerNode` rows. `updateInfo` reloads the table without updating the selection. If someone adds or removes a tracker above the selected row, the selection can land on a tier row. `TrackerTableView`'s `copy:` handled tier rows before this change.

```suggestion
        id item = self.fTrackers[idx];
        if ([item isKindOfClass:[TrackerNode class]]) {
            [addresses addObject:((TrackerNode*)item).fullAnnounceAddress];
        }
```

ckerr (changes requested):
(Opus 5.5 xhigh) Please add the `notes:none` label. This refactor makes no user-facing change once the tracker table stops being a drag source.

macosx/InfoTrackersViewController.mm
lolgear: Yes, this is handled by `pasteboardWriterForRow` which returns nil for tiers.

ckerr (approved):

PR #231, opened 2026-08-14
build: Update Xcode project version to Xcode 16.0

nevack:

PR #297, opened 2026-09-04
refactor: remove custom PeerTableView and replace mouseDown override with target and action

lolgear:
## Description

This PR resolves the `mouseDown:` override issue mentioned in #289. Overriding `mouseDown:` in modern macOS is highly discouraged as it interferes with system gestures, blocks tracking loops, and degrades performance. 

This change completely removes the custom `PeerTableView` subclass, rolls the table back to a standard `NSTableView` in `InfoPeersView.xib`, and refactors the click-handling logic into the controller using standard AppKit Target-Action.

## Related Issues

Addresses the `mouseDown:` removal goal outlined in #289.

## Changes

- **UI (`InfoPeersView.xib`)**: Changed the table view class from `PeerTableView` to standard `NSTableView`.
- **Controller (`InfoPeersViewController.mm`)**:
  - Set the controller as the target and implemented `-tableViewDidClick:` action for the peer table.
  - Migrated the `DisplayPeerProgressBarNumber` toggle logic from the old view class.
  - Optimized the UI refresh code: instead of reloading all rows in the dataset, it now explicitly targets only the `visibleRect` rows within the "Progress" column.
- **Cleanup**: Deleted `PeerTableView.h` and `PeerTableView.mm`.


macosx/InfoPeersViewController.mm
```diff
@@ -87,6 +87,27 @@ - (void)awakeFromNib
...
+    self.fPeerTable.target = self;
+    self.fPeerTable.action = @selector(tableViewDidClick:);
+}
+
+- (void)tableViewDidClick:(NSTableView*)sender
+{
+    auto row = sender.clickedRow;
+    auto column = sender.clickedColumn;
+
+    NSLog(@"row: %@ column: %@", @(row), @(column));
```
ckerr: Every other NSLog in macosx/ is an error path; this is the only tracing call. Let's delete it

macosx/PeerTableView.h
```diff
@@ -1,9 +1 @@
-// This file Copyright © Transmission authors and contributors.
-// It may be used under the MIT (SPDX: MIT) license.
-// License text can be found in the licenses/ folder.
 
-#import <AppKit/AppKit.h>
-
-@interface PeerTableView : NSTableView
-
-@end
```
ckerr, line 9: These files have been emptied out but are still present as 1-byte files. Shouldn't we just delete the files?

lolgear: Of course, but I want to merge another PR which touches Xcode project file.
It would be very difficult to rebase that PR.

macosx/InfoPeersViewController.mm
lolgear: Hm, it is strange.
But yes, this one is debug print. ( better than breakpoints )

macosx/PeerTableView.h
ckerr: OK, if you want we can wait for #231 and #224 to land first.

macosx/InfoPeersViewController.mm
ckerr: Yes I agree it's a debug print, I already said so while asking you to remove it

ckerr (dismissed):

macosx/PeerTableView.h
lolgear: Rebased.

lolgear:
Ready for review.

ckerr (approved):

PR #298, opened 2026-09-04
refactor: remove custom WebSeedTableView and move clipboard logic to controller

lolgear:
## Description

This PR continues the effort to refactor legacy views and remove redundant sub-classes as discussed in #289. It eliminates the custom `WebSeedTableView` class and updates `InfoPeersView.xib` to use a standard `NSTableView` for the web seeds list.

The copy-to-clipboard functionality (`copy:`) and context menu validation (`validateMenuItem:`) have been consolidated into `InfoPeersViewController`. The new implementation uses standard AppKit delegation and introduces a generic `copy:` handler capable of processing clipboard events for table views cleanly.

## Related Issues
Addresses the table view refactoring goals described in #289.

## Changes
- **UI (`InfoPeersView.xib`)**: Swapped `WebSeedTableView` for standard `NSTableView`.
- **Controller (`InfoPeersViewController.mm`)**:
  - Declared compliance with `<NSTableViewDataSource, NSTableViewDelegate, NSMenuItemValidation>`.
  - Implemented `-tableView:pasteboardWriterForRow:` to safely retrieve the `"Address"` string value for a web seed row.
  - Added a generic `-copy:` method that checks if the sender is an `NSTableView` and copies its selected items separated by newlines.
  - Ported `-validateMenuItem:` to ensure the Copy option is enabled only when rows are selected.
  - Cleared legacy `.webSeeds` data assignments on the table view.
- **Cleanup**: Deleted `WebSeedTableView.h` and `WebSeedTableView.mm`.


macosx/InfoPeersViewController.mm
```diff
@@ -307,6 +305,23 @@ - (NSInteger)numberOfRowsInTableView:(NSTableView*)tableView
...
+        return self.fWebSeeds[row][@"Address"];
+    }
+
+    else if (tableView == self.fPeerTable) {
+        return nil; // look at tableView:(NSTableView*)tableView shouldSelectRow:(NSInteger)row
+    }
+
+    return nil;
+}
+
```
ckerr: (Opus 5.5 xhigh) This data source method makes the web seed table a drag source. Users who click-drag down the list now start a drag instead of selecting a range. Please delete it and read the addresses in `copy:`, as the suggestion there does.

```suggestion
```

macosx/InfoPeersViewController.mm
```diff
@@ -570,4 +585,39 @@ - (void)setWebSeedTableHidden:(BOOL)hide animate:(BOOL)animate
...
+                [results addObject:result];
+            }
+        }];
+
+        NSString* text = [results componentsJoinedByString:@"\n"];
+
+        NSPasteboard* pb = NSPasteboard.generalPasteboard;
+        [pb clearContents];
+        [pb writeObjects:@[ text ]];
+    }
```
ckerr: (Opus 5.5 xhigh) This `copy:` never copies anything. AppKit sends ⌘C and Edit > Copy with the menu item as `sender`, so the `NSTableView` check always fails. Please read the selection from `self.fWebSeedTable` instead. The suggestion ports the `copy:` body from the deleted `WebSeedTableView`. It also clears the CI warning about the unused `stop` parameter.

```suggestion
    NSIndexSet* indexes = self.fWebSeedTable.selectedRowIndexes;
    NSMutableArray* addresses = [NSMutableArray arrayWithCapacity:indexes.count];
    [self.fWebSeeds enumerateObjectsAtIndexes:indexes options:0
                                   usingBlock:^(NSDictionary* webSeed, NSUInteger /*idx*/, BOOL* /*stop*/) {
                                       [addresses addObject:webSeed[@"Address"]];
                                   }];

    NSString* text = [addresses componentsJoinedByString:@"\n"];

    NSPasteboard* pb = NSPasteboard.generalPasteboard;
    [pb clearContents];
    [pb writeObjects:@[ text ]];
```

macosx/InfoPeersViewController.mm
```diff
@@ -570,4 +585,39 @@ - (void)setWebSeedTableHidden:(BOOL)hide animate:(BOOL)animate
...
+}
+
+- (BOOL)validateMenuItem:(NSMenuItem*)menuItem
+{
+    SEL const action = menuItem.action;
+    id target = menuItem.target;
+
+    if (action == @selector(copy:) && [target isKindOfClass:NSTableView.class]) {
+        return ((NSTableView*)target).numberOfSelectedRows > 0;
```
ckerr: (Opus 5.5 xhigh) This method enables Copy even when no web seed is selected. Please check `self.fWebSeedTable` directly, because `menuItem.target` is nil for Edit > Copy. This controller also receives Copy while the peer table has focus, so the suggestion checks focus too.

```suggestion
    SEL const action = menuItem.action;

    if (action == @selector(copy:)) {
        return self.view.window.firstResponder == self.fWebSeedTable && self.fWebSeedTable.numberOfSelectedRows > 0;
```

ckerr (changes requested):
(Opus 5.5 xhigh) Please add the `notes:none` label. This refactor makes no user-facing change once Copy works again and the web seed table stops being a drag source.

PR #301, opened 2026-09-05
Fix moving files split across torrent directories

### LINKED ISSUE #300: Move Data may report success while leaving partial torrent data in its old directory
## What is the issue?

## Reproducer

2. Add a multi-file torrent.

## Actual result

## Expected result

Move Data should either:

None

_No response_

h1de0x:
Fixes #300

Notes: Fixed moving, completing, or removing a torrent leaving some of its files behind when they were split between the download and incomplete folders.

## Summary

Fixes the case where **Move Data** can report success and update the
torrent's Location while leaving existing torrent data behind in the
old directory.

## Root cause

`set_location_in_session_thread()` searched for a torrent's files using
only `current_dir()` as the source. `current_dir()` is derived from
file #0's location, so if file #0 happens to be absent (e.g. not
selected for download) while a *separate incomplete directory* is
configured, `current_dir()` falls back to `incompleteDir` — even
though the torrent's other files may actually live in `downloadDir`.

As a result, `tr_torrent_files::move()` never found those files, had
nothing to move for them, but still returned success. The torrent's
Location was then updated to the new destination, and the already
downloaded data was silently left behind under the old `downloadDir`.

## Changes

- `tr_torrent_files::move()` now takes a `std::span<std::string_view
  const>` of candidate source directories instead of a single
  `old_parent`, and searches each file across all of them — the same
  set of paths (`downloadDir`, `incompleteDir`) already used elsewhere
  for locating torrent data.
- When the same file exists under more than one source directory, the
  first match is moved and the duplicate is left in place rather than
  overwritten or lost.
- Empty-directory cleanup is skipped for any source directory that
  still contains torrent data after the move, so a duplicate copy left
  behind doesn't get swept into a temp directory.
- `local-data.cc` updated to pass its single source path through the
  new span-based signature.

## Testing

Added three test cases to `move-test.cc`:

- `setLocationFindsFilesOutsideCurrentDir` — file #0 absent,
  `current_dir()` resolves to `incompleteDir`, but the rest of the
  torrent's files live in `downloadDir`; verifies they're now found
  and moved.
- `setLocationMovesFromIncompleteDirToDownloadDir` — destination is
  itself one of the source directories; verifies cleanup doesn't
  remove the just-moved files.
- `setLocationLeavesDuplicateSourceFileInPlace` — the same file exists
  under both `downloadDir` and `incompleteDir`; verifies the unused
  copy is left untouched rather than deleted.

### DIFF
--- a/libtransmission/local-data.cc
+++ b/libtransmission/local-data.cc
@@ -166,7 +166,7 @@ public:
         }
 
         auto error = tr_error{};
-        if (tor->files().move(old_parent, parent, parent_name, &error)) {
+        if (tor->files().move(std::span{ &old_parent, 1U }, parent, parent_name, &error)) {
             return 0;
         }
 
@@ -185,7 +185,7 @@ public:
         }
 
         auto error = tr_error{};
-        tor->files().remove(tor->current_dir().sv(), tor->name(), remove_func, &error);
+        tor->files().remove(tor->search_paths(), tor->name(), remove_func, &error);
         return error ? error.code() : 0;
     }
 
--- a/libtransmission/torrent-files.cc
+++ b/libtransmission/torrent-files.cc
@@ -15,8 +15,10 @@
 #include <string>
 #include <string_view>
 #include <utility>
+#include <vector>
 
 #include <fmt/format.h>
+#include <fmt/ranges.h>
 
 #include <small/map.hpp>
 
@@ -77,6 +79,39 @@ bool is_junk_file(std::string_view filename)
     });
 
     return std::ranges::find(Files, base) != std::ranges::end(Files);
+}
+
+void remove_junk(std::string_view const filename)
+{
+    if (is_empty_folder(filename) || is_junk_file(filename)) {
+        tr_sys_path_remove(filename);
+    }
+}
+
+// Keep configured roots out of cleanup, including on case-insensitive
+// filesystems. A tree inside a root can still be cleaned. Only confirmed
+// absence rules out containment; other errors conservatively prevent cleanup.
+bool tree_contains_path(std::string_view tree, std::string_view path, tr_error& error)
+{
+    auto const resolved_tree = tr_sys_path_resolve(tree, &error);
+    if (std::empty(resolved_tree)) {
+        return !tr_error_is_enoent(error.code());
+    }
+    auto const resolved_path = tr_sys_path_resolve(path, &error);
+    if (std::empty(resolved_path)) {
+        return !tr_error_is_enoent(error.code());
+    }
+
+    for (auto walk = std::string_view{ resolved_path };;) {
+        if (tr_sys_path_is_same(resolved_tree, walk, &error) || error) {
+            return true;
+        }
+        auto const parent = tr_sys_path_dirname(walk);
+        if (parent == walk) {
+            return false;
+        }
+        walk = parent;
+    }
 }
 
 } // unnamed namespace
@@ -142,75 +177,128 @@ std::string_view tr_torrent_files::primary_mime_type() const
 // ---
 
 bool tr_torrent_files::move(
-    std::string_view old_parent_in,
+    std::span<std::string_view const> old_parents,
     std::string_view parent_in,
     std::string_view parent_name,
     tr_error* error) const
 {
-    auto const old_parent = tr_pathbuf{ old_parent_in };
     auto const parent = tr_pathbuf{ parent_in };
-    tr_logAddTrace(fmt::format("Moving files from '{:s}' to '{:s}'", old_parent, parent), parent_name);
+    tr_logAddTrace(fmt::format("Moving files from [{}] to '{:s}'", fmt::join(old_parents, ", "), parent.sv()), parent_name);
 
-    if (tr_sys_path_is_same(old_parent, parent)) {
+    if (std::ranges::all_of(old_parents, [&parent](auto old_parent) { return tr_sys_path_is_same(old_parent, parent); })) {
         return true;
     }
 
-    if (!tr_sys_dir_create(parent, TR_SYS_DIR_CREATE_PARENTS, 0777, error)) {
-        return false;
-    }
+    auto local_error = tr_error{};
+    error = error != nullptr ? error : &local_error;
 
-    auto const paths = std::to_array<std::string_view>({ old_parent.sv() });
-
-    auto err = bool{};
-
+    // Select sources before moving files; retain them to identify cleanup trees.
+    auto moves = std::vector<std::pair<tr_file_index_t, FoundFile>>{};
     for (tr_file_index_t i = 0, n = file_count(); i < n; ++i) {
-        auto const found = find(i, paths);
+        auto const found = find(i, old_parents);
         if (!found) {
             continue;
         }
 
         auto const old_path = found->filename<tr_pathbuf>();
         auto const path = found->filename_under<tr_pathbuf>(parent);
-        tr_logAddTrace(fmt::format("Found file #{:d} '{:s}'", i, old_path), parent_name);
+        tr_logAddTrace(fmt::format("Found file #{:d} '{:s}'", i, old_path.sv()), parent_name);
 
         if (tr_sys_path_is_same(old_path, path)) {
             continue;
         }
 
-        tr_logAddTrace(fmt::format("Moving file #{:d} to '{:s}'", i, old_path, path), parent_name);
+        moves.emplace_back(i, *found);
+    }
+
+    if (!tr_sys_dir_create(parent, TR_SYS_DIR_CREATE_PARENTS, 0777, error)) {
+        return false;
+    }
+
+    for (auto const& [i, found] : moves) {
+        auto const old_path = found.filename<tr_pathbuf>();
+        auto const path = found.filename_under<tr_pathbuf>(parent);
+        if (tr_sys_path_exists(path)) {
+            tr_logAddWarn(
+                fmt::format(
+                    fmt::runtime(_("Moving '{source}' will overwrite '{path}'")),
+                    fmt::arg("source", old_path.sv()),
+                    fmt::arg("path", path.sv())),
+                parent_name);
+        }
+        tr_logAddTrace(fmt::format("Moving file #{:d} from '{:s}' to '{:s}'", i, old_path.sv(), path.sv()), parent_name);
         if (!tr_file_move(old_path, path, true, error)) {
-            err = true;
-            break;
+            error->prefix_message(fmt::format("'{:s}': ", old_path.sv()));
+            return false;
         }
     }
 
-    // after moving the files, remove any leftover empty directories
-    if (!err) {
-        auto const remove_empty_directories = [](std::string_view const path, tr_error* /*err*/) {
-            if (is_empty_folder(path)) {
-                // Since the files have already been moved, errors in this step
-                // are considered secondary and aren't propagated back in `err`.
-                // Log them instead.
-                if (auto log_error = tr_error{}; !tr_sys_path_remove(path, &log_error)) {
-                    tr_logAddWarn(
-                        fmt::format(
-                            fmt::runtime(_("Couldn't remove '{path}': {error} ({error_code})")),
-                            fmt::arg("path", path),
-                            fmt::arg("error", log_error.message()),
-                            fmt::arg("error_code", log_error.code())));
-                }
+    for (auto const old_parent : old_parents) {
+        if (tr_sys_path_is_same(old_parent, parent)) {
+            continue;
+        }
+
+        // Derive each tree from a moved file, not the torrent name. Scan it
+        // once and leave unused trees alone.
+        auto trees = std::set<std::string_view>{};
+        for (auto const& found : moves | std::views::values) {
+            auto const slash = found.subpath.find('/');
+            if (found.base == old_parent && slash != std::string_view::npos) {
+                trees.emplace(found.subpath.substr(0, slash));
             }
-
-            return true;
-        };
-
-        remove(old_parent, parent_name, remove_empty_directories);
+        }
+        for (auto const top : trees) {
+            auto const tree = tr_pathbuf{ old_parent, '/', top };
+            auto const contains_root = [&tree, parent_name](auto root) {
+                auto path_error = tr_error{};
+                if (!tree_contains_path(tree, root, path_error)) {
+                    return false;
+                }
+                tr_logAddDebug(
+                    path_error ? fmt::format("Skipping cleanup of '{:s}': {:s}", tree.sv(), path_error.message()) :
+                                 fmt::format("Skipping cleanup of '{:s}': contains root '{:s}'", tree.sv(), root),
+                    parent_name);
+                return true;
+            };
+            // Do not enter a configured root. A tree inside another root is
+            // allowed; cleanup only removes junk and empty directories.
+            if (contains_root(parent.sv()) ||
+                std::ranges::any_of(old_parents, [&](auto root) { return root != old_parent && contains_root(root); })) {
+                continue;
+            }
+            depth_first_walk(tree, remove_junk);
+        }
     }
 
-    return !err;
+    return true;
 }
 
 // ---
+
+void tr_torrent_files::remove(
+    std::span<std::string_view const> const parents,
+    std::string_view const tmpdir_prefix,
+    tr_torrent_remove_func const& func,
+    tr_error* error) const
+{
+    auto local_error = tr_error{};
+    error = error != nullptr ? error : &local_error;
+
+    for (auto const parent : parents) {
+        // The single-root remove() also cleans junk and empty directories.
+        // Do not run that cleanup in a root with no matching torrent files.
+        if (!has_any_local_data(std::span{ &parent, 1U })) {
+            continue;
+        }
+        auto root_error = tr_error{};
+        remove(parent, tmpdir_prefix, func, &root_error);
+        // Process independent roots even when an earlier root reported an
+        // error, and preserve the first error for the caller.
+        if (root_error && !*error) {
+            *error = std::move(root_error);
+        }
+    }
+}
 
 /**
  * This convoluted code does something (seemingly) simple:
@@ -284,11 +372,6 @@ void tr_torrent_files::remove(
     // OK we've removed the local data.
     // What's left are empty folders, junk, and user-generated files.
     // Remove the first two categories and leave the third alone.
-    auto const remove_junk = [](std::string_view const filename) {
-        if (is_empty_folder(filename) || is_junk_file(filename)) {
-            tr_sys_path_remove(filename);
-        }
-    };
     for (auto const& filename : top_files) {
         depth_first_walk(filename, remove_junk);
     }
--- a/libtransmission/torrent-files.h
+++ b/libtransmission/torrent-files.h
@@ -106,13 +106,21 @@ public:
     }
 
     bool move(
-        std::string_view old_parent_in,
+        std::span<std::string_view const> old_parents,
         std::string_view parent_in,
         std::string_view parent_name = "",
         tr_error* error = nullptr) const;
 
     void remove(
         std::string_view parent_in,
+        std::string_view tmpdir_prefix,
+        tr_torrent_remove_func const& func,
+        tr_error* error = nullptr) const;
+
+    // Remove a matching copy from each root using the single-root deletion
+    // path and its callback. Roots without matching files are left alone.
+    void remove(
+        std::span<std::string_view const> parents,
         std::string_view tmpdir_prefix,
         tr_torrent_remove_func const& func,
         tr_error* error = nullptr) const;
--- a/libtransmission/torrent.cc
+++ b/libtransmission/torrent.cc
@@ -702,7 +702,7 @@ void tr_torrentRemoveInSessionThread(
         }
 
         auto error = tr_error{};
-        tor->files().remove(tor->current_dir().sv(), tor->name(), remove_func, &error);
+        tor->files().remove(tor->search_paths(), tor->name(), remove_func, &error);
         if (error) {
             tr_logAddWarnTor(
                 tor,
@@ -1084,12 +1084,11 @@ void tr_torrent::set_location_in_session_thread(std::string_view const path, boo
         session->verify_remove(this);
 
         auto error = tr_error{};
-        ok = files().move(current_dir().sv(), path, name(), &error);
+        ok = files().move(search_paths(), path, name(), &error);
         if (error) {
             this->error().set_local_error(
                 fmt::format(
-                    fmt::runtime(_("Couldn't move '{old_path}' to '{path}': {error} ({error_code})")),
-                    fmt::arg("old_path", current_dir().sv()),
+                    fmt::runtime(_("Couldn't move torrent files to '{path}': {error} ({error_code})")),
                     fmt::arg("path", path),
                     fmt::arg("error", error.message()),
                     fmt::arg("error_code", error.code())));
@@ -1113,26 +1112,20 @@ void tr_torrent::set_location_in_session_thread(std::string_view const path, boo
     }
 }
 
-namespace
+small::max_size_vector<std::string_view, 2> tr_torrent::search_paths() const
 {
-namespace location_helpers
-{
-size_t buildSearchPathArray(tr_torrent const* tor, std::string_view* paths)
-{
-    auto* walk = paths;
+    auto paths = small::max_size_vector<std::string_view, 2>{};
 
-    if (auto const& path = tor->download_dir(); !std::empty(path)) {
-        *walk++ = path.sv();
+    if (!std::empty(download_dir())) {
+        paths.push_back(download_dir().sv());
     }
 
-    if (auto const& path = tor->incomplete_dir(); !std::empty(path)) {
-        *walk++ = path.sv();
+    if (!std::empty(incomplete_dir()) && incomplete_dir() != download_dir()) {
+        paths.push_back(incomplete_dir().sv());
     }
 
-    return walk - paths;
+    return paths;
 }
-} // namespace location_helpers
-} // namespace
 
 void tr_torrent::set_location(std::string_view location, bool move_from_old_path, int volatile* setme_state)
 {
@@ -1159,20 +1152,12 @@ void tr_torrentSetLocation(
 
 std::optional<tr_torrent_files::FoundFile> tr_torrent::find_file(tr_file_index_t file_index) const
 {
-    using namespace location_helpers;
-
-    auto paths = std::array<std::string_view, 4>{};
-    auto const n_paths = buildSearchPathArray(this, std::data(paths));
-    return files().find(file_index, { paths.data(), n_paths });
+    return files().find(file_index, search_paths());
 }
 
 bool tr_torrent::has_any_local_data() const
 {
-    using namespace location_helpers;
-
-    auto paths = std::array<std::string_view, 4>{};
-    auto const n_paths = buildSearchPathArray(this, std::data(paths));
-    return files().has_any_local_data({ paths.data(), n_paths });
+    return files().has_any_local_data(search_paths());
 }
 
 void tr_torrentSetDownloadDir(tr_torrent* tor, std::string_view const path)
@@ -1802,7 +1787,8 @@ void tr_torrent::recheck_completeness()
             }
             date_done_ = tr_time();
 
-            if (current_dir() == incomplete_dir()) {
+            auto const incomplete = incomplete_dir().sv();
+            if (!std::empty(incomplete) && files().has_any_local_data(std::span{ &incomplete, 1U })) {
                 set_location(download_dir().sv(), true, nullptr);
             }
 
--- a/libtransmission/torrent.h
+++ b/libtransmission/torrent.h
@@ -23,6 +23,8 @@
 
 #include <sigslot/signal.hpp>
 
+#include <small/vector.hpp>
+
 #include "libtransmission/announce-list.h"
 #include "libtransmission/bandwidth.h"
 #include "libtransmission/bitfield.h"
@@ -496,6 +498,9 @@ struct tr_torrent {
     {
         metainfo_.set_file_subpath(i, subpath);
     }
+
+    // The views refer to the torrent's directories, not to the returned container.
+    [[nodiscard]] small::max_size_vector<std::string_view, 2> search_paths() const;
 
     [[nodiscard]] std::optional<tr_torrent_files::FoundFile> find_file(tr_file_index_t file_index) const;
 
--- a/tests/libtransmission/move-test.cc
+++ b/tests/libtransmission/move-test.cc
@@ -4,19 +4,24 @@
 // License text can be found in the licenses/ folder.
 
 #include <algorithm>
+#include <chrono>
+#include <future>
 #include <memory>
 #include <string>
 #include <string_view>
 #include <utility>
+#include <vector>
 
 #include <gtest/gtest.h>
 
 #include <libtransmission/transmission.h>
 
 #include <libtransmission/block-info.h>
+#include <libtransmission/file-utils.h>
 #include <libtransmission/file.h> // tr_sys_path_*()
 #include <libtransmission/local-data.h>
 #include <libtransmission/quark.h>
+#include <libtransmission/session.h>
 #include <libtransmission/torrent-files.h>
 #include <libtransmission/torrent.h>
 #include <libtransmission/tr-strbuf.h>
@@ -51,6 +56,96 @@ protected:
     }
 
     static auto constexpr MaxWaitMsec = 3000;
+
+    int setLocation(tr_torrent* tor, std::string_view path, bool move = true)
+    {
+        auto result = std::make_shared<std::promise<int>>();
+        auto ready = result->get_future();
+        session_->run_in_session_thread([tor, path = std::string{ path }, move, result]() {
+            auto state = int{ TR_LOC_MOVING };
+            tr_torrentSetLocation(tor, path, move, &state);
+            result->set_value(state);
+        });
+        auto const status = ready.wait_for(5s);
+        EXPECT_EQ(std::future_status::ready, status);
+        return status == std::future_status::ready ? ready.get() : -1;
+    }
+
+    void checkRemoveLocalData(bool via_backend)
+    {
+        std::string const download_dir = tr_sessionGetDownloadDir(session_);
+        std::string const incomplete_dir = tr_sessionGetIncompleteDir(session_);
+        auto* const tor = zeroTorrentInit(ZeroTorrentState::NoFiles);
+        ASSERT_NE(nullptr, tor);
+        ASSERT_EQ(incomplete_dir, tor->current_dir().sv());
+        createFileWithContents(
+            tr_pathbuf{ incomplete_dir, '/', tr_torrentFile(tor, 0).name, tr_torrent_files::PartialFileSuffix },
+            "partial"sv);
+        for (tr_file_index_t i = 1; i < tor->file_count(); ++i) {
+            createFileWithContents(tr_pathbuf{ download_dir, '/', tr_torrentFile(tor, i).name }, "data"sv);
+        }
+        auto const download_tree = tr_pathbuf{ download_dir, '/', tor->name() };
+        auto const incomplete_tree = tr_pathbuf{ incomplete_dir, '/', tor->name() };
+
+        auto result = std::make_shared<std::promise<void>>();
+        auto ready = result->get_future();
+        session_->run_in_session_thread([session = session_, tor, via_backend, result]() {
+            if (via_backend) {
+                session->local_data.remove(tor->id(), {});
+            }
+            tr_torrentRemove(tor, !via_backend);
+            result->set_value();
+        });
+        ASSERT_EQ(std::future_status::ready, ready.wait_for(5s));
+
+        EXPECT_FALSE(tr_sys_path_exists(download_tree));
+        EXPECT_FALSE(tr_sys_path_exists(incomplete_tree));
+    }
+
+    void checkSetLocationToNewDirectory(bool partial_first_file)
+    {
+        std::string const download_dir = tr_sessionGetDownloadDir(session_);
+        std::string const incomplete_dir = tr_sessionGetIncompleteDir(session_);
+        auto const target_dir = tr_pathbuf{ session_->configDir(), "/target"sv };
+
+        // File #0 starts absent, so current_dir falls back to incompleteDir.
+        // Optionally add it there as .part; all other files are in downloadDir.
+        auto* const tor = zeroTorrentInit(ZeroTorrentState::NoFiles);
+        ASSERT_NE(nullptr, tor);
+        EXPECT_EQ(incomplete_dir, tor->current_dir().sv());
+        if (partial_first_file) {
+            createFileWithContents(
+                tr_pathbuf{ incomplete_dir, '/', tr_torrentFile(tor, 0).name, tr_torrent_files::PartialFileSuffix },
+                "partial data"sv);
+        }
+        for (tr_file_index_t i = 1; i < tor->file_count(); ++i) {
+            auto const file = tr_torrentFile(tor, i);
+            createFileWithContents(
+                tr_pathbuf{ download_dir, '/', file.name },
+                std::string(static_cast<size_t>(file.length), '\0'));
+        }
+        createFileWithContents(tr_pathbuf{ download_dir, '/', tor->name(), "/.DS_Store"sv }, "junk"sv);
+        if (partial_first_file) {
+            createFileWithContents(tr_pathbuf{ incomplete_dir, '/', tor->name(), "/desktop.ini"sv }, "junk"sv);
+        }
+
+        ASSERT_EQ(TR_LOC_DONE, setLocation(tor, target_dir));
+        if (!partial_first_file) {
+            EXPECT_EQ(""s, tr_torrentFindFile(tor, 0));
+        }
+        for (tr_file_index_t i = partial_first_file ? 0U : 1U; i < tor->file_count(); ++i) {
+            // Only file #0 was created with .part; moving must preserve that name.
+            auto const suffix = i == 0 ? tr_torrent_files::PartialFileSuffix : ""sv;
+            auto const expected = tr_pathbuf{ target_dir, '/', tr_torrentFile(tor, i).name, suffix };
+            EXPECT_EQ(expected, tr_torrentFindFile(tor, i));
+        }
+        EXPECT_FALSE(tr_sys_path_exists(tr_pathbuf{ download_dir, '/', tor->name() }));
+        EXPECT_FALSE(tr_sys_path_exists(tr_pathbuf{ incomplete_dir, '/', tor->name() }));
+        EXPECT_TRUE(tor->incomplete_dir().empty());
+        EXPECT_EQ(target_dir.sv(), tor->download_dir().sv());
+        EXPECT_EQ(target_dir.sv(), tor->current_dir().sv());
+        tr_torrentRemove(tor, true);
+    }
 };
 } // namespace
 
@@ -130,6 +225,80 @@ TEST_P(IncompleteDirTest, incompleteDir)
     }
 
     // cleanup
+    tr_torrentRemove(tor, true);
+}
+
+TEST_P(IncompleteDirTest, removeFindsFilesInBothRoots)
+{
+    checkRemoveLocalData(false);
+}
+
+TEST_P(IncompleteDirTest, backendRemoveFindsFilesInBothRoots)
+{
+    checkRemoveLocalData(true);
+}
+
+TEST_P(IncompleteDirTest, setLocationFindsFilesOutsideCurrentDir)
+{
+    checkSetLocationToNewDirectory(false);
+}
+
+TEST_P(IncompleteDirTest, setLocationMovesFilesFromBothRoots)
+{
+    checkSetLocationToNewDirectory(true);
+}
+
+TEST_P(IncompleteDirTest, setLocationOverwritesExistingDestination)
+{
+    std::string const download_dir = tr_sessionGetDownloadDir(session_);
+    std::string const incomplete_dir = tr_sessionGetIncompleteDir(session_);
+    auto* const tor = zeroTorrentInit(ZeroTorrentState::NoFiles);
+    for (tr_file_index_t i = 1; i < tor->file_count(); ++i) {
+        createFileWithContents(tr_pathbuf{ download_dir, '/', tr_torrentFile(tor, i).name }, "source"sv);
+    }
+    auto const last = tr_torrentFile(tor, tor->file_count() - 1);
+    createFileWithContents(tr_pathbuf{ incomplete_dir, '/', last.name }, "destination"sv);
+
+    ASSERT_EQ(TR_LOC_DONE, setLocation(tor, incomplete_dir));
+    EXPECT_EQ(incomplete_dir, tor->download_dir().sv());
+    EXPECT_TRUE(tor->incomplete_dir().empty());
+    EXPECT_EQ(incomplete_dir, tor->current_dir().sv());
+    for (tr_file_index_t i = 1; i < tor->file_count(); ++i) {
+        auto const name = tr_torrentFile(tor, i).name;
+        EXPECT_FALSE(tr_sys_path_exists(tr_pathbuf{ download_dir, '/', name }));
+        auto contents = std::vector<char>{};
+        ASSERT_TRUE(tr_file_read(tr_pathbuf{ incomplete_dir, '/', name }, contents));
+        EXPECT_EQ("source"sv, (std::string_view{ contents.data(), contents.size() }));
+    }
+    EXPECT_FALSE(tr_sys_path_exists(tr_pathbuf{ download_dir, '/', tor->name() }));
+    tr_torrentRemove(tor, false);
+}
+
+TEST_P(IncompleteDirTest, completionMovesFilesOutsideCurrentDir)
+{
+    std::string const download_dir = tr_sessionGetDownloadDir(session_);
+    std::string const incomplete_dir = tr_sessionGetIncompleteDir(session_);
+    auto* const tor = zeroTorrentInit(ZeroTorrentState::NoFiles);
+    for (tr_file_index_t i = 0; i < tor->file_count(); ++i) {
+        auto const file = tr_torrentFile(tor, i);
+        // The first file makes current_dir point at downloadDir; all the other
+        // files must still be consolidated when verification completes.
+        auto const& base = i == 0 ? download_dir : incomplete_dir;
+        createFileWithContents(tr_pathbuf{ base, '/', file.name }, std::string(static_cast<size_t>(file.length), '\0'));
+    }
+    ASSERT_EQ(TR_LOC_DONE, setLocation(tor, download_dir, false));
+    ASSERT_EQ(download_dir, tor->current_dir().sv());
+    ASSERT_FALSE(tor->is_done());
+    createFileWithContents(tr_pathbuf{ incomplete_dir, '/', tor->name(), "/.DS_Store"sv }, "junk"sv);
+
+    blockingTorrentVerify(tor);
+    EXPECT_TRUE(tor->is_done());
+    EXPECT_TRUE(tor->incomplete_dir().empty());
+    for (tr_file_index_t i = 0; i < tor->file_count(); ++i) {
+        auto const expected = tr_pathbuf{ download_dir, '/', tr_torrentFile(tor, i).name };
+        EXPECT_EQ(expected, tr_torrentFindFile(tor, i));
+    }
+    EXPECT_FALSE(tr_sys_path_exists(tr_pathbuf{ incomplete_dir, '/', tor->name() }));
     tr_torrentRemove(tor, true);
 }
 
--- a/tests/libtransmission/torrent-files-test.cc
+++ b/tests/libtransmission/torrent-files-test.cc
@@ -15,6 +15,8 @@
 
 #include <libtransmission/transmission.h>
 
+#include <libtransmission/error.h>
+#include <libtransmission/file-utils.h>
 #include <libtransmission/file.h>
 #include <libtransmission/torrent-files.h>
 #include <libtransmission/torrent-metainfo.h>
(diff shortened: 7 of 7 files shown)


ckerr:
This should land before #308

libtransmission/torrent-files.cc
```diff
@@ -204,7 +197,18 @@ bool tr_torrent_files::move(
...
+        for (auto i = size_t{}; i < old_parents.size(); ++i) {
+            auto const is_duplicate = std::ranges::find(old_parents.first(i), old_parents[i]) != old_parents.first(i).end();
+            if (is_duplicate || tr_sys_path_is_same(tr_pathbuf{ old_parents[i] }, parent)) {
+                continue;
+            }
+
+            // A matching file may exist under more than one source. Do not let
+            // remove() move an unselected duplicate into a temporary directory.
+            if (!has_any_local_data(old_parents.subspan(i, 1))) {
+                remove(old_parents[i], parent_name, remove_empty_directories);
```
ckerr: (Fable 5.1, extra) Data safety: this now runs `remove()` on every root that holds none of the torrent's files, including roots the torrent never used. `remove()` first creates and deletes `root/<name>__XXXXXX`, then walks `root/<name>` and deletes every empty folder and every junk file (`.DS_Store`, `Thumbs.db`, `desktop.ini`) under it.

Two ways that reaches directories this torrent does not own:
- Every completed torrent gets `incomplete_dir_` re-populated from the session on restart (`init()` at torrent.cc:900-902; the resume file never clears it), so Move Data to a third directory prunes `incomplete_dir/<name>`. If another torrent with the same name is downloading there, or the user keeps a folder by that name, its empty subfolders and junk files are deleted.
- When `<name>` equals a path component, `root/<name>` *is* the other root or the destination. download_dir `/dl`, incomplete_dir `/dl/Incomplete`, torrent named `Incomplete`: the junk walk covers the whole incomplete dir and removes it if empty. Torrent `Show` in `/inc` moved to `/dl/Show`: the walk prunes the destination.

The base only ever cleaned `current_dir()`. Fix: record `found->base` for each file actually moved in the loop above and call `remove()` only for those roots, keeping the `has_any_local_data` guard (a root can both contribute a moved file and still hold a duplicate of another file, so the guard stays load-bearing). None of the new tests would catch either case: every test torrent is named `files-filled-with-zeroes` and no test puts a same-named foreign folder in the unused root.

libtransmission/torrent-files.cc
```diff
@@ -204,7 +197,18 @@ bool tr_torrent_files::move(
...
-        remove(old_parent, parent_name, remove_empty_directories);
+        for (auto i = size_t{}; i < old_parents.size(); ++i) {
+            auto const is_duplicate = std::ranges::find(old_parents.first(i), old_parents[i]) != old_parents.first(i).end();
+            if (is_duplicate || tr_sys_path_is_same(tr_pathbuf{ old_parents[i] }, parent)) {
+                continue;
+            }
+
+            // A matching file may exist under more than one source. Do not let
+            // remove() move an unselected duplicate into a temporary directory.
+            if (!has_any_local_data(old_parents.subspan(i, 1))) {
```
ckerr: (Fable 5.1, extra) With this guard, `remove()` can never move anything: it runs only when `find(i, {root})` misses for every file, and `remove()`'s own move-to-tmpdir loop uses the same `find` on the same root. So the `remove_empty_directories` callback above is unreachable, the tmpdir is always empty, and every call still pays `mkdtemp` + `rmdir` plus a second full `find` pass over all files right after this guard did one (two stats per file per root each time).

All the work that actually happens is `remove()`'s final phase, `depth_first_walk(root/<name>, remove_junk)`. Hoisting `remove_junk` out of `remove()` into the unnamed namespace and calling `depth_first_walk(tr_pathbuf{ old_parents[i], '/', parent_name }, remove_junk)` here does the same job without the callback, the tmpdir, the second stat pass, and the "one duplicate anywhere disables pruning of the whole root" effect. The comment at line 180 ("remove any leftover empty directories") would then also match what the code does, since junk files are removed too.

libtransmission/torrent-files.cc
```diff
@@ -142,29 +142,22 @@ std::string_view tr_torrent_files::primary_mime_type() const
...
         return false;
     }
 
-    auto const paths = std::to_array<std::string_view>({ old_parent.sv() });
-
     auto err = bool{};
 
     for (tr_file_index_t i = 0, n = file_count(); i < n; ++i) {
-        auto const found = find(i, paths);
+        auto const found = find(i, old_parents);
```
ckerr: (Fable 5.1, extra) Data loss in one layout: the same file under both roots and the destination is the incomplete dir. `find` returns the download_dir copy (root 0 first), `path` is `I/name/X`, `tr_sys_path_is_same` is false because they are two distinct files, and `tr_file_move` calls `rename(2)` (`MOVEFILE_REPLACE_EXISTING` on Windows), which replaces the incomplete-dir copy. Cleanup then skips root I as the destination. The base returned early when `current_dir()` was the destination, so it never overwrote in that state.

The mirror case (destination = download_dir) is safe because `is_same` short-circuits, so behaviour is asymmetric and contradicts the PR's "left in place rather than overwritten" guarantee. The overwritten copy is the one the torrent was not reading, which limits the damage, but the RPC, GTK and Qt clients all allow this destination. Searching any root that is the destination first (so the existing destination copy wins via `is_same` in both directions), or skipping the move when `find(i, { &parent_sv, 1 })` already hits, closes it. The duplicate test only uses a third directory; a variant whose destination is `incomplete_dir` would cover this.

libtransmission/torrent-files.cc
```diff
@@ -204,7 +197,18 @@ bool tr_torrent_files::move(
...
         };
 
-        remove(old_parent, parent_name, remove_empty_directories);
+        for (auto i = size_t{}; i < old_parents.size(); ++i) {
+            auto const is_duplicate = std::ranges::find(old_parents.first(i), old_parents[i]) != old_parents.first(i).end();
+            if (is_duplicate || tr_sys_path_is_same(tr_pathbuf{ old_parents[i] }, parent)) {
+                continue;
+            }
+
+            // A matching file may exist under more than one source. Do not let
```
ckerr: (Fable 5.1, extra) The copy this leaves behind is never reported and, after the caller's `incomplete_dir_.clear()` (torrent.cc:1129), never reachable again: `find_file`, delete-local-data and the next Move Data all build their roots from `download_dir_`/`incomplete_dir_`, neither of which names that directory any more. Move Data reports success and the user has no way to learn a copy was skipped or where.

Preferring the download_dir copy is right, since it is the one `find_file` already serves to I/O, but the choice should surface: `tr_logAddWarn` (or Info) per skipped copy with its path. With the per-file prune from the comment below, iterating the `find(i, {old_parents[i]})` hits gives the list to log instead of collapsing it into one boolean.

libtransmission/torrent-files.cc
```diff
@@ -204,7 +197,18 @@ bool tr_torrent_files::move(
             return true;
         };
 
-        remove(old_parent, parent_name, remove_empty_directories);
+        for (auto i = size_t{}; i < old_parents.size(); ++i) {
+            auto const is_duplicate = std::ranges::find(old_parents.first(i), old_parents[i]) != old_parents.first(i).end();
+            if (is_duplicate || tr_sys_path_is_same(tr_pathbuf{ old_parents[i] }, parent)) {
```
ckerr: (Fable 5.1, extra) Two simplifications:
- `is_duplicate` only catches identically spelled roots. `remove()` on an already-cleaned root is a no-op (creates and removes an empty tmpdir, walks `root/<name>`, finds nothing), so the scan buys nothing. If dedupe is wanted, the producer is the place: `buildSearchPathArray` never compares `incomplete_dir()` to `download_dir()`, and a dedupe there also stops `find_file`/`has_any_local_data` from stat'ing every file twice when a user sets both dirs to the same path.
- `tr_sys_path_is_same` takes `std::string_view`, so the `tr_pathbuf{}` copy is a 4 KiB stack buffer for nothing; line 209 already passes `old_parents[i]` directly.
```suggestion
            if (tr_sys_path_is_same(old_parents[i], parent)) {
```

libtransmission/torrent-files.cc
```diff
@@ -142,29 +142,22 @@ std::string_view tr_torrent_files::primary_mime_type() const
...
-    auto const old_parent = tr_pathbuf{ old_parent_in };
     auto const parent = tr_pathbuf{ parent_in };
-    tr_logAddTrace(fmt::format("Moving files from '{:s}' to '{:s}'", old_parent, parent), parent_name);
-
-    if (tr_sys_path_is_same(old_parent, parent)) {
-        return true;
-    }
+    tr_logAddTrace(fmt::format("Moving files to '{:s}'", parent), parent_name);
 
     if (!tr_sys_dir_create(parent, TR_SYS_DIR_CREATE_PARENTS, 0777, error)) {
```
ckerr: (Fable 5.1, extra) The deleted early return (`if (tr_sys_path_is_same(old_parent, parent)) return true;`) made a move to the current directory cost three syscalls. Now the same request (RPC `torrent-set-location` with `move=true` to the torrent's own download dir, no incomplete dir) runs `tr_sys_dir_create` and, per file, `find` plus `tr_sys_path_is_same` on two identical existing paths: about seven syscalls per file on the session thread to conclude nothing moves. Computing `tr_sys_path_is_same(root, parent)` once per root before the loop (the cleanup loop needs the same answers at line 202) restores the early-out when every root matches, and lets the loop `continue` on `found->base` for a matching root without the per-file `is_same`.

libtransmission/torrent-files.cc
```diff
@@ -142,29 +142,22 @@ std::string_view tr_torrent_files::primary_mime_type() const
...
     tr_error* error) const
 {
-    auto const old_parent = tr_pathbuf{ old_parent_in };
     auto const parent = tr_pathbuf{ parent_in };
-    tr_logAddTrace(fmt::format("Moving files from '{:s}' to '{:s}'", old_parent, parent), parent_name);
-
-    if (tr_sys_path_is_same(old_parent, parent)) {
-        return true;
-    }
+    tr_logAddTrace(fmt::format("Moving files to '{:s}'", parent), parent_name);
```
ckerr: (Fable 5.1, extra) The per-file trace at line 173 has two placeholders and three arguments: `fmt::format("Moving file #{:d} to '{:s}'", i, old_path, path)`. fmt drops the unused `path`, so the log labels the source as the destination. Pre-existing, but this PR removed the only line that named the source root (`Moving files from ... to ...`), so a trace of a split-directory move now shows no destination for any file and no source root anywhere. `"Moving file #{:d} from '{:s}' to '{:s}'"` fixes the former; listing `old_parents` here restores the latter.

libtransmission/torrent.cc
```diff
@@ -1083,13 +1104,15 @@ void tr_torrent::set_location_in_session_thread(std::string_view const path, boo
...
+
         auto error = tr_error{};
-        ok = files().move(current_dir().sv(), path, name(), &error);
+        ok = files().move({ source_paths.data(), n_source_paths }, path, name(), &error);
         if (error) {
             this->error().set_local_error(
                 fmt::format(
-                    fmt::runtime(_("Couldn't move '{old_path}' to '{path}': {error} ({error_code})")),
-                    fmt::arg("old_path", current_dir().sv()),
+                    fmt::runtime(_("Couldn't move torrent files to '{path}': {error} ({error_code})")),
```
ckerr: (Fable 5.1, extra) The user-facing error now names neither the source directory nor the file. `tr_file_move`'s messages ("Unable to get information on old file: ...", "Unable to copy: ...") carry no path, so a failure on one file of thousands renders as `Couldn't move torrent files to '/target': Unable to copy: Permission denied (13)` with two possible source roots and nothing to inspect. The base at least named the source dir. The only place that knows `old_path` is `tr_torrent_files::move` at the `tr_file_move` call; an `error->prefix_message(fmt::format("'{}': ", old_path))` before its `break` makes this message actionable without adding a placeholder here. (New msgid either way, so translators need the final string.)

libtransmission/torrent.cc
```diff
@@ -1083,13 +1104,15 @@ void tr_torrent::set_location_in_session_thread(std::string_view const path, boo
         session->close_torrent_files(id());
         session->verify_remove(this);
 
+        auto source_paths = std::array<std::string_view, 4>{};
```
ckerr: (Fable 5.1, extra) This is the third copy of `std::array<std::string_view, 4>{}` + `buildSearchPathArray` + `{ .data(), n }` (also `find_file` at 1166 and `has_any_local_data` at 1175), and the reason the whole `location_helpers` block had to move up. A `tr_torrent::search_paths()` accessor returning a fixed-capacity container (`small::max_size_vector<std::string_view, 2>` is already used in session-settings) makes each call site one line and gives the root set a name.

The name matters because the root cause the PR describes (current_dir() derives from file #0) still drives two sibling consumers this PR does not touch:
- delete-local-data: `tor->files().remove(tor->current_dir().sv(), ...)` at torrent.cc:705 and local-data.cc:190. In exactly the state `setLocationFindsFilesOutsideCurrentDir` builds, "Remove and delete local data" deletes nothing and leaves files 1..n under download_dir.
- completion consolidation: `if (current_dir() == incomplete_dir())` at torrent.cc:1807. With file #0 in download_dir and the rest downloading into incomplete_dir, the finished data stays in incomplete_dir; the new `move()` would handle it if the gate were `files().has_any_local_data({ &incomplete_dir_sv, 1 })`.

Neither has a test.

libtransmission/local-data.cc
```diff
@@ -165,8 +166,9 @@ class DefaultBackend final : public LocalData::Backend
             return TR_ERROR_EINVAL;
         }
 
+        auto const old_parents = std::array<std::string_view, 1>{ old_parent };
```
ckerr: (Fable 5.1, extra) This keeps the single-source contract alive at the backend boundary: `Backend::move` takes a caller-supplied `old_parent` while its sibling `remove` (line 190) derives the root from the torrent. Nothing in production calls `LocalData::move` today (only local-data-test.cc via a stub), so the first caller that routes `set_location` through this layer will pass `current_dir()` and reintroduce the bug this PR fixes. Dropping `old_parent` from `Backend::move`/`LocalData::move` and using the torrent's roots here makes the interface unable to express the wrong thing. If the parameter stays, `std::span{ &old_parent, 1U }` is the idiom used elsewhere (torrent.h:441, crypto-utils.cc:173) and needs no `<array>` include.

tests/libtransmission/move-test.cc
```diff
@@ -133,6 +133,93 @@ TEST_P(IncompleteDirTest, incompleteDir)
     tr_torrentRemove(tor, true);
 }
 
+TEST_P(IncompleteDirTest, setLocationFindsFilesOutsideCurrentDir)
```
ckerr: (Fable 5.1, extra) The headline case is not exercised: file 0 in `incomplete_dir/<name>/1048576.part`, files 1-2 in `download_dir/<name>/`, destination a third directory. This test populates only download_dir, the next only incomplete_dir with download_dir as the destination, and the third is the duplicate case. A regression that moves only the `current_dir()` root, or cleans only one root, passes all three.

None of the three asserts the source side either: that `<source>/<name>` is gone after the move, that no `<name>__XXXXXX` temp dir is left behind, or that `incomplete_dir()` is empty and `current_dir()` equals the target afterwards. The `has_any_local_data` guard at torrent-files.cc:208 could be inverted, or `remove()` skipped entirely, and every new test still passes. `EXPECT_FALSE(tr_sys_path_exists(tr_pathbuf{ download_dir, '/', tor->name() }))` after the move here and in the duplicate test, plus the mirror check on `incomplete_dir` in the middle test, closes that.

tests/libtransmission/move-test.cc
```diff
@@ -133,6 +133,93 @@ TEST_P(IncompleteDirTest, incompleteDir)
...
+    auto state = -1;
+    tr_torrentSetLocation(tor, target_dir, true, &state);
+    EXPECT_TRUE(waitFor([&state]() { return state == TR_LOC_DONE; }, MaxWaitMsec));
+    EXPECT_EQ(TR_LOC_DONE, state);
+
+    auto const expected = tr_pathbuf{ target_dir, '/', file.name };
+    EXPECT_EQ(expected, tr_torrentFindFile(tor, 1));
+    EXPECT_TRUE(tr_sys_path_exists(incomplete_file));
+
+    tr_sys_path_remove(incomplete_file);
```
ckerr: (Fable 5.1, extra) `Sandbox::~Sandbox` rimrafs the whole sandbox at teardown (test-fixtures.h), and the incomplete dir lives under it, so this line does nothing the fixture does not already do; no other test in the file cleans up by hand.
```suggestion
```

tests/libtransmission/move-test.cc
```diff
@@ -133,6 +133,93 @@ TEST_P(IncompleteDirTest, incompleteDir)
...
+    for (tr_file_index_t i = 1; i < n; ++i) {
+        auto const file = tr_torrentFile(tor, i);
+        auto const filename = tr_pathbuf{ download_dir, '/', file.name };
+        createFileWithContents(filename, std::string(static_cast<size_t>(file.length), '\0'));
+    }
+
+    auto state = -1;
+    tr_torrentSetLocation(tor, target_dir, true, &state);
+    EXPECT_TRUE(waitFor([&state]() { return state == TR_LOC_DONE; }, MaxWaitMsec));
+    EXPECT_EQ(TR_LOC_DONE, state);
```
ckerr: (Fable 5.1, extra) Nit: this four-line block appears three times in the new tests (and a variant in `MoveTest.setLocation`). A fixture helper that waits on `state != TR_LOC_MOVING` and returns `state` lets each test say `EXPECT_EQ(TR_LOC_DONE, setLocationAndWait(tor, target_dir))`, and reports a timeout and a `TR_LOC_ERROR` through the same assertion.

libtransmission/torrent-files.cc
```diff
@@ -142,29 +142,22 @@ std::string_view tr_torrent_files::primary_mime_type() const
...
 
     if (!tr_sys_dir_create(parent, TR_SYS_DIR_CREATE_PARENTS, 0777, error)) {
         return false;
     }
```
ckerr: (Fable 5.1, extra) Keep the same-directory early-out. Without it, a move to the current directory stats every file (`find` plus `tr_sys_path_is_same`) to conclude nothing moves. The RPC and GTK clients allow that request.
```suggestion
    auto const is_parent = [&parent](auto const old_parent) {
        return tr_sys_path_is_same(old_parent, parent);
    };
    if (std::ranges::all_of(old_parents, is_parent)) {
        return true;
    }

    if (!tr_sys_dir_create(parent, TR_SYS_DIR_CREATE_PARENTS, 0777, error)) {
        return false;
    }
```

libtransmission/torrent-files.cc
ckerr: (Fable 5.1, extra) Two trace fixes:
- This line dropped the source roots. Please log them.
- Line 173 has two placeholders and three arguments, so it prints the source as the destination. Use `"Moving file #{:d} from '{:s}' to '{:s}'"`.

libtransmission/torrent-files.cc
ckerr, line 198: (Fable 5.1, extra) Data loss when the destination is the incomplete dir and the file exists under both roots. `find` returns the download_dir copy. `tr_file_move` renames it over `I/name/X`, destroying that copy. The mirror case (destination = download_dir) skips via `is_same`, so only this direction overwrites. The base returned early here.

Please warn before overwriting, say so in the PR description ("left in place" is not true here), and add a test whose destination is `incomplete_dir`.

libtransmission/torrent-files.cc
```diff
@@ -204,7 +197,18 @@ bool tr_torrent_files::move(
...
+            auto const is_duplicate = std::ranges::find(old_parents.first(i), old_parents[i]) != old_parents.first(i).end();
+            if (is_duplicate || tr_sys_path_is_same(tr_pathbuf{ old_parents[i] }, parent)) {
+                continue;
+            }
+
+            // A matching file may exist under more than one source. Do not let
+            // remove() move an unselected duplicate into a temporary directory.
+            if (!has_any_local_data(old_parents.subspan(i, 1))) {
+                remove(old_parents[i], parent_name, remove_empty_directories);
+            }
```
ckerr: (Fable 5.1, extra) Simplify the loop and log what it leaves behind:
- `is_duplicate` only matches identical spellings, and a second `remove()` on a cleaned root is a no-op. Drop it. Dedupe in `buildSearchPathArray` if you want it.
- `tr_sys_path_is_same` takes `string_view`; the `tr_pathbuf` copy is unneeded.
- A copy left under a root is silent today. After `incomplete_dir_.clear()` the torrent can never find it again. Warn with the path.
```suggestion
        for (auto const old_parent : old_parents) {
            if (tr_sys_path_is_same(old_parent, parent)) {
                continue;
            }

            // A file may exist under more than one source. Leave the extra
            // copy in place; remove() would sweep it into a temp directory.
            if (has_any_local_data(std::span{ &old_parent, 1U })) {
                tr_logAddWarn(
                    fmt::format(
                        fmt::runtime(_("Left torrent data behind in '{path}'")),
                        fmt::arg("path", old_parent)),
                    parent_name);
                continue;
            }

            remove(old_parent, parent_name, remove_empty_directories);
        }
```

libtransmission/torrent-files.cc
ckerr: (Fable 5.1, extra) Data safety: `remove()` now runs on roots this torrent never used. It creates `root/<name>__XXXXXX`, then deletes every empty folder and junk file under `root/<name>`. Every completed torrent gets `incomplete_dir_` re-set from the session on restart (torrent.cc:900-902), so Move Data prunes `incomplete_dir/<name>` even when this torrent was never there. That folder can belong to another torrent with the same name. When `<name>` is a path component, `root/<name>` is the other root or the destination (download_dir `/dl`, incomplete_dir `/dl/Incomplete`, torrent named `Incomplete`).

Please clean only roots a file was moved from: record `found->base` after each successful `tr_file_move` and check it here. Keep the `has_any_local_data` guard. Add a test with a same-named foreign folder in the unused root.

libtransmission/torrent-files.cc
ckerr: (Fable 5.1, extra) `remove_empty_directories` can never run. This guard means `find` misses every file under the root, and `remove()` uses the same `find`, so its temp dir is always empty and the callback is never called. What remains is a second stat pass over every file, `mkdtemp` + `rmdir`, and the junk walk of `root/<name>`.

Please call the junk walk directly: hoist `remove_junk` out of `remove()` and use `depth_first_walk(tr_pathbuf{ old_parents[i], '/', parent_name }, remove_junk)`. Then delete the callback above.

libtransmission/torrent.cc
ckerr: (Fable 5.1, extra) Third copy of this idiom (`find_file` 1166, `has_any_local_data` 1175), and the reason the helper block moved. Please add a `tr_torrent::search_paths()` accessor returning `small::max_size_vector<std::string_view, 2>` and use it in all three.

Two consumers still use `current_dir()` alone and have the bug this PR fixes:
- Delete-local-data, torrent.cc:705 and local-data.cc:190. In the state `setLocationFindsFilesOutsideCurrentDir` builds, it deletes nothing. Please pass the same roots.
- Completion consolidation, torrent.cc:1807. With file #0 in download_dir and the rest in incomplete_dir, the finished data is never moved. Please gate on data actually being in incomplete_dir:
```cpp
            auto const incomplete = incomplete_dir().sv();
            if (!std::empty(incomplete) && files().has_any_local_data(std::span{ &incomplete, 1U })) {
                set_location(download_dir().sv(), true, nullptr);
            }
```

libtransmission/torrent.cc
ckerr, line 1091: (Fable 5.1, extra) This message no longer names the source or the file. `tr_file_move`'s errors carry no path either, so a failure reads `Couldn't move torrent files to '/target': Unable to copy: Permission denied (13)`. Please prefix the error with the file in `move()`:
```cpp
        if (!tr_file_move(old_path, path, true, error)) {
            if (error != nullptr) {
                error->prefix_message(fmt::format("'{:s}': ", old_path));
            }
            err = true;
            break;
        }
```

libtransmission/local-data.cc
```diff
@@ -165,8 +166,9 @@ class DefaultBackend final : public LocalData::Backend
...
         auto error = tr_error{};
-        if (tor->files().move(old_parent, parent, parent_name, &error)) {
+        if (tor->files().move(old_parents, parent, parent_name, &error)) {
```
ckerr: (Fable 5.1, extra) `Backend::move` still takes one source root; `remove` below it (line 190) derives the root from the torrent. Please drop `old_parent` and use the torrent's roots here, so no future caller can pass `current_dir()` again. Only `local-data-test.cc` calls this.

If the parameter stays, use the existing one-element span idiom (torrent.h:441) and drop the new `<array>` include:
```suggestion
        auto error = tr_error{};
        if (tor->files().move(std::span{ &old_parent, 1U }, parent, parent_name, &error)) {
```

tests/libtransmission/move-test.cc
ckerr, line 241: (Fable 5.1, extra) Please add the headline case: file 0 partial in `incomplete_dir`, files 1-2 in `download_dir`, destination a third directory. Test 1 uses only download_dir, test 2 only incomplete_dir, test 3 the duplicate. A move that handles one root passes all three.

tests/libtransmission/move-test.cc
```diff
@@ -133,6 +133,93 @@ TEST_P(IncompleteDirTest, incompleteDir)
...
+
+    // The absent file remains absent, while every existing file is moved from
+    // downloadDir even though it was not the torrent-wide current_dir.
+    EXPECT_EQ(""s, tr_torrentFindFile(tor, 0));
+    for (tr_file_index_t i = 1; i < n; ++i) {
+        auto const expected = tr_pathbuf{ target_dir, '/', tr_torrentFile(tor, i).name };
+        EXPECT_EQ(expected, tr_torrentFindFile(tor, i));
+    }
+
+    tr_torrentRemove(tor, true);
```
ckerr: (Fable 5.1, extra) Assert the source side too. Nothing checks that `download_dir/<name>` is pruned or that the torrent's dirs are updated, so an inverted guard at torrent-files.cc:208 passes.
```suggestion
    EXPECT_FALSE(tr_sys_path_exists(tr_pathbuf{ download_dir, '/', tr_torrentName(tor) }));
    EXPECT_TRUE(tor->incomplete_dir().empty());
    EXPECT_EQ(target_dir.sv(), tor->current_dir().sv());

    tr_torrentRemove(tor, true);
```

tests/libtransmission/move-test.cc
ckerr: (Fable 5.1, extra) The sandbox is removed at teardown, so this cleanup is unneeded. Replace it with the check that the used root was pruned:
```suggestion
    EXPECT_FALSE(tr_sys_path_exists(tr_pathbuf{ download_dir, '/', tr_torrentName(tor) }));
```

ckerr (changes requested):
Reviewed by Fable 5.1 extra with prompt:

> /code-review PR 301. Add any review comments and/or code suggestions as drafts in a GitHub PR review. Do not submit. Preface each comment/suggestions with "(Fable 5.1, extra)". Please make the comments actionable, short sentences. If you have specific code suggestions, show rather than tell: add them as code suggestions.

h1de0x:
@ckerr I've updated the patch based on your review. A closer look revealed several complications, but I think this version strikes a reasonable balance between code complexity and performance.

* Added `search_paths()` and updated completion consolidation to check for data in `incompleteDir`.
* Restored the same-directory early return and corrected the trace messages and error paths.
* Replaced `remove()` with direct cleanup of trees derived from moved files, protecting configured roots and the destination without following directory symlinks.
* Kept the existing overwrite behavior and added a warning immediately before overwriting an existing destination.
* Updated the tests to cover both source roots, source cleanup, directory-state updates, and overwriting a copy in `incompleteDir`.

Cleanup uses the existing junk-name policy. Ordinary remaining files are preserved.

I left out the warning about copies remaining behind. The revised cleanup no longer needs a scan for remaining torrent files, so adding this warning would require an additional pass with filesystem lookups solely for diagnostics. I'd prefer to avoid that additional filesystem overhead, especially for torrents with many files, but I'm happy to add the warning if you consider it necessary.

I also left deletion from both roots outside this fix, since it expands the scope of deletion and needs separate consideration. For `local-data.cc`, I used your suggested one-element span alternative.

libtransmission/torrent-files.cc
```diff
@@ -142,72 +196,100 @@ std::string_view tr_torrent_files::primary_mime_type() const
...
+    tr_logAddTrace(fmt::format("Moving files from [{}] to '{:s}'", fmt::join(old_parents, ", "), parent.sv()), parent_name);
 
-    if (tr_sys_path_is_same(old_parent, parent)) {
+    if (std::ranges::all_of(old_parents, [&parent](auto old_parent) { return tr_sys_path_is_same(old_parent, parent); })) {
         return true;
     }
 
-    if (!tr_sys_dir_create(parent, TR_SYS_DIR_CREATE_PARENTS, 0777, error)) {
-        return false;
-    }
```
tearfur: I don't think this should be removed. Was this on purpose?

libtransmission/torrent-files.cc
```diff
@@ -79,6 +81,58 @@ bool is_junk_file(std::string_view filename)
...
+    if (std::empty(resolved_path)) {
+        return !tr_error_is_enoent(error.code());
+    }
+
+    for (auto walk = std::string_view{ resolved_path };;) {
+        if (tr_sys_path_is_same(resolved_tree, walk, &error) || error) {
+            return true;
+        }
+        auto const parent = tr_sys_path_dirname(walk);
+        if (parent == walk || std::empty(parent)) {
```
tearfur: What is `std::empty(parent)` meant to catch? `tr_sys_path_dirname` never returns an empty string AFAICT.

libtransmission/torrent-files.cc
```diff
@@ -79,6 +81,58 @@ bool is_junk_file(std::string_view filename)
...
+// Do not follow symlinks or descend into entries with unknown types.
+// Attempt removal of junk-named entries even if metadata cannot be read,
+// including files listed in the torrent. Directory removal checks emptiness.
+void cleanup_moved_tree(std::string_view path, std::string_view log_name)
+{
+    auto error = tr_error{};
+    auto const info = tr_sys_path_get_info(path, TR_SYS_PATH_NO_FOLLOW, &error);
+    if (info && info->isFolder()) {
+        for (auto const& name : tr_sys_dir_get_files(path, tr_basename_accept_all, &error)) {
+            cleanup_moved_tree(tr_pathbuf{ path, '/', name }, log_name);
```
tearfur: This is essentially `depth_first_walk` that does not follow symlinks, deleting is baked in, and has a few extra logs.

Is there really a need for new code here? Why shouldn't we delete junk files and empty directories within symlinked directory trees, just in the move path?

libtransmission/torrent.cc
```diff
@@ -1083,13 +1083,13 @@ void tr_torrent::set_location_in_session_thread(std::string_view const path, boo
         session->close_torrent_files(id());
         session->verify_remove(this);
 
+        auto const paths = search_paths();
         auto error = tr_error{};
-        ok = files().move(current_dir().sv(), path, name(), &error);
+        ok = files().move(std::span{ paths.data(), paths.size() }, path, name(), &error);
```
tearfur: *(nit)*

```suggestion
        ok = files().move(search_paths(), path, name(), &error);
```

libtransmission/torrent.cc
```diff
@@ -1159,20 +1153,14 @@ void tr_torrentSetLocation(
 
 std::optional<tr_torrent_files::FoundFile> tr_torrent::find_file(tr_file_index_t file_index) const
 {
-    using namespace location_helpers;
-
-    auto paths = std::array<std::string_view, 4>{};
-    auto const n_paths = buildSearchPathArray(this, std::data(paths));
-    return files().find(file_index, { paths.data(), n_paths });
+    auto const paths = search_paths();
+    return files().find(file_index, std::span{ paths.data(), paths.size() });
```
tearfur:
```suggestion
    return files().find(file_index, search_paths());
```

libtransmission/torrent.cc
```diff
@@ -1159,20 +1153,14 @@ void tr_torrentSetLocation(
...
 
 bool tr_torrent::has_any_local_data() const
 {
-    using namespace location_helpers;
-
-    auto paths = std::array<std::string_view, 4>{};
-    auto const n_paths = buildSearchPathArray(this, std::data(paths));
-    return files().has_any_local_data({ paths.data(), n_paths });
+    auto const paths = search_paths();
+    return files().has_any_local_data(std::span{ paths.data(), paths.size() });
```
tearfur:
```suggestion
    return files().has_any_local_data(search_paths());
```

libtransmission/torrent-files.cc
```diff
@@ -142,72 +196,100 @@ std::string_view tr_torrent_files::primary_mime_type() const
...
+        auto trees = std::set<std::string_view>{};
+        for (auto const& found : moves | std::views::values) {
+            auto const slash = found.subpath.find('/');
+            if (found.base == old_parent && slash != std::string_view::npos) {
+                trees.emplace(found.subpath.substr(0, slash));
+            }
+        }
+        for (auto const top : trees) {
+            auto const tree = tr_pathbuf{ old_parent, '/', top };
+            auto const contains_root = [&tree, parent_name](auto root) {
```
tearfur, line 252: What sort of situation is `contains_root` covering?

libtransmission/torrent.cc
```diff
@@ -1083,13 +1083,13 @@ void tr_torrent::set_location_in_session_thread(std::string_view const path, boo
         session->close_torrent_files(id());
         session->verify_remove(this);
 
+        auto const paths = search_paths();
```
ckerr: (Fable 5.1, extra) Delete-local-data still uses `current_dir()` alone (torrent.cc:705, local-data.cc:188). Pass `search_paths()` there too and add a test.

h1de0x:
@tearfur  @ckerr  thanks for the detailed review. I have updated the patch. Could you please check?

The main changes addressing the review are:

* Post-move cleanup now reuses the existing `depth_first_walk()` and `remove_junk()` implementation instead of maintaining a separate traversal. This also restores the pre-PR symlink behavior.
* The redundant empty-result check after `tr_sys_path_dirname()` has been removed.
* `search_paths()` is now passed directly to `move()`, `find()`, and `has_any_local_data()`.
* I kept the containment guard because a cleanup tree derived from a moved file may encompass the destination or another configured data root. In that case, cleanup must not enter or remove the protected directory. A cleanup tree located inside another root remains eligible for cleanup.
* Local-data removal now processes both configured search roots in both removal entry points. Roots with no matching torrent data are skipped so that unrelated same-named directories are not cleaned. All populated roots are attempted, while the first reported error is preserved.

The tests were reduced to the essential regression coverage, including moving and removing data split across both roots and preserving an unused root.

tearfur:
Hey @h1de0x, thanks for the updates.

The questions I left in the review aren't rhetorical, they are genuine questions and so I'd like you to tell me your reasoning. Particularly the 2 questions about `depth_first_walk` and `contains_root`.

I see you re-used `depth_first_walk` now, but I still want to understand what made you feel the need to write new code in the first place.

libtransmission/torrent-files.cc
```diff
@@ -142,76 +177,129 @@ std::string_view tr_torrent_files::primary_mime_type() const
...
 
-        tr_logAddTrace(fmt::format("Moving file #{:d} to '{:s}'", i, old_path, path), parent_name);
+        moves.emplace_back(i, *found);
+    }
+
+    if (!tr_sys_dir_create(parent, TR_SYS_DIR_CREATE_PARENTS, 0777, error)) {
+        return false;
+    }
+
+    for (auto const& [i, found] : moves) {
```
tearfur: Let's subsume the body of this loop below `tr_sys_path_is_same`. Should shave off a little overhead.

h1de0x:
@tearfur you're right, my previous reply described the changes without properly answering your questions.

For `depth_first_walk`, my initial concern was that cleanup would now cover trees under both source roots. Following a directory symlink could take that cleanup outside the source tree, so I introduced a separate traversal that did not follow symlinks. However, that also introduced a move-specific cleanup policy and duplicated existing code. I reconsidered that scope and restored the existing traversal: changing symlink handling is not necessary to fix the source-directory lookup bug.

For `contains_root`, the concern is independent of symlinks. For example, suppose the source root is `/data`, a moved file is `name/file`, and the destination is `/data/name/new-location`. The resulting cleanup tree is `/data/name`, which contains the destination. An unrestricted junk walk would enter the destination and could delete junk files there. The same problem occurs if another configured root, such as `incompleteDir`, is `/data/name/Incomplete`.

The guard skips the cleanup tree when it contains or equals the destination or another configured root. It does not skip a tree merely because it is inside a configured root, so ordinary nested download/incomplete layouts still get cleaned. `cleanupPreservesNestedRootAndDestination` covers both protected-directory cases.

tests/libtransmission/torrent-files-test.cc
```diff
@@ -27,6 +29,109 @@ using namespace std::literals;
 
 using TorrentFilesTest = ::tr::test::SandboxedTest;
 
+namespace
+{
+class TorrentFilesMoveTest : public ::tr::test::SandboxedTest
```
tearfur: *(nit)* Let's subsume `TorrentFilesMoveTest` into `TorrentFilesTest`, instead of having 2 test fixtures. They are just testing different methods of the same class anyway.

tearfur:
OK, I guess the question about symlinks is resolved for the scope of this PR.

I still have a question for `contains_root`. Expanding on your example about `/data/name/file`, if the torrent has another file `name/dir1/file`, in your current logic, `dir1` will be empty but not deleted after moving, right? Can something be done about it?

h1de0x:
@tearfur yes, the previous guard left `dir1` behind because it skipped the entire cleanup tree.

I've addressed this in b50ed948a by adding an optional directory-skip predicate to `depth_first_walk()`. Cleanup now skips the protected directory itself, including its contents and the cleanup callback, while continuing through sibling directories. In your example, `dir1` is removed and `new-location` is preserved.

Existing callers in `remove()` use no predicate and retain their previous behavior.

I extended `cleanupPreservesNestedRootAndDestination` to cover the empty sibling directory.

tearfur:
Ooh I like the way you went about it, nifty!

libtransmission/torrent-files.cc
```diff
@@ -36,6 +38,7 @@ namespace
 {
 
 using file_func_t = std::function<void(std::string_view filename)>;
+using directory_filter_t = std::function<bool(std::string_view directory)>;
```
tearfur: This one's optional, up to you if you adopt it or not.

```suggestion
using path_func_t = std::function<void(std::string_view path)>;
```

libtransmission/torrent-files.cc
```diff
@@ -48,11 +51,21 @@ using file_func_t = std::function<void(std::string_view filename)>;
...
+        // A skipped directory is neither traversed nor passed to func.
+        if (skip_directory && skip_directory(path)) {
+            return;
+        }
+        if (!max_depth || *max_depth > 0) {
+            for (auto const& name : tr_sys_dir_get_files(path, tr_basename_accept_all)) {
+                depth_first_walk(tr_pathbuf{ path, '/', name }, func, max_depth ? *max_depth - 1 : max_depth, skip_directory);
+            }
         }
     }
```
tearfur, line 70: How about something like this? More succinct. Less expensive checks first, so expensive checks are potentially short-circuited.

```suggestion
    if ((!max_depth || *max_depth > 0) && //
        skip_directory && skip_directory(path) // A skipped directory is neither traversed nor passed to func.
        && is_folder(path) //
    ) {
        for (auto const& name : tr_sys_dir_get_files(path, tr_basename_accept_all)) {
            depth_first_walk(tr_pathbuf{ path, '/', name }, func, max_depth ? *max_depth - 1 : max_depth, skip_directory);
        }
    }
```

libtransmission/torrent-files.cc
```diff
@@ -142,76 +162,129 @@ std::string_view tr_torrent_files::primary_mime_type() const
...
-
-    auto const paths = std::to_array<std::string_view>({ old_parent.sv() });
-
-    auto err = bool{};
+    auto local_error = tr_error{};
+    error = error != nullptr ? error : &local_error;
 
+    // Select sources before moving files; retain them to identify cleanup trees.
+    auto moves = std::vector<std::pair<tr_file_index_t, FoundFile>>{};
     for (tr_file_index_t i = 0, n = file_count(); i < n; ++i) {
```
tearfur:
```suggestion
    auto const n = file_count();
    auto moves = std::vector<std::pair<tr_file_index_t, FoundFile>>{};
    moves.reserve(n);
    for (tr_file_index_t i = 0; i < n; ++i) {
```

libtransmission/torrent.cc
```diff
@@ -1802,7 +1787,8 @@ void tr_torrent::recheck_completeness()
             }
             date_done_ = tr_time();
 
-            if (current_dir() == incomplete_dir()) {
+            auto const incomplete = incomplete_dir().sv();
+            if (!std::empty(incomplete) && files().has_any_local_data(std::span{ &incomplete, 1U })) {
```
tearfur: *(nit)* Reduce scope.

```suggestion
            if (auto const incomplete = incomplete_dir().sv();
                !std::empty(incomplete) && files().has_any_local_data(std::span{ &incomplete, 1U })) {
```

libtransmission/torrent.h
```diff
@@ -23,6 +23,8 @@
 
 #include <sigslot/signal.hpp>
 
+#include <small/vector.hpp>
```
tearfur, line 26: `torrent.h` is included by a lot of files, so this new include statement is going to spread very far.

Let's make `tr_torrent::search_path` a local helper function just like `buildSearchPathArray`, so that we don't need to include `<small/vector.hpp>` here.

libtransmission/torrent.cc
ckerr: Restore the old test as the first clause.

When file 0 is absent, `refresh_current_dir()` falls back to `incomplete_dir`. If the data sits in `download_dir`, this gate is false, `set_location()` never runs, and nothing resets `current_dir_` or clears `incomplete_dir_` (lines 1105-1106). The done script's `TR_TORRENT_DIR` and `tr_torrentGetCurrentDir()` then name a directory that lacks the data. The old code repaired this with a no-op move.

The first clause also keeps the common case a string compare. `has_any_local_data()` stops only on a hit; a miss costs two stats per file under the torrent lock.

```suggestion
            auto const incomplete = incomplete_dir().sv();
            if (current_dir() == incomplete_dir() ||
                (!std::empty(incomplete) && files().has_any_local_data(std::span{ &incomplete, 1U }))) {
```

libtransmission/torrent.cc
```diff
@@ -702,7 +702,7 @@ void tr_torrentRemoveInSessionThread(
         }
 
         auto error = tr_error{};
-        tor->files().remove(tor->current_dir().sv(), tor->name(), remove_func, &error);
+        tor->files().remove(tor->search_paths(), tor->name(), remove_func, &error);
```
ckerr, line 705: `search_paths()` matches files by subpath, not by owner, and this call now reaches both roots.

`init()` re-seeds `incomplete_dir_` from the session on every start, so a seeding torrent searches the shared incomplete dir again. If another torrent with the same name is downloading there, deleting this torrent's data trashes the other's `.part` files. The completion gate at line 1791 has the same reach: a verify that ends in seed state moves those `.part` files into `download_dir` while the other torrent holds them open.

The old code touched only `current_dir()`. Can `init()` skip the re-seed for a torrent that loads as done and has no data under the incomplete dir? That would confine both paths to roots the torrent uses.

libtransmission/torrent-files.cc
```diff
@@ -142,76 +162,129 @@ std::string_view tr_torrent_files::primary_mime_type() const
...
             }
-
+            tr_logAddDebug(
+                path_error ? fmt::format("Skipping cleanup of '{:s}': {:s}", directory, path_error.message()) :
+                             fmt::format("Skipping cleanup of '{:s}': configured root '{:s}'", directory, root),
+                parent_name);
             return true;
         };
+        return is_root(parent.sv()) || std::ranges::any_of(old_parents, is_root);
+    };
```
ckerr, line 238: Decide once per tree whether a filter is needed; do not test every directory.

Two problems with the per-directory test:

- Cost. Each visited directory makes up to three `tr_sys_path_is_same()` calls, about twelve stats, on the session thread. The layout it guards, a configured root nested inside the torrent's own folder, is rare.
- Failure mode. Any error from `tr_sys_path_is_same()` returns "skip". An unreadable or stale root (`EACCES`, `ESTALE`) fails for every directory, so nothing is cleaned in the healthy root either, and the only trace is a debug log.

Before walking a tree, resolve `parent` and `old_parents` with `tr_sys_path_resolve()` and check whether any of them lies under the tree. If none does, walk with no filter. If resolution fails, warn once and skip that tree. `depth_first_walk()` can then keep its old signature, with the root check inside the callback.

`remove_junk()` also discards the result of `tr_sys_path_remove()`. The code it replaced logged a warning on failure; please keep that.

libtransmission/torrent-files.cc
```diff
@@ -142,76 +162,129 @@ std::string_view tr_torrent_files::primary_mime_type() const
...
+    tr_error* error) const
+{
+    auto local_error = tr_error{};
+    error = error != nullptr ? error : &local_error;
+
+    for (auto const parent : parents) {
+        // The single-root remove() also cleans junk and empty directories.
+        // Do not run that cleanup in a root with no matching torrent files.
+        if (!has_any_local_data(std::span{ &parent, 1U })) {
+            continue;
```
ckerr, line 278: This guard also skips the root the torrent did use once its payload is gone.

If the user deletes the files by hand and then removes the torrent with its data, `<root>/<name>/` keeps its empty folders and `.DS_Store`. The single-root `remove()` swept them, because it seeds `top_files` with `parent/<name>` whether or not it finds files (line 331).

Either sweep `<root>/<name>` when it exists under the root the torrent last used, or state in this comment that the skeleton is left on purpose, and add a test for the choice.

libtransmission/torrent.cc
```diff
@@ -1113,26 +1112,20 @@ void tr_torrent::set_location_in_session_thread(std::string_view const path, boo
     }
 }
 
-namespace
-{
-namespace location_helpers
-{
-size_t buildSearchPathArray(tr_torrent const* tor, std::string_view* paths)
+small::max_size_vector<std::string_view, 2> tr_torrent::search_paths() const
```
ckerr, line 1115: This PR reroutes three callers from `current_dir()` to `search_paths()`. The cause remains in `refresh_current_dir()` (line 2276): `current_dir_` comes from file 0 alone and falls back to `incomplete_dir` when file 0 is absent.

The remaining readers still get the wrong directory for such a torrent:

- `inout.cc:99` and `create_empty_files()` (line 1724) create new files there, which is how a torrent becomes split in the first place.
- `TR_TORRENT_DIR` (line 384) and `tr_torrentGetCurrentDir()` (line 1183) report it to scripts and to the GTK and macOS clients.

Derive `current_dir_` from the first file found, of any index. That fixes every reader at once.

libtransmission/local-data.cc
```diff
@@ -166,7 +166,7 @@ class DefaultBackend final : public LocalData::Backend
         }
 
         auto error = tr_error{};
-        if (tor->files().move(old_parent, parent, parent_name, &error)) {
+        if (tor->files().move(std::span{ &old_parent, 1U }, parent, parent_name, &error)) {
```
ckerr, line 169: Still open from the last round: `Backend::move()` takes one source root while `remove()` below uses `search_paths()`. Drop `old_parent` from `LocalData::move()` and `Backend::move()` and use `tor->search_paths()` here. Only `local-data-test.cc` calls it today, so the change is cheap now.

libtransmission/torrent-files.cc
```diff
@@ -142,76 +162,129 @@ std::string_view tr_torrent_files::primary_mime_type() const
 // ---
 
 bool tr_torrent_files::move(
-    std::string_view old_parent_in,
+    std::span<std::string_view const> old_parents,
     std::string_view parent_in,
     std::string_view parent_name,
```
ckerr: None of these change in the body. The new `remove()` overload below already does this.

```suggestion
    std::span<std::string_view const> const old_parents,
    std::string_view const parent_in,
    std::string_view const parent_name,
```

libtransmission/torrent-files.cc
```diff
@@ -142,76 +162,129 @@ std::string_view tr_torrent_files::primary_mime_type() const
...
-                if (auto log_error = tr_error{}; !tr_sys_path_remove(path, &log_error)) {
-                    tr_logAddWarn(
-                        fmt::format(
-                            fmt::runtime(_("Couldn't remove '{path}': {error} ({error_code})")),
-                            fmt::arg("path", path),
-                            fmt::arg("error", log_error.message()),
-                            fmt::arg("error_code", log_error.code())));
-                }
+    // Protect the roots themselves while still cleaning their siblings and
+    // ancestors. If identity cannot be checked, leave the directory alone.
```
ckerr: Break comments at sentence or clause boundaries. Same at 243, 280, `torrent-files.h:120`, and `move-test.cc:284`.

```suggestion
    // Protect the roots themselves while still cleaning their siblings and ancestors.
    // If identity cannot be checked, leave the directory alone.
```

libtransmission/torrent-files.cc
```diff
@@ -142,76 +162,129 @@ std::string_view tr_torrent_files::primary_mime_type() const
...
+    };
 
-        remove(old_parent, parent_name, remove_empty_directories);
+    for (auto const old_parent : old_parents) {
+        if (tr_sys_path_is_same(old_parent, parent)) {
+            continue;
+        }
+
+        // Derive each tree from a moved file, not the torrent name. Scan it
+        // once and leave unused trees alone.
```
ckerr:
```suggestion
        // Derive each tree from a moved file, not the torrent name.
        // Scan it once and leave unused trees alone.
```

libtransmission/torrent-files.cc
```diff
@@ -142,76 +162,129 @@ std::string_view tr_torrent_files::primary_mime_type() const
...
+    for (auto const parent : parents) {
+        // The single-root remove() also cleans junk and empty directories.
+        // Do not run that cleanup in a root with no matching torrent files.
+        if (!has_any_local_data(std::span{ &parent, 1U })) {
+            continue;
+        }
+        auto root_error = tr_error{};
+        remove(parent, tmpdir_prefix, func, &root_error);
+        // Process independent roots even when an earlier root reported an
+        // error, and preserve the first error for the caller.
```
ckerr:
```suggestion
        // Process independent roots even when an earlier root reported an error,
        // and preserve the first error for the caller.
```

libtransmission/torrent-files.h
```diff
@@ -117,6 +117,14 @@ struct tr_torrent_files {
         tr_torrent_remove_func const& func,
         tr_error* error = nullptr) const;
 
+    // Remove a matching copy from each root using the single-root deletion
+    // path and its callback. Roots without matching files are left alone.
```
ckerr:
```suggestion
    // Remove a matching copy from each root using the single-root deletion path and its callback.
    // Roots without matching files are left alone.
```

tests/libtransmission/move-test.cc
```diff
@@ -133,6 +228,80 @@ TEST_P(IncompleteDirTest, incompleteDir)
...
+
+TEST_P(IncompleteDirTest, completionMovesFilesOutsideCurrentDir)
+{
+    std::string const download_dir = tr_sessionGetDownloadDir(session_);
+    std::string const incomplete_dir = tr_sessionGetIncompleteDir(session_);
+    auto* const tor = zeroTorrentInit(ZeroTorrentState::NoFiles);
+    for (tr_file_index_t i = 0; i < tor->file_count(); ++i) {
+        auto const file = tr_torrentFile(tor, i);
+        // The first file makes current_dir point at downloadDir; all the other
+        // files must still be consolidated when verification completes.
```
ckerr:
```suggestion
        // The first file makes current_dir point at downloadDir;
        // all the other files must still be consolidated when verification completes.
```

libtransmission/torrent.h
tearfur: @ckerr I see you suggested to make `search_paths` a method instead of a helper at https://github.com/retransmission/retransmission/pull/301#discussion_r3952923901. Any objections from you about my suggestion?

libtransmission/torrent.cc
tearfur: ~Sounds to me the problem is assuming file 0 to be selected for download, no? Maybe `refresh_current_dir` should use `tr_torrent::is_done` instead?~

~If we go the `tr_torrent::is_done` route, I will merge a PR for that, then continue work in this PR.~

Edit: Nevermind all that, I understand now.

tearfur: Together with https://github.com/retransmission/retransmission/pull/301#discussion_r4023486129, the final code should be something like:

```suggestion
            if (auto const incomplete = incomplete_dir().sv(); current_dir() == incomplete ||
                (!std::empty(incomplete) && files().has_any_local_data(std::span{ &incomplete, 1U }))) {
```

libtransmission/torrent.cc
h1de0x: I agree that searching both roots can pick up another torrent's files when their relative paths overlap.

However, I'm not sure the proposed `init()` check would prevent that case. If torrent A has `download/name/file` and torrent B has `incomplete/name/file.part`, checking for A's data under the incomplete directory would find B's file. We would therefore retain `incomplete_dir_` and still include it during removal.

Did you have a different check in mind for deciding whether the incomplete directory belongs in this torrent's search paths? A path-based existence check cannot distinguish another torrent's file from a legitimate leftover of this one.

libtransmission/torrent-files.cc
h1de0x: Agreed, we shouldn't need those extra filesystem checks for every directory in the usual case.

My concern is moving the root check into the callback: `depth_first_walk()` calls it after visiting the children, so by then we may already have deleted files inside the directory we're trying to protect.

Could we check once whether the tree needs protection, then keep the check before descending only where it's needed? That should let us avoid the extra work in the usual case while still cleaning siblings of protected directories.

libtransmission/torrent.cc
h1de0x: Agreed. I'll update `refresh_current_dir()` to use the first file found, regardless of its index, and keep the existing fallback when none are found. We'll still need the searches across both roots to handle data that's already split between them.

libtransmission/local-data.cc
h1de0x: Ok, this move path should use both roots too. I'll remove `old_parent` from `LocalData::move()` and `Backend::move()`, use `tor->search_paths()` in the implementation, and update the tests.

libtransmission/torrent-files.cc
h1de0x: I'd keep those directories untouched. Once no matching torrent files remain, we can't reliably tell whether the directory was used by this torrent or is an unrelated directory with the same name. I'll make that trade-off explicit in the comment and add a small regression test.

libtransmission/torrent-files.cc
h1de0x: I think this changes the behavior in a few places: with no `skip_directory` callback, we would stop recursing, while a `true` result would make us enter the directory we're meant to skip. We also need to skip the final `func(path)` call for protected directories.

`skip_directory()` currently makes filesystem calls too, so running it before `is_folder()` would add those checks for ordinary files.

I'll revisit this together with @ckerr suggestion to decide once per tree whether filtering is needed.

tearfur: Ehem, there's quite a few stupid mistakes in my suggestion, yes. But I think you get what I'm trying to say.

Here's a corrected version.

```suggestion
    if ((!max_depth || *max_depth > 0) && //
        is_folder(path) && //
        (!skip_directory || !skip_directory(path)) // A skipped directory is neither traversed nor passed to func.
    ) {
        for (auto const& name : tr_sys_dir_get_files(path, tr_basename_accept_all)) {
            depth_first_walk(tr_pathbuf{ path, '/', name }, func, max_depth ? *max_depth - 1 : max_depth, skip_directory);
        }
    }
```

libtransmission/torrent-files.cc
h1de0x: @ckerr I've added the warning when junk removal fails.

After looking further into the cleanup checks, I'd prefer to keep the current per-directory guard in this PR. Checking containment before walking a tree can miss a symlink inside it that leads to a protected root. The current identity check catches that before descending.

I agree that the repeated filesystem checks are extra work. However, I haven't measured their impact, and the simpler alternatives would weaken existing protection. I'd prefer to address that optimization separately, with measurements, while keeping the current behavior here.

libtransmission/torrent.cc
ckerr: No, I think you're right. Let's mark this as resolved.

libtransmission/torrent-files.cc
ckerr: Yeah, it's not worth blocking this PR over

libtransmission/torrent.h
ckerr: No objections, to making `search_paths()` a local helper :+1:

ckerr (approved):
It looks like the `depth_first_walk` topic is still being discussed, deferring to @tearfur on that.

Everything else LGTM, all my review comments have been addressed

libtransmission/torrent-files.cc
h1de0x: Thanks. I took the reordered depth check, but kept the early return when `skip_directory()` matches. Otherwise the protected directory would still be passed to `func()`, and `remove_junk()` could remove it if it is empty.

The updated version avoids calling `is_folder()` when the depth is exhausted and no filter is present. When a filter is present, it still checks the directory and prevents both traversal and the final callback.

libtransmission/torrent.h
h1de0x: @tearfur agreed on removing `<small/vector.hpp>` from `torrent.h`.

Since the same search path logic is now also used by `local-data.cc`, making it a local helper would duplicate the root ordering and deduplication rules in two translation units.

How about keeping the method but letting the caller provide the storage?

```cpp
// The returned view refers to `buffer`; `buffer` must outlive it.
[[nodiscard]] std::span<std::string_view const> search_paths(
    std::span<std::string_view, 2> buffer) const;
```

A caller would use it like this:

```cpp
auto search_path_buffer = std::array<std::string_view, 2>{};
tor->files().remove(tor->search_paths(search_path_buffer), tor->name(), remove_func, &error);
```

This removes the `small::max_size_vector` dependency from the header while keeping the path selection logic in one place. The buffer is stack-allocated, and the two loops that reuse the paths would construct the view once before iterating.

Attaching the patch below.

<details>
<summary>Patch</summary>

```diff
diff --git a/libtransmission/local-data.cc b/libtransmission/local-data.cc
index 2d2e04a..6b5bc39 100644
--- a/libtransmission/local-data.cc
+++ b/libtransmission/local-data.cc
@@ -4,6 +4,7 @@
 // License text can be found in the licenses/ folder.
 
 #include <algorithm> // std::ranges::shuffle
+#include <array>
 #include <cerrno>
 #include <functional>
 #include <memory>
@@ -165,7 +166,8 @@ public:
         }
 
         auto error = tr_error{};
-        if (tor->files().move(tor->search_paths(), parent, parent_name, &error)) {
+        auto search_path_buffer = std::array<std::string_view, 2>{};
+        if (tor->files().move(tor->search_paths(search_path_buffer), parent, parent_name, &error)) {
             return 0;
         }
 
@@ -184,7 +186,8 @@ public:
         }
 
         auto error = tr_error{};
-        tor->files().remove(tor->search_paths(), tor->name(), remove_func, &error);
+        auto search_path_buffer = std::array<std::string_view, 2>{};
+        tor->files().remove(tor->search_paths(search_path_buffer), tor->name(), remove_func, &error);
         return error ? error.code() : 0;
     }
 
diff --git a/libtransmission/torrent.cc b/libtransmission/torrent.cc
index 4be33e0..e6d75f1 100644
--- a/libtransmission/torrent.cc
+++ b/libtransmission/torrent.cc
@@ -702,7 +702,8 @@ void tr_torrentRemoveInSessionThread(
         }
 
         auto error = tr_error{};
-        tor->files().remove(tor->search_paths(), tor->name(), remove_func, &error);
+        auto search_path_buffer = std::array<std::string_view, 2>{};
+        tor->files().remove(tor->search_paths(search_path_buffer), tor->name(), remove_func, &error);
         if (error) {
             tr_logAddWarnTor(
                 tor,
@@ -1084,7 +1085,8 @@ void tr_torrent::set_location_in_session_thread(std::string_view const path, boo
         session->verify_remove(this);
 
         auto error = tr_error{};
-        ok = files().move(search_paths(), path, name(), &error);
+        auto search_path_buffer = std::array<std::string_view, 2>{};
+        ok = files().move(search_paths(search_path_buffer), path, name(), &error);
         if (error) {
             this->error().set_local_error(
                 fmt::format(
@@ -1112,19 +1114,19 @@ void tr_torrent::set_location_in_session_thread(std::string_view const path, boo
     }
 }
 
-small::max_size_vector<std::string_view, 2> tr_torrent::search_paths() const
+std::span<std::string_view const> tr_torrent::search_paths(std::span<std::string_view, 2> buffer) const
 {
-    auto paths = small::max_size_vector<std::string_view, 2>{};
+    auto size = size_t{};
 
     if (!std::empty(download_dir())) {
-        paths.push_back(download_dir().sv());
+        buffer[size++] = download_dir().sv();
     }
 
     if (!std::empty(incomplete_dir()) && incomplete_dir() != download_dir()) {
-        paths.push_back(incomplete_dir().sv());
+        buffer[size++] = incomplete_dir().sv();
     }
 
-    return paths;
+    return buffer.first(size);
 }
 
 void tr_torrent::set_location(std::string_view location, bool move_from_old_path, int volatile* setme_state)
@@ -1152,12 +1154,14 @@ void tr_torrentSetLocation(
 
 std::optional<tr_torrent_files::FoundFile> tr_torrent::find_file(tr_file_index_t file_index) const
 {
-    return files().find(file_index, search_paths());
+    auto search_path_buffer = std::array<std::string_view, 2>{};
+    return files().find(file_index, search_paths(search_path_buffer));
 }
 
 bool tr_torrent::has_any_local_data() const
 {
-    return files().has_any_local_data(search_paths());
+    auto search_path_buffer = std::array<std::string_view, 2>{};
+    return files().has_any_local_data(search_paths(search_path_buffer));
 }
 
 void tr_torrentSetDownloadDir(tr_torrent* tor, std::string_view const path)
@@ -2275,7 +2279,8 @@ void tr_torrent::refresh_current_dir(std::optional<tr::shared_string> const& fir
         dir = !first_found_dir->empty() ? *first_found_dir : incomplete_dir();
     } else {
         dir = incomplete_dir();
-        auto const paths = search_paths();
+        auto search_path_buffer = std::array<std::string_view, 2>{};
+        auto const paths = search_paths(search_path_buffer);
         for (tr_file_index_t i = 0, n = file_count(); i < n; ++i) {
             if (auto const found = files().find(i, paths); found) {
                 dir = tr::shared_string{ found->base };
@@ -2535,7 +2540,8 @@ void tr_torrent::ResumeHelper::load_checked_pieces(tr_bitfield const& checked, t
     tor_.file_mtimes_.resize(n_files);
 
     first_found_dir_.emplace();
-    auto const paths = tor_.search_paths();
+    auto search_path_buffer = std::array<std::string_view, 2>{};
+    auto const paths = tor_.search_paths(search_path_buffer);
     for (size_t file = 0; file < n_files; ++file) {
         auto const found = tor_.files().find(file, paths);
         if (found && first_found_dir_->empty()) {
diff --git a/libtransmission/torrent.h b/libtransmission/torrent.h
index d24f126..00a9007 100644
--- a/libtransmission/torrent.h
+++ b/libtransmission/torrent.h
@@ -23,8 +23,6 @@
 
 #include <sigslot/signal.hpp>
 
-#include <small/vector.hpp>
-
 #include "libtransmission/announce-list.h"
 #include "libtransmission/bandwidth.h"
 #include "libtransmission/bitfield.h"
@@ -501,8 +499,8 @@ struct tr_torrent {
         metainfo_.set_file_subpath(i, subpath);
     }
 
-    // The views refer to the torrent's directories, not to the returned container.
-    [[nodiscard]] small::max_size_vector<std::string_view, 2> search_paths() const;
+    // The returned view refers to `buffer`; `buffer` must outlive it.
+    [[nodiscard]] std::span<std::string_view const> search_paths(std::span<std::string_view, 2> buffer) const;
 
     [[nodiscard]] std::optional<tr_torrent_files::FoundFile> find_file(tr_file_index_t file_index) const;
 
```
</details>

libtransmission/torrent-files.cc
```diff
@@ -142,76 +170,146 @@ std::string_view tr_torrent_files::primary_mime_type() const
...
 
         if (tr_sys_path_is_same(old_path, path)) {
             continue;
         }
 
-        tr_logAddTrace(fmt::format("Moving file #{:d} to '{:s}'", i, old_path, path), parent_name);
+        moves.emplace_back(i, *found);
+    }
+
+    if (!tr_sys_dir_create(parent, TR_SYS_DIR_CREATE_PARENTS, 0777, error)) {
```
tearfur, line 213: I think we should early return `true` before creating the target directory if `moves` is empty.

tests/libtransmission/move-test.cc
```diff
@@ -51,6 +57,112 @@ class IncompleteDirTest
     }
 
     static auto constexpr MaxWaitMsec = 3000;
+
+    int setLocation(tr_torrent* tor, std::string_view path, bool move = true)
+    {
+        auto result = std::make_shared<std::promise<int>>();
+        auto ready = result->get_future();
+        session_->run_in_session_thread([tor, path = std::string{ path }, move, result]() {
+            auto state = int{ TR_LOC_MOVING };
```
tearfur: Doesn't matter for this PR, but let's future proof `setLocation` by avoiding false negatives in case any future tests that uses this method expects `TR_LOC_MOVING`.

```suggestion
            auto state = -1;
```

tearfur (commented):
I notice there isn't any test cases for "protecting the root while cleaning their siblings and ancestors", but I think this case is rare enough that we can defer adding those cases in a future PR.

For now I'll be happy if you could address these 2 comments.

libtransmission/torrent.h
tearfur: Don't worry about it, I didn't notice `search_paths` is used outside of `torrent.cc`. 🤡

libtransmission/torrent-files.cc
h1de0x: @tearfur thanks! All three updates are now in:

* `setLocation()` initializes its test state to `-1`, as you suggested.
* `move()` returns successfully before creating the destination directory when there are no files to move (6caaa50ed).
* I've extended `cleanupPreservesNestedRootAndDestination` to check that junk in the protected root's parent is removed (a0582a4d4). It already checks that the protected root survives and its empty sibling is removed, so we can cover this here without waiting for another PR.

tearfur (approved):

h1de0x:
@ckerr @tearfur thanks for all the time and care you put into reviewing this, and for sticking with it through so many iterations. Your suggestions and the edge cases you caught made the patch much better. Glad we got it merged!

PR #302, opened 2026-09-05
fix: correct semantics for RPC error code

tearfur:
## Description

Fetching via curl in RPC is not always on HTTP, it can be on FTP too. Fix the assumption in the error codes that the curl requests are always HTTP. 

Notes: Corrected RPC error messages.

## Checklist


ckerr (approved):
👍

PR #303, opened 2026-09-05
fix: don't assume HTTP error codes

tearfur:
## Description

Fix some strings in `torrent_add` that assumes curl fetches are always HTTP.

Notes: Corrected RPC error messages.

## Checklist

### DIFF
--- a/libtransmission/rpcimpl.cc
+++ b/libtransmission/rpcimpl.cc
@@ -1508,13 +1508,9 @@ void onMetadataFetched(tr_web::FetchResponse const& web_response)
     auto* data = static_cast<struct add_torrent_idle_data*>(user_data);
 
     tr_logAddTrace(
-        fmt::format(
-            "torrentAdd: HTTP response code was {} ({}); response length was {} bytes",
-            status,
-            tr_webGetResponseStr(status),
-            std::size(body)));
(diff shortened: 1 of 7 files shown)


libtransmission/rpcimpl.cc
```diff
@@ -1508,24 +1508,17 @@ void onMetadataFetched(tr_web::FetchResponse const& web_response)
...
         add_torrent_impl(data->data, data->builder);
     } else {
         tr_rpc_idle_done(
             data->data,
             JsonRpc::Error::FETCH_ERROR,
-            fmt::format(
-                fmt::runtime(_("Couldn't fetch torrent: {error} ({error_code})")),
-                fmt::arg("error", tr_webGetResponseStr(status)),
-                fmt::arg("error_code", status)));
+            fmt::format(fmt::runtime(_("Couldn't fetch torrent: ({error_code})")), fmt::arg("error_code", status)));
```
ckerr: Feels like this makes things a little worse?  If the user doesn't know what protocol was in use, now they've just got a number with no protocol to attach it to.

What if we keep the previous version for requests we know to be HTTP/HTTPS?

tearfur: Something like this? 0e8d07556460781380ea9c6d2cf11cef6d1d3cff

ckerr (approved):

PR #305, opened 2026-09-05
fix: handle curl error codes

tearfur:
## Description

In curl, it is possible for the fetch to fail (return a non-`CURLE_OK` code), while still getting a successful response code (e.g. HTTP 200), yet we never handled this case for the longest time.

Having said that, I don't think our current usage of curl will trigger this. However, this will matter for the next PR in the stack, which can abort an HTTP 200 transfer if the remote sends too much data.

## Checklist

### DIFF
--- a/libtransmission/webseed.cc
+++ b/libtransmission/webseed.cc
@@ -401,8 +401,8 @@ void tr_webseed_task::on_data_received(size_t const n_bytes)
 
 void tr_webseed_task::on_partial_data_fetched(tr_web::FetchResponse const& web_response)
 {
-    auto const& [url, status, headers, body, primary_ip, did_connect, did_timeout, vtask] = web_response;
-    auto const success = status == 206;
+    auto const& [url, errmsg, status, headers, body, primary_ip, did_connect, did_timeout, vtask] = web_response;
+    auto const success = !errmsg && status == 206;
 
     auto* const task = static_cast<tr_webseed_task*>(vtask);
 
--- a/libtransmission/webseed.cc
+++ b/libtransmission/webseed.cc
@@ -402,7 +402,11 @@ void tr_webseed_task::on_data_received(size_t const n_bytes)
 void tr_webseed_task::on_partial_data_fetched(tr_web::FetchResponse const& web_response)
 {
     auto const& [url, errmsg, status, headers, body, primary_ip, did_connect, did_timeout, vtask] = web_response;
-    auto const success = !errmsg && status == 206;
+
+    // We could check curl error codes here, but this code has been working fine
+    // without it for years. The piece hash verifier will catch any corrupt data.
+    // Let's keep it this way until we see a problem.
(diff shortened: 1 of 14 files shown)


libtransmission/webseed.cc
```diff
@@ -401,8 +401,8 @@ void tr_webseed_task::on_data_received(size_t const n_bytes)
 
 void tr_webseed_task::on_partial_data_fetched(tr_web::FetchResponse const& web_response)
 {
-    auto const& [status, headers, body, primary_ip, did_connect, did_timeout, vtask] = web_response;
-    auto const success = status == 206;
+    auto const& [errmsg, status, headers, body, primary_ip, did_connect, did_timeout, vtask] = web_response;
+    auto const success = !errmsg && status == 206;
```
ckerr: (GPT 6 Astra, High) [P2] Preserve partial webseed progress on timeouts/short reads

A 206 response with `CURLE_OPERATION_TIMEDOUT` or `CURLE_PARTIAL_FILE` can contain complete torrent blocks. This guard now deletes the task before `use_fetched_blocks()` saves them. A 1 MiB range at 4 KiB/s exceeds the default 120-second timeout, so every retry discards its progress; previously those blocks were saved and the remaining range requested.

Please preserve complete blocks for these recoverable errors before rejecting the remainder, with a regression test for retry progress.

tearfur: I don't know.. How strict this check should be looks like a balancing act to me, rather than having a definitive answer.

Ideally, we should only accept data from webseeds that respects the `Range` header. But in practice, I've seen some webseeds that handles a lot of HTTP headers poorly, but serves legit data.

I've kept the logic unchanged for now ff214b25a2b9697e791a8a9a8be179c858bd57e1.

ckerr (approved):

tearfur:
409264b73d48c30832102a24d27fc1347ecbc82e

Fixed regressions in some message formats that crept in while resolving merge conflicts.

qt/RpcClient.cc
```diff
@@ -37,8 +37,8 @@ RpcClient::RpcClient(QObject* parent)
     : QObject{ parent }
     , impl_{ makeUiMarshaler(this) }
 {
-    connections_[0] = impl_.network_response.connect_scoped([this](long const status, std::string_view const message) {
-        auto const code = status == 200 ? QNetworkReply::NoError : QNetworkReply::UnknownNetworkError;
+    connections_[0] = impl_.network_response.connect_scoped([this](bool const is_success, std::string_view const message) {
+        auto const code = is_success ? QNetworkReply::NoError : QNetworkReply::UnknownNetworkError;
```
tearfur, line 41: Something for future PRs.

Since the RPC client is no longer built on Qt, there's not much point in using `QNetworkReply` now. We can refactor the signals and slots to move away that.

tearfur: #337

PR #306, opened 2026-09-05
fix: limit curl response size to 10 MiB by default

tearfur:
## Description

Added an opt-out response body size limit to all curl requests. One exception is requests using the `Range` header, the limit will be set to exactly the range span, and there is no opting out or customising the limit.

Requests will fail immediately if `Content-Length` exceeded the limit, or abort as soon as the received data size exceeded the limit.

Requests that are aborted in the midst of receiving the response will preserve the status code and response body received so far, so successful response codes like HTTP 200 does not mean the fetch operation succeeded.

I arbitrarily chose the default limit to be 10 MiB. This should comfortably accommodate almost all torrent files, tracker responses and whatnot.

Ref: https://curl.se/libcurl/security.html

Notes: Improved the security of connections to trackers, web seeds, and blocklist servers.

## Checklist

### DIFF
--- a/libtransmission/web.cc
+++ b/libtransmission/web.cc
@@ -14,6 +14,7 @@
 #include <cstdint> // for uint64_t
 #include <functional> // for std::less()
 #include <list>
+#include <limits>
 #include <map>
 #include <memory>
 #include <ranges>
@@ -353,14 +354,25 @@ public:
             }
         }
 
-        void add_data(void const* data, size_t const n_bytes)
+        [[nodiscard]] bool add_data(void const* data, size_t const n_bytes)
         {
+            // Needed for curl < 8.4.0, newer versions handles this for us
+            if (auto const max = options().effective_max_file_size(); response.body.size() + n_bytes > max) {
+                tr_logAddWarn(
+                    fmt::format(
+                        fmt::runtime(_("Aborting request: response body exceeded max size of {max_file_size} bytes")),
+                        fmt::arg("max_file_size", max)));
+                return false;
+            }
+
             response.body.append(static_cast<char const*>(data), n_bytes);
             tr_logAddTrace(fmt::format("wrote {} bytes to task {}'s buffer", n_bytes, fmt::ptr(this)));
 
             if (options_.on_data_received) {
                 options_.on_data_received(n_bytes);
             }
+
+            return true;
         }
 
         void add_response_header(std::string_view line)
@@ -505,7 +517,7 @@ public:
         auto* task = static_cast<Task*>(vtask);
         TR_ASSERT(std::this_thread::get_id() == task->impl.curl_thread->get_id());
 
-        if (auto const range = task->options().range) {
+        if (task->options().range) {
             // https://curl.se/libcurl/c/CURLINFO_RESPONSE_CODE.html
             // "The stored value will be zero if no server response code has been received"
             static auto constexpr NoResponseCode = 0L;
@@ -540,7 +552,11 @@ public:
             }
         }
 
-        task->add_data(data, bytes_used);
+        if (!task->add_data(data, bytes_used)) {
+            // Tell curl to error out. Failed to add data to the buffer
+            return bytes_used + 1;
+        }
+
         return bytes_used;
     }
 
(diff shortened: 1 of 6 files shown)


libtransmission/web.cc
```diff
@@ -582,6 +598,12 @@ class tr_web::Impl
         (void)curl_easy_setopt(e, CURLOPT_AUTOREFERER, 1L);
         (void)curl_easy_setopt(e, CURLOPT_ACCEPT_ENCODING, "");
         (void)curl_easy_setopt(e, CURLOPT_FOLLOWLOCATION, 1L);
+        (void)curl_easy_setopt(
+            e,
+            CURLOPT_MAXFILESIZE_LARGE,
+            static_cast<curl_off_t>(std::min(
+                task.options().effective_max_file_size(),
```
ckerr: (GPT 6 Astra, High)

[P2] Don't cap intermediate redirect responses

A four-byte range behind a 301 with `Content-Length: 162` fails before reaching the valid 206 target (reproduced on curl 8.18). This breaks small redirected webseed fragments. Disable the curl-level cap for ranged requests; `add_data()` still enforces their body limit on every curl version.

```suggestion
                task.options().range ? uint64_t{ 0 } : task.options().effective_max_file_size(),
```

libtransmission/web.h
```diff
@@ -134,7 +143,12 @@ class tr_web
         // IP protocol to use when making the request
         IPProtocol ip_proto = IPProtocol::ANY;
 
+        // Maximum size of the response body to accept before aborting the request.
+        // Use `effective_max_file_size` instead of reading this value directly
+        size_t max_file_size = DefaultMaxFileSize;
```
ckerr, line 151: (charles) I kind of doubt a `torrent_get` reply of 10 MiB is realistic, but passing this along for consideration:

(GPT 6 Astra, High) [P2] Give remote RPC responses a separate size policy

Both RPC clients inherit this default, so valid, unpaginated `torrent_get` replies over 10 MiB now fail. Explicitly opt these callers out while preserving the default for other downloads.

These two files are outside the PR diff, so this is a patch rather than an apply-suggestion block at the header line:

```diff
diff --git a/libtransmission-app/rpc-client.cc b/libtransmission-app/rpc-client.cc
--- a/libtransmission-app/rpc-client.cc
+++ b/libtransmission-app/rpc-client.cc
@@ -6,6 +6,7 @@
 #include "libtransmission-app/rpc-client.h"
 
 #include <cstdint> // int64_t
+#include <limits>
 #include <string>
 #include <string_view>
 #include <utility>
@@ -227,6 +228,7 @@
     };
 
     options.body = body;
+    options.max_file_size = std::numeric_limits<size_t>::max();
     options.auth_scheme = tr_web::FetchOptions::AuthScheme::Any;
     options.headers.insert_or_assign("Content-Type", "application/json; charset=UTF-8");
     options.headers.insert_or_assign("User-Agent", TR_PROJ_APPNAME_CAPITALIZED "/" SHORT_VERSION_STRING);
diff --git a/utils/remote.cc b/utils/remote.cc
--- a/utils/remote.cc
+++ b/utils/remote.cc
@@ -2184,3 +2184,4 @@
     };
 
     options.body = payload;
+    options.max_file_size = std::numeric_limits<size_t>::max();
```

ckerr (commented):
Idea is good and code mostly LGTM as well.

One needed change to handle redirects, and one possible change for RPC.

ckerr (approved):

libtransmission/web.cc
tearfur: dc1d241941807507385498404db316b907f65b27 I wrote a test for your scenario, but it passes. Your AI output even claims it reproduced this problem, did I miss something?

Edit: I ran my test on curl 8.22

libtransmission/web.h
tearfur: e07ba75261aff3aedd9b18730492fe1fb6117449 I understood your intent, but I didn't apply your patch directly.

tearfur:
@ckerr I see you already approved, but would you like to go through the new changes?

Edit: Ack I jumped the gun.

libtransmission/web.cc
tearfur: 78e33c07a12adfa018952fbdde8cff044e24411d Turns out it's a bug in curl < 8.10.1, here's a workaround for it.

GPT 6 probably hallucinated reproducing on curl 8.18.

PR #308, opened 2026-09-05
perf: async disk writes of Torrent data

ckerr:
## Description

Implement phase 2 of the "nonblocking disk IO" from #200.

Needs review, needs local testing, needs everything. Treat this as slop; it's only pushed here to let CI take a run at it.

Notes: Torrent disk reads and writes now run in the background, so a slow or busy drive no longer stalls the UI or other transfers, and seeding from disk is faster.

## Checklist

- [ ] I have manually written or manually audited all changes in this PR and vouch for them.
- [ ] I have explained user-facing changes in a 'Notes:' paragraph for the release notes script.


tearfur:
So this can be reviewed now yeah?

ckerr:
@tearfur I don't vouch for any of the code yet. I drove the AI through several steps; but still, this is all written by Fable 5.1 and GPT 6 Astra and I need to go through all of it more thoroughly. It's not quite vibe coding but it's vibe-adjacent, which is why I say "treat this as slop", at least until humans have gone through it more carefully.

I wouldn't force that task on anyone; but if you're interested, absolutely. Feel free! I'm happy to have the help.

h1de0x:
disk_write_budget_mib = 0 stalls downloads completely (session.cc)

```cpp
auto const budget = uint64_t{ settings_.disk_write_budget_mib } * 1024U * 1024U;
auto const requested = uint64_t{ active_request_count_ } * TrBlockSize;
auto const in_flight = local_data.enqueued_write_bytes() + requested;
return in_flight >= budget ? size_t{} : static_cast<size_t>((budget - in_flight) / TrBlockSize);
```

With budget == 0, in_flight >= budget is always true, so spare_request_blocks() always returns 0. Downstream, peers clamp to their current active count and webseeds stop entirely:

```cpp
// peer-msgs.cc
if (auto const spare = session->spare_request_blocks(); spare) {
    n_reqs = std::min(n_reqs, active_req_count(tr_direction::ClientToPeer) + *spare);
}
```

With spare == 0, n_reqs is capped at requests already in flight — nothing new can ever be requested, and the download stalls permanently. There is no minimum validation anywhere. Needs a clamp (e.g., >= 1 MiB) or treat 0 as "unbounded".

h1de0x:
`write_runs_` / `blocks_written_` counted on failed writes (local-data.cc:):

```cpp
write_runs_.fetch_add(1U, std::memory_order_relaxed);
blocks_written_.fetch_add(std::size(run), std::memory_order_relaxed);
enqueued_write_bytes_.fetch_sub(n_bytes, std::memory_order_relaxed);
```

These increment unconditionally after `tr_ioWrite`, before `result.error` is checked. A failed run still counts as a completed write/blocks-written, which will overstate throughput and mask disk errors in these stats. A couple lines below, `retained_.stash(...)` is already gated on `result.error == 0` — these two could just move inside that same check, or get a separate `write_errors_` counter if you want failures visible rather than silently excluded. Either is a small change with no architectural impact.

h1de0x:
@ckerr I didn't find any problems in the code. There's just one addition regarding the disk_write_budget_mib function — I put together a patch for it, take whichever parts of it you want, or none.

### Things to consider

`disk_write_budget_mib` documents itself as "the most memory ... that downloaded data may occupy while it waits for the disk." In practice the process's actual peak is `budget + 32 MiB`, because `MaxRetainedBytes` (the retained-block cache used to hash a piece from memory instead of reading it back) is a fixed constant, independent of whatever the user configured. The setting undercounts its own effect for every value below the default, and the gap grows as the budget shrinks:

| `disk_write_budget_mib` | peak today | peak after |
|---|---|---|
| 64 (default) | 96 MiB | 96 MiB (unchanged) |
| 32 | 64 MiB | 48 MiB |
| 16 | 48 MiB | 24 MiB |
| 8 | 40 MiB | 12 MiB |
| 1 | 33 MiB | 1.5 MiB |
| 0 (floored to 1) | 33 MiB | 1.5 MiB |

The knob is also non-linear today: asking for −87.5% (64→8) only buys −58.3% of peak memory (96→40). After this change the reduction is proportional (96→12, −87.5%), because the retained cache scales down with the same budget the user already chose.

This matters most for the audience the threaded backend exists for: NAS boxes, OpenWrt routers, and containerized seedboxes — not desktops. On a 256 MiB OpenWrt router (a common `luci-app-transmission` install target) or a container with a 512 MiB memory limit, 33 MiB of disk-IO cache is a meaningful share of the whole machine's RAM; on a desktop it's noise. Nobody on default settings is affected either way.

### The change

- `tr_session::effective_write_budget_bytes()` — the floor logic that used to live inline in `spare_request_blocks()` moves here unchanged, so both the request-budget check and the new cache-sizing code read from a single computed value. No behavior change; `spare_request_blocks()` calls it instead of recomputing the floor itself.
- `LocalData::start_workers()` gains a `retained_bytes` parameter (default `MaxRetainedBytes`, so every existing call site — including both test fixtures — keeps compiling and behaving exactly as before). `Threaded`'s constructor takes it through the same path down to `RetainedBlocks`'s existing constructor, which already accepted a runtime capacity.
- Session wires it up as `min(MaxRetainedBytes, effective_write_budget_bytes() / 2)` when starting the workers, so the effective budget — not the raw setting — drives the cache size. A budget of 0 already floors to 1 MiB elsewhere, so it floors the retained cache to 512 KiB here too, instead of needing a separate zero/unlimited special case.
- The stale `MaxRetainedBytes` comment in `local-data.cc` is replaced with the actual worst-case memory formula for the backend (`budget + retained + n_workers × MaxRunBytes staging + n_workers × 16 KiB` hash-read buffers), and the new cap in `local-data.h` documents the hit-rate side separately: below roughly `n_workers × piece size`, buffered hashing starts missing and pieces get re-read from disk. That degradation is graceful by construction — it's the same read-from-disk path piece hashes took before this PR existed (then the file-local `recalculate_hash()` in inout.cc, now `tr_ioRecalculateHash()`), just taken more often — and it's already observable via the existing `hashes_from_buffers` / `hashes_from_disk` counters, so a small budget's cost is visible rather than silent.
- One docs line: `disk_io_workers` values above 64 are silently clamped.

### Who this trades off against

This favors the memory-constrained user over the slow-disk one, who throttles the same setting for a different reason. That's deliberate: throttling doesn't depend on cache size, and on a RAM-rich machine the resulting disk reads are mostly served by the page cache anyway.

### Why now rather than later

`disk_write_budget_mib` hasn't shipped yet. Fixing what the setting means costs a few dozen lines today; doing it after real configs depend on the current (wrong) semantics would be a behavior change for anyone already running with a low budget. Nothing here is a performance optimization — on small budgets it trades a bit more disk IO (against data that's still hot in the OS page cache) for an honest memory ceiling.

<details>
<summary>Patch</summary>

```diff
diff --git a/docs/Editing-Configuration-Files.md b/docs/Editing-Configuration-Files.md
index 8eb92bc..e80a4bd 100644
--- a/docs/Editing-Configuration-Files.md
+++ b/docs/Editing-Configuration-Files.md
@@ -91,7 +91,7 @@ Here is a sample of the three basic types: respectively Boolean, Number and Stri
  * **cache_size_mib:** Number (default = 4) **DEPRECATED** and has no effect: Transmission's memory cache has been removed. The key is kept only for backwards compatibility and will be removed in the future.
  * **default_trackers:** String (default = "") A list of double-newline separated tracker announce URLs. These are used for all torrents in addition to the per torrent trackers specified in the torrent file. If a tracker is only meant to be a backup, it should be separated from its main tracker by a single newline character. If a tracker should be used additionally to another tracker it should be separated by two newlines. (e.g. "udp://tracker.example.invalid:1337/announce\n\nudp://tracker.another-example.invalid:6969/announce\nhttps://backup-tracker.another-example.invalid:443/announce\n\nudp://tracker.yet-another-example.invalid:1337/announce", in this case tracker.example.invalid, tracker.another-example.invalid and tracker.yet-another-example.invalid would be used as trackers and backup-tracker.another-example.invalid as backup in case tracker.another-example.invalid is unreachable.
  * **dht_enabled:** Boolean (default = true) Enable [Distributed Hash Table (DHT)](https://wiki.theory.org/BitTorrentSpecification#Distributed_Hash_Table).
- * **disk_io_workers:** Number (default = 0) Number of background threads for writing torrent data to disk and checking finished pieces. 0 does that work on Transmission's main network thread. 2 is a good starting point for trying the threaded writer: a slow disk no longer stalls the network or the UI, and downloads settle at the disk's speed with bounded memory. Changes take effect after a restart.
+ * **disk_io_workers:** Number (default = 0) Number of background threads for writing torrent data to disk and checking finished pieces. 0 does that work on Transmission's main network thread. 2 is a good starting point for trying the threaded writer: a slow disk no longer stalls the network or the UI, and downloads settle at the disk's speed with bounded memory. Changes take effect after a restart. Worker counts above 64 are treated as 64.
  * **disk_write_budget_mib:** Number (default = 64) When `disk_io_workers` is above 0, the most memory, in MiB, that downloaded data may occupy while it waits for the disk: blocks requested from peers but not yet received, plus blocks received but not yet written. When the disk can't keep up with the swarm, downloads slow to the disk's speed instead of buffering without limit. Values below 1 are treated as 1.
  * **encryption:** String ("allowed" = Prefer unencrypted connections, "preferred" = Prefer encrypted connections, "required" = Require encrypted connections; default = "preferred") [Encryption](https://wiki.vuze.com/w/Message_Stream_Encryption) preference. Encryption may help get around some ISP filtering, but at the cost of slightly higher CPU use.
  * **ip_endpoints_ipv4** String[] (default = ["https://ipv4.retransmission.org"]) A list of IP query endpoints to be tried in order, and the first valid result will be cached as the client's global IPv4 address. Specifying an empty array disables IP query for IPv4, but Transmission might not be able to announce your IPv4 address via legacy BEP-7 `&ipv4=` query parameter.
diff --git a/libtransmission/local-data.cc b/libtransmission/local-data.cc
index da0e37a..f6e6e76 100644
--- a/libtransmission/local-data.cc
+++ b/libtransmission/local-data.cc
@@ -359,12 +359,14 @@ public:
         Marshal marshal,
         ReadExec read_exec,
         OnFilesCreated on_files_created,
-        size_t const n_workers)
+        size_t const n_workers,
+        size_t const max_retained_bytes)
         : open_files_{ open_files }
         , provider_{ std::move(provider) }
         , marshal_{ std::move(marshal) }
         , read_exec_{ std::move(read_exec) }
         , on_files_created_{ std::move(on_files_created) }
+        , retained_{ max_retained_bytes }
     {
         workers_.reserve(n_workers);
 
@@ -537,8 +539,11 @@ private:
     // bytes and written with one call.
     static auto constexpr MaxRunBytes = uint64_t{ 1024U * 1024U };
 
-    // How much written block data to keep around for piece hashes.
-    static auto constexpr MaxRetainedBytes = size_t{ 32U * 1024U * 1024U };
+    // Worst-case memory held by the disk-IO backends:
+    //   write budget (in-flight data)
+    //   + retained blocks (<= LocalData::MaxRetainedBytes)
+    //   + n_workers x MaxRunBytes staging (<= 64 MiB at the worker cap)
+    //   + n_workers x 16 KiB (the hash-read buffer in tr_ioRecalculateHash)
 
     // --- the admission gate. Session thread only.
 
@@ -1039,7 +1044,7 @@ private:
     std::deque<std::unique_ptr<Parked>> done_;
     bool pump_scheduled_ = false;
 
-    RetainedBlocks retained_{ MaxRetainedBytes };
+    RetainedBlocks retained_;
 
     std::atomic<uint64_t> enqueued_write_bytes_ = 0U;
     std::atomic<uint64_t> write_runs_ = 0U;
@@ -1068,7 +1073,8 @@ void LocalData::start_workers(
     size_t worker_count,
     Marshal marshal,
     DescriptorProvider provider,
-    OnFilesCreated on_files_created)
+    OnFilesCreated on_files_created,
+    size_t retained_bytes)
 {
     TR_ASSERT(!threaded_);
     TR_ASSERT(completions_ == Completions::Inline); // see set_completions()
@@ -1110,7 +1116,8 @@ void LocalData::start_workers(
             return backend_->read(id, span, setme);
         },
         std::move(on_files_created),
-        worker_count);
+        worker_count,
+        retained_bytes);
 }
 
 void LocalData::read(tr_torrent_id_t const id, tr_byte_span_t const byte_span, OnRead on_read)
diff --git a/libtransmission/local-data.h b/libtransmission/local-data.h
index 19e6693..43b05ea 100644
--- a/libtransmission/local-data.h
+++ b/libtransmission/local-data.h
@@ -280,11 +280,18 @@ public:
      * Throws if the worker threads can't be spawned. The synchronous
      * backend stays in place when it does.
      */
+    // The largest size the retained-block cache may reach. The session
+    // derives the actual size from its write budget; this is the cap.
+    // Below roughly n_workers x piece size, buffered hashing misses and
+    // pieces are re-read from disk; degradation is graceful.
+    static auto constexpr MaxRetainedBytes = size_t{ 32U * 1024U * 1024U };
+
     void start_workers(
         size_t worker_count,
         Marshal marshal,
         DescriptorProvider provider = {},
-        OnFilesCreated on_files_created = {});
+        OnFilesCreated on_files_created = {},
+        size_t retained_bytes = MaxRetainedBytes);
 
     [[nodiscard]] bool is_threaded() const noexcept
     {
diff --git a/libtransmission/session.cc b/libtransmission/session.cc
index c59141c..3e21392 100644
--- a/libtransmission/session.cc
+++ b/libtransmission/session.cc
@@ -678,9 +678,17 @@ void tr_session::initImpl(init_data& data)
     // Stopping a running worker pool safely isn't worth the
     // complexity of a live switch.
     try {
-        local_data.start_workers(settings_.disk_io_workers, [this](std::function<void()> fn) {
-            queue_session_thread(std::move(fn));
-        });
+        // Scaling the cache with the budget is deliberate: it favors
+        // memory-constrained hosts at the cost of hash hit-rate; misses just
+        // re-read from disk, and throttling doesn't read this value.
+        auto const retained_bytes = static_cast<size_t>(
+            std::min<uint64_t>(tr::LocalData::MaxRetainedBytes, effective_write_budget_bytes() / 2U));
+        local_data.start_workers(
+            settings_.disk_io_workers,
+            [this](std::function<void()> fn) { queue_session_thread(std::move(fn)); },
+            {},
+            {},
+            retained_bytes);
     } catch (std::exception const& e) {
         tr_logAddError(
             fmt::format(
@@ -1937,16 +1945,20 @@ std::optional<size_t> tr_session::spare_request_blocks() const noexcept
         return {};
     }
 
-    // Floor the budget at one write run. Below that nothing is ever
-    // spare, so no peer could add a request and every download would
-    // stall.
-    static auto constexpr MinBudget = uint64_t{ 1024U } * 1024U;
-    auto const budget = std::max(uint64_t{ settings_.disk_write_budget_mib } * 1024U * 1024U, MinBudget);
+    auto const budget = effective_write_budget_bytes();
     auto const requested = uint64_t{ active_request_count_ } * TrBlockSize;
     auto const in_flight = local_data.enqueued_write_bytes() + requested;
     return in_flight >= budget ? size_t{} : static_cast<size_t>((budget - in_flight) / TrBlockSize);
 }
 
+uint64_t tr_session::effective_write_budget_bytes() const noexcept
+{
+    // Floor the budget at one write run. A budget that admits nothing
+    // would stall every download.
+    static auto constexpr MinBudget = uint64_t{ 1024U } * 1024U;
+    return std::max(uint64_t{ settings_.disk_write_budget_mib } * 1024U * 1024U, MinBudget);
+}
+
 void tr_session::invalidate_storage_descriptors()
 {
     TR_ASSERT(am_in_session_thread());
diff --git a/libtransmission/session.h b/libtransmission/session.h
index 45cc947..323dd65 100644
--- a/libtransmission/session.h
+++ b/libtransmission/session.h
@@ -744,6 +744,10 @@ public:
     // buffers there.
     [[nodiscard]] std::optional<size_t> spare_request_blocks() const noexcept;
 
+    // The effective write budget in bytes: disk_write_budget_mib,
+    // floored at one write run so that zero can't stall requests.
+    [[nodiscard]] uint64_t effective_write_budget_bytes() const noexcept;
+
     void update_active_request_count(size_t const previous, size_t const current) noexcept
     {
         TR_ASSERT(active_request_count_ >= previous);

```

</details>

rsekman:
Note that #283 said it implemented P1 but only implemented RAII wrappers for file handles, not asynchronous reads and not the scheduling that was projected to greatly improve seeding performance. So that still remains to be done. #200 still correctly notes that P1 hasn't fully landed.

PR #310, opened 2026-09-06
refactor: NSPopover management and lifecycle in TorrentTableView

lolgear:
### 📝 Description

This PR refactors how `NSPopover` is managed and displayed inside `TorrentTableView`. It cleans up legacy workarounds, improves lifecycle handling, and leverages native API properties instead of maintaining custom boolean state flags.

### 🛠 Changes

* **Protocol Adoption**: Explicitly added `<NSPopoverDelegate>` conformance to the `TorrentTableView` class extension.
* **State Management**: Replaced the custom `fActionPopoverShown` boolean flag with a direct check on the native `fActionPopover.isShown` property.
* **Code Cleanup**: Removed the `fPositioningView` layout hack previously used to alter arrow positioning.
* **Lifecycle Fix**: Added the `popoverWillClose:` delegate method to properly clear the `fActionPopover` reference when the popover is dismissed by the user or system, avoiding potential memory leaks or stale references.

### 🔍 How to Test
1. Open the torrent list view on macOS.
2. Click the gear/action button on a torrent row to open the action popover.
3. Click the button again to verify it closes instantly.
4. Open the popover and click outside to dismiss it (transient behavior). Verify that reopening it works correctly without getting stuck in a broken state.

### DIFF
--- a/macosx/TorrentTableView.mm
+++ b/macosx/TorrentTableView.mm
@@ -28,7 +28,7 @@ static CGFloat const kErrorImageSize = 20.0;
 
 static NSTimeInterval const kToggleProgressSeconds = 0.175;
 
-@interface TorrentTableView ()
+@interface TorrentTableView ()<NSPopoverDelegate>
 
 @property(nonatomic) IBOutlet Controller* fController;
 
@@ -42,7 +42,7 @@ static NSTimeInterval const kToggleProgressSeconds = 0.175;
 @property(nonatomic) CGFloat piecesBarPercent;
 @property(nonatomic) NSAnimation* fPiecesBarAnimation;
 
-@property(nonatomic) BOOL fActionPopoverShown;
+@property(nonatomic) NSPopover* fActionPopover;
 @property(nonatomic) NSView* fPositioningView;
 
 @property(nonatomic) NSDictionary* fHoverEventDict;
@@ -68,8 +68,6 @@ static NSTimeInterval const kToggleProgressSeconds = 0.175;
         if (_fCollapsedGroups == nil) {
             _fCollapsedGroups = [[NSMutableIndexSet alloc] init];
         }
-
-        _fActionPopoverShown = NO;
 
         self.delegate = self;
         self.indentationPerLevel = 0;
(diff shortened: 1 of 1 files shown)


macosx/TorrentTableView.mm
```diff
@@ -675,48 +673,33 @@ - (IBAction)revealTorrentFile:(id)sender
...
+    NSRect rect = [sender bounds];
     [popover showRelativeToRect:rect ofView:sender preferredEdge:NSMaxYEdge];
+
     [infoViewController setInfoForTorrents:@[ torrent ]];
     [infoViewController updateInfo];
-
-    CGFloat width = NSWidth(rect);
-
-    if (NSMinX(self.window.frame) < width || NSMaxX(self.window.screen.visibleFrame) - NSMinX(self.window.frame) < 72) {
-        // Ugly hack to hide NSPopover arrow.
```
nevack, line 698: How did you check and decide this hack was no longer needed?

lolgear: Put torrents window near edge and press action button.

Ok, I'll apply screenshots.

nevack: Checked with your PR
IMO this just looks bad, as it was before the hack was introduced.

Screenshot 2026-09-06 at 17 30 39

nevack: Related to
https://github.com/transmission/transmission/issues/3293
https://github.com/transmission/transmission/pull/3294

lolgear: Well...

```objective-c
[self.fActionPopover setValue:@(YES) forKey:@"shouldHideAnchor"];
```

nevack: I do not understand. Is it some undocumented private API?

lolgear: Yes, it is a private API.

https://github.com/mstg/OSX-Runtime-Headers/blob/master/AppKit/NSPopover.h

ckerr: Not sure if I'd block it or not but I don't love using undocumented API

lolgear:
#338

I guess these two are connected.

The whole app is a house of mad decisions—it’s like a time capsule from the mid-2000s and early 2010s. By 'mad,' I mean 'crazy' in the sense that nothing worked out of the box back in that era, so developers had to constantly reinvent the wheel or rely on every bad practice in the book just to get things to function.

But back then, hacking AppKit wasn't even the main goal; they were just trying to survive the platform's limitations.

My takeaway from this?
I suggest we stop hacking. Either use the appropriate tool, component, or approach, or don't use it at all.

ckerr:
@lolgear please resolve conflicts and avoid the undocumented API usage

macosx/TorrentTableView.mm
```diff
@@ -28,7 +28,7 @@
 
 static NSTimeInterval const kToggleProgressSeconds = 0.175;
 
-@interface TorrentTableView ()
+@interface TorrentTableView ()<NSPopoverDelegate>
```
ckerr: (Opus 5.5 xhigh) TorrentTableView.h:11 already declares `NSPopoverDelegate`. Please revert this line. That also clears the only merge conflict with main.

```suggestion
@interface TorrentTableView ()
```

macosx/TorrentTableView.mm
```diff
@@ -42,7 +42,7 @@ @interface TorrentTableView ()
 @property(nonatomic) CGFloat piecesBarPercent;
 @property(nonatomic) NSAnimation* fPiecesBarAnimation;
 
-@property(nonatomic) BOOL fActionPopoverShown;
+@property(nonatomic) NSPopover* fActionPopover;
 @property(nonatomic) NSView* fPositioningView;
```
ckerr: (Opus 5.5 xhigh) This change removes every use of `fPositioningView`. Please delete the property unless you restore the arrow hack.

```suggestion
```

macosx/TorrentTableView.mm
```diff
@@ -675,48 +673,33 @@ - (IBAction)revealTorrentFile:(id)sender
...
-}
-
-//don't show multiple popovers when clicking the gear button repeatedly
-- (void)popoverWillShow:(NSNotification*)notification
-{
-    self.fActionPopoverShown = YES;
 }
 
-- (void)popoverDidClose:(NSNotification*)notification
+- (void)popoverWillClose:(NSNotification*)notification
```
ckerr: (Opus 5.5 xhigh) This clears `fActionPopover` when the fade-out starts. A gear click during the fade then opens a second popover. Main ignored those clicks until `popoverDidClose:`.

Please clear the reference in `popoverDidClose:` instead. Then test `self.fActionPopover != nil` at line 676 and delete line 678.

Step 3 of the test plan likely hits this. AppKit closes a transient popover on mouse-down. The gear sends its action on the following mouse-up, after `popoverWillClose:` has cleared `fActionPopover`. I could not test this on macOS. Please confirm it with a quick click on the gear while the popover is open.

```suggestion
- (void)popoverDidClose:(NSNotification*)notification
```

macosx/TorrentTableView.mm
ckerr: (Opus 5.5 xhigh) You can keep the arrow hidden with public API only. Please restore this block and its cleanup in `popoverDidClose:`. Add the positioning view to `sender` instead of `self`, as `showGlobalPopover:` in Controller.mm does.

Main adds the view to `self` with `[sender bounds]` as its frame. That puts the anchor at the table's top-left corner instead of the gear. Upstream transmission/transmission#5147 introduced this when it made `rect` sender-local.

ckerr (changes requested):
(Opus 5.5 xhigh) This PR changes how the torrent options popover looks when the window is near the screen edge. Please add a `Notes:` paragraph to the description that tells users what changed.

lolgear:
@ckerr 
I mentioned issue ( #338 ) which shows how this implementation of this feature corrupts macOS UI state. ( very serious UI state corruption ).
Either follow Apple recommendation ( Popover **should have an arrow** ) or rewrite this feature completely/remove it.

PR #311, opened 2026-09-06
perf: optimize table view rendering with targeted row reloads

lolgear:
### 📝 Description
This PR optimizes the rendering performance of `TorrentTableView` by replacing full view redraw requests with targeted updates of visible rows only. It also slightly improves the visual fidelity of the pieces bar toggle transition.

### 🛠 Changes
* **Performance Optimization**: Replaced `self.needsDisplay = YES` and the deprecated `[self display]` method with `[self reloadVisibleRows]`. This ensures that macOS only re-renders the rows currently visible on the screen during general table refreshes, modifier key changes (`flagsChanged:`), and animation progress updates, significantly reducing CPU/GPU overhead.
* **Animation Polish**: Changed the animation curve for the pieces bar toggle (`fPiecesBarAnimation`) from `NSAnimationEaseIn` to `NSAnimationEaseInOut` to provide a smoother, more natural-looking acceleration and deceleration transition.

### 🔍 How to Test
1. Open the app with a large list of torrents.
2. Hold down the `Option` key and release it to verify that the pause/resume buttons smoothly change color to orange and back without flickering or lagging the UI.
3. Toggle the pieces bar view and ensure the animation triggers and completes smoothly.
4. Verify that general automatic state updates in the torrent list reflect correctly without full table stutters.


macosx/TorrentTableView.mm
```diff
@@ -751,7 +751,7 @@ - (void)animation:(NSAnimation*)animation didReachProgressMark:(NSAnimationProgr
             self.piecesBarPercent = 1.0 - progress;
         }
 
-        self.needsDisplay = YES;
+        [self reloadVisibleRows];
```
ckerr: (GPT 5.6 Sol, xhigh) This runs a synchronous row reload at each of 16 animation ticks in 175 ms. AppKit drops every visible view-cell and calls its delegate; this class then rebuilds labels/icons and resets controls. `needsDisplay` is deferred and coalesced, so this can increase CPU and make the toggle stutter. Keep redraw invalidation on this hot path.

```suggestion
        self.needsDisplay = YES;
```

ckerr (dismissed):

macosx/TorrentTableView.mm
lolgear: Both are bad, but `reloadVisibleRows` at least _animates_ changes.

ckerr: wdyt about changing it this way? This should remove the redundant computations and still animate?

```suggestion
        [self enumerateAvailableRowViewsUsingBlock:^(NSTableRowView* rowView, NSInteger) {
            [[rowView viewAtColumn:0] setNeedsDisplay:YES];
        }];
```

macosx/Controller.mm
```diff
@@ -3489,7 +3489,7 @@ - (void)togglePiecesBar:(id)sender
 - (void)toggleAvailabilityBar:(id)sender
 {
     [self.fDefaults setBool:![self.fDefaults boolForKey:@"DisplayProgressBarAvailable"] forKey:@"DisplayProgressBarAvailable"];
-    [self.fTableView display];
+    [self.fTableView needsDisplay];
```
ckerr, line 3492: `needsDisplay` is the getter. This line reads a BOOL and discards it, so the availability bar no longer redraws when toggled.

```suggestion
    [self.fTableView reloadVisibleRows];
```

macosx/TorrentTableView.mm
```diff
@@ -487,7 +487,7 @@ - (NSMenu*)menuForEvent:(NSEvent*)event
 //make sure that the pause buttons become orange when holding down the option key
 - (void)flagsChanged:(NSEvent*)event
 {
-    [self display];
+    [self reloadVisibleRows];
```
ckerr, line 490: Reloading rows here causes two problems.

1. `reloadDataForRowIndexes:` calls `resetImage` on every visible control button, which forces the `Off` image. Hover a paused torrent's resume button and press Option: the button shows `ResumeNoWaitOff`, not `ResumeNoWaitHover`, until the mouse leaves and returns.

2. It costs more than the `display` it replaces. A view-based table keeps views only for visible rows, so `display` already skipped offscreen rows. This now reruns `outlineView:viewForTableColumn:item:` for each visible row on every modifier change, including Shift, Cmd, Ctrl, and Fn. With SortByGroup on, it also walks every row from 0 to the last visible one.

Only the control button image depends on Option. Expose `updateImage` on `TorrentCellControlButton` and call it on the visible rows' buttons, and only when `NSEventModifierFlagOption` changes.

macosx/TorrentTableView.mm
```diff
@@ -91,7 +91,7 @@ - (void)awakeFromNib
 
 - (void)refreshTorrentTable
 {
-    self.needsDisplay = YES;
+    [self reloadVisibleRows];
```
ckerr, line 94: `refreshTorrentTable` now only forwards to `reloadVisibleRows`. Register the `RefreshTorrentTable` observer in `awakeFromNib` with `@selector(reloadVisibleRows)` and delete this method.

PR #312, opened 2026-09-06
refactor: replace PiecesView mouseDown override with gesture recognizer.

lolgear:
### Description
This PR addresses one of the legacy macOS code architectural issues mentioned in #289. It removes the custom `mouseDown:` override from `PiecesView` and migrates the click interaction to a modern `NSClickGestureRecognizer` within `InfoActivityViewController`.

### Changes Included
- **UI & Architecture:** Removed the legacy action/target connections from `InfoActivityView.xib` and deleted the `mouseDown:` override in `PiecesView.mm`.
- **Modern AppKit Standards:** Initialized and attached an `NSClickGestureRecognizer` to `fPiecesView` in `InfoActivityViewController.mm` to abstract core input handling [TN3212].
- **UX Improvement:** Clicking now triggers standard macOS behavior (on mouse release instead of press), aligning with Apple's Human Interface Guidelines.
- **Separation of Concerns:** Moved the settings toggling logic (`PiecesViewShowAvailability`) out of the View class and into the View Controller.
- **Bug Fix:** Added a strict safeguard (`self.fTorrents.count == 1`) to ensure this interaction changes state only when a single torrent is selected.

### Verification
- [x] Verified that clicking `PiecesView` correctly toggles the availability view when one torrent is selected.
- [x] Verified that clicking is completely ignored when multiple torrents are selected.
- [x] Checked cross-device compliance (works seamlessly with trackpads, mouse clicks, and simulated taps).

### DIFF
--- a/macosx/InfoActivityViewController.mm
+++ b/macosx/InfoActivityViewController.mm
@@ -260,7 +260,7 @@ static CGFloat const kStackViewVerticalSpacing = 8.0;
 
 - (void)updatePiecesView:(id)sender
 {
-    if (self.fTorrents.count == 1) {
+    if (self.fTorrents.count == 1 && self.fTorrents[0].magnet == NO) {
         BOOL const availablity = [NSUserDefaults.standardUserDefaults boolForKey:@"PiecesViewShowAvailability"];
         [self setPiecesViewAvailability:!availablity];
     }
--- a/macosx/InfoActivityViewController.mm
+++ b/macosx/InfoActivityViewController.mm
@@ -40,7 +40,6 @@ static CGFloat const kStackViewVerticalSpacing = 8.0;
 
 @property(nonatomic) IBOutlet PiecesView* fPiecesView;
 @property(nonatomic) IBOutlet NSSegmentedControl* fPiecesControl;
-@property(nonatomic, readonly) NSClickGestureRecognizer* fGestureRecognizer;
 
 @property(nonatomic) IBOutlet NSStackView* fActivityStackView;
 @property(nonatomic) IBOutlet NSView* fDatesView;
@@ -68,10 +67,8 @@ static CGFloat const kStackViewVerticalSpacing = 8.0;
     [super awakeFromNib];
     [self checkWindowSize];
 
-    if (self.fGestureRecognizer == nil) {
-        _fGestureRecognizer = [[NSClickGestureRecognizer alloc] initWithTarget:self action:@selector(updatePiecesView:)];
-        [self.fPiecesView addGestureRecognizer:_fGestureRecognizer];
-    }
+    auto gestureRecognizer = [[NSClickGestureRecognizer alloc] initWithTarget:self action:@selector(togglePiecesViewAvailability:)];
+    [self.fPiecesView addGestureRecognizer:gestureRecognizer];
 }
 
 - (CGFloat)fHorizLayoutHeight
@@ -242,27 +239,39 @@ static CGFloat const kStackViewVerticalSpacing = 8.0;
     }
 }
 
+- (BOOL)shouldTogglePiecesViewAvailability
+{
+    return self.fPiecesView.torrent != nil;
+}
+
 - (void)setPiecesViewAvailability:(BOOL)piecesAvailableSegment
 {
     [NSUserDefaults.standardUserDefaults setBool:piecesAvailableSegment forKey:@"PiecesViewShowAvailability"];
 
-    [self.fPiecesControl setSelected:piecesAvailableSegment forSegment:PiecesControlSegmentAvailable];
-    [self.fPiecesControl setSelected:!piecesAvailableSegment forSegment:PiecesControlSegmentProgress];
+    [self syncPiecesControl:piecesAvailableSegment];
 
     [self.fPiecesView updateView];
 }
 
-- (void)setPiecesView:(id)sender
+- (void)syncPiecesControl:(BOOL)piecesAvailableSegment
 {
-    BOOL const availability = [sender selectedSegment] == PiecesControlSegmentAvailable;
-    [self setPiecesViewAvailability:availability];
+    [self.fPiecesControl setSelected:piecesAvailableSegment forSegment:PiecesControlSegmentAvailable];
+    [self.fPiecesControl setSelected:!piecesAvailableSegment forSegment:PiecesControlSegmentProgress];
 }
 
-- (void)updatePiecesView:(id)sender
+- (IBAction)setPiecesView:(id)sender
 {
-    if (self.fTorrents.count == 1 && self.fTorrents[0].magnet == NO) {
-        BOOL const availablity = [NSUserDefaults.standardUserDefaults boolForKey:@"PiecesViewShowAvailability"];
-        [self setPiecesViewAvailability:!availablity];
+    if (self.shouldTogglePiecesViewAvailability) {
+        BOOL const availability = [sender selectedSegment] == PiecesControlSegmentAvailable;
+        [self setPiecesViewAvailability:availability];
+    }
+}
+
+- (void)togglePiecesViewAvailability:(id)sender
+{
+    if (self.shouldTogglePiecesViewAvailability) {
+        BOOL const availability = [NSUserDefaults.standardUserDefaults boolForKey:@"PiecesViewShowAvailability"];
+        [self setPiecesViewAvailability:!availability];
     }
 }
 
@@ -300,17 +309,17 @@ static CGFloat const kStackViewVerticalSpacing = 8.0;
 
         [self.fPiecesControl setSelected:NO forSegment:PiecesControlSegmentAvailable];
         [self.fPiecesControl setSelected:NO forSegment:PiecesControlSegmentProgress];
-        self.fPiecesControl.enabled = NO;
         self.fPiecesView.torrent = nil;
+
+        self.fPiecesControl.enabled = self.shouldTogglePiecesViewAvailability;
     } else {
         Torrent* torrent = self.fTorrents[0];
 
         BOOL const piecesAvailableSegment = [NSUserDefaults.standardUserDefaults boolForKey:@"PiecesViewShowAvailability"];
-        [self.fPiecesControl setSelected:piecesAvailableSegment forSegment:PiecesControlSegmentAvailable];
-        [self.fPiecesControl setSelected:!piecesAvailableSegment forSegment:PiecesControlSegmentProgress];
-        self.fPiecesControl.enabled = YES;
-
+        [self syncPiecesControl:piecesAvailableSegment];
         self.fPiecesView.torrent = torrent;
+
+        self.fPiecesControl.enabled = self.shouldTogglePiecesViewAvailability;
     }
 
     self.fSet = YES;
--- a/macosx/InfoActivityViewController.mm
+++ b/macosx/InfoActivityViewController.mm
@@ -40,6 +40,7 @@ static CGFloat const kStackViewVerticalSpacing = 8.0;
 
 @property(nonatomic) IBOutlet PiecesView* fPiecesView;
 @property(nonatomic) IBOutlet NSSegmentedControl* fPiecesControl;
+@property(nonatomic, readonly) NSClickGestureRecognizer* fGestureRecognizer;
 
 @property(nonatomic) IBOutlet NSStackView* fActivityStackView;
 @property(nonatomic) IBOutlet NSView* fDatesView;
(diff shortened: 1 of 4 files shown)


macosx/InfoActivityViewController.mm
```diff
@@ -236,21 +242,28 @@ - (void)updateInfo
     }
 }
 
+- (void)setPiecesViewAvailability:(BOOL)piecesAvailableSegment
+{
+    [NSUserDefaults.standardUserDefaults setBool:piecesAvailableSegment forKey:@"PiecesViewShowAvailability"];
```
nevack, line 249: nit: I would suggest extracting `@"PiecesViewShowAvailability"` string into top-level const for the sake of preventing mistyping and reusability.

macosx/InfoActivityViewController.mm
```diff
@@ -236,21 +242,28 @@ - (void)updateInfo
...
+
+    [self.fPiecesView updateView];
+}
+
 - (void)setPiecesView:(id)sender
 {
     BOOL const availability = [sender selectedSegment] == PiecesControlSegmentAvailable;
-    [NSUserDefaults.standardUserDefaults setBool:availability forKey:@"PiecesViewShowAvailability"];
-    [self updatePiecesView:nil];
+    [self setPiecesViewAvailability:availability];
```
nevack: nit:
Could we change BOOL parameter here to a typed enum?

nevack (approved):
Overall LGTM

macosx/InfoActivityViewController.mm
lolgear: Like:

```objective-c
availability ? PiecesControlSegmentAvailable : PiecesControlSegmentProgress
```

or

```objective-c
(PiecesControlSegment)availability
```

macosx/InfoActivityViewController.mm
lolgear: Not now, @"PiecesViewShowAvailability" is still in `PiecesView`.
I don't want to touch Xcode project until `groups to folders conversion` arrival.

macosx/InfoActivityViewController.mm
```diff
@@ -236,21 +242,28 @@ - (void)updateInfo
...
 
 - (void)updatePiecesView:(id)sender
 {
-    BOOL const piecesAvailableSegment = [NSUserDefaults.standardUserDefaults boolForKey:@"PiecesViewShowAvailability"];
-
-    [self.fPiecesControl setSelected:piecesAvailableSegment forSegment:PiecesControlSegmentAvailable];
-    [self.fPiecesControl setSelected:!piecesAvailableSegment forSegment:PiecesControlSegmentProgress];
-
-    [self.fPiecesView updateView];
+    if (self.fTorrents.count == 1) {
```
ckerr: For a magnet awaiting metadata, `PiecesView.torrent` is nil but `fTorrents.count` is one.

```suggestion
    if (self.fTorrents.count == 1 && self.fPiecesView.torrent != nil) {
```

lolgear: Maybe check that `self.fTorrents[0].magnet == NO`?

lolgear: Added.

macosx/PiecesView.mm
```diff
@@ -191,21 +191,4 @@ - (void)updateView
     fRenderedHashString = self.torrent.hashString;
 }
 
-- (BOOL)acceptsFirstMouse:(NSEvent*)event
-{
-    return YES;
-}
```
ckerr, line 197: (Opus 5, extra) Restore `acceptsFirstMouse:`. The gesture recognizer replaces `mouseDown:`, not first-mouse acceptance, and AppKit still gates mouse-down delivery on the hit view's `acceptsFirstMouse:`. With the Inspector window not key, the click is now spent activating the window, so toggling takes two clicks.

```objectivec
- (BOOL)acceptsFirstMouse:(NSEvent*)event
{
    return YES;
}
```

macosx/InfoActivityViewController.mm
```diff
@@ -236,21 +242,28 @@ - (void)updateInfo
...
 
 - (void)updatePiecesView:(id)sender
 {
-    BOOL const piecesAvailableSegment = [NSUserDefaults.standardUserDefaults boolForKey:@"PiecesViewShowAvailability"];
-
-    [self.fPiecesControl setSelected:piecesAvailableSegment forSegment:PiecesControlSegmentAvailable];
-    [self.fPiecesControl setSelected:!piecesAvailableSegment forSegment:PiecesControlSegmentProgress];
-
-    [self.fPiecesView updateView];
+    if (self.fTorrents.count == 1 && self.fTorrents[0].magnet == NO) {
```
ckerr: (Opus 5, extra) Ask the view instead of re-deriving its rule. `PiecesView.setTorrent:` already stores nil for a magnet, and `setupInfo` already nils it unless exactly one torrent is selected, so `self.fPiecesView.torrent != nil` is the same condition read from the object that owns it - and it is the value `updateView` itself checks.

It is also correct in a case this isn't: when a magnet gains metadata, `isMagnet` flips to `NO` immediately but `fPiecesView.torrent` stays nil until the next `setupInfo`. In that window a click writes the preference and moves the segmented control while the display does not change.

```suggestion
    if (self.fPiecesView.torrent != nil) {
```

Separately, `setupInfo` leaves `fPiecesControl` enabled for a magnet, so the segmented control still switches availability where the click no longer does. Disable the control too, or drop the magnet condition.

macosx/InfoActivityViewController.mm
```diff
@@ -236,21 +242,28 @@ - (void)updateInfo
...
+        BOOL const availablity = [NSUserDefaults.standardUserDefaults boolForKey:@"PiecesViewShowAvailability"];
+        [self setPiecesViewAvailability:!availablity];
```
ckerr: (Opus 5, extra) Typo: `availablity`.

```suggestion
        BOOL const availability = [NSUserDefaults.standardUserDefaults boolForKey:@"PiecesViewShowAvailability"];
        [self setPiecesViewAvailability:!availability];
```

macosx/InfoActivityViewController.mm
```diff
@@ -236,21 +242,28 @@ - (void)updateInfo
...
+
 - (void)setPiecesView:(id)sender
 {
     BOOL const availability = [sender selectedSegment] == PiecesControlSegmentAvailable;
-    [NSUserDefaults.standardUserDefaults setBool:availability forKey:@"PiecesViewShowAvailability"];
-    [self updatePiecesView:nil];
+    [self setPiecesViewAvailability:availability];
 }
 
 - (void)updatePiecesView:(id)sender
```
ckerr: (Opus 5, extra) Rename - this no longer updates, it toggles. `togglePiecesViewAvailability:` says what it does.

Also move the declaration out of `InfoActivityViewController.h` and into the class extension. With the xib action removed the gesture recognizer is the only caller, so it needs to be neither public nor an `IBAction`. An `IBAction` named `updatePiecesView:` left in the header invites a future nib to wire it up expecting a refresh and get a toggle.

macosx/InfoActivityViewController.mm
```diff
@@ -66,6 +67,11 @@ - (void)awakeFromNib
 {
     [super awakeFromNib];
     [self checkWindowSize];
+
+    if (self.fGestureRecognizer == nil) {
+        _fGestureRecognizer = [[NSClickGestureRecognizer alloc] initWithTarget:self action:@selector(updatePiecesView:)];
+        [self.fPiecesView addGestureRecognizer:_fGestureRecognizer];
+    }
```
ckerr: (Opus 5, extra) Drop `fGestureRecognizer` (declared on line 43). `fPiecesView` retains the recognizer and nothing reads the property, so it exists only to guard a repeat `awakeFromNib` that a single nib load never produces. If you keep the guard, read and write the ivar consistently - this block mixes `self.fGestureRecognizer` with `_fGestureRecognizer`.

macosx/InfoActivityViewController.mm
```diff
@@ -236,21 +242,28 @@ - (void)updateInfo
     }
 }
 
+- (void)setPiecesViewAvailability:(BOOL)piecesAvailableSegment
+{
+    [NSUserDefaults.standardUserDefaults setBool:piecesAvailableSegment forKey:@"PiecesViewShowAvailability"];
+
+    [self.fPiecesControl setSelected:piecesAvailableSegment forSegment:PiecesControlSegmentAvailable];
+    [self.fPiecesControl setSelected:!piecesAvailableSegment forSegment:PiecesControlSegmentProgress];
```
ckerr: (Opus 5, extra) `setupInfo` repeats these two lines. Extract them so the two segments cannot be updated in one place and missed in the other:

```objectivec
- (void)syncPiecesControl:(BOOL)availability
{
    [self.fPiecesControl setSelected:availability forSegment:PiecesControlSegmentAvailable];
    [self.fPiecesControl setSelected:!availability forSegment:PiecesControlSegmentProgress];
}
```

ckerr (commented):
re-reviewed with opus 5, extra, prompt:

> /code-review PR 312. and note the prior review that can be found on that page. OP has pushed more commits. If there are existing discussions that are resolved, mark them as resolved. If there are new findings. add them as draft comments and/or draft code diff suggestions . Precede each one with "(Opus 5, extra)". Use direct, actionable language. Omit unnecessary context. Do not publish the re-review; leave it in draft form.

macosx/InfoActivityViewController.mm
```diff
@@ -287,17 +309,17 @@ - (void)setupInfo
...
         BOOL const piecesAvailableSegment = [NSUserDefaults.standardUserDefaults boolForKey:@"PiecesViewShowAvailability"];
-        [self.fPiecesControl setSelected:piecesAvailableSegment forSegment:PiecesControlSegmentAvailable];
-        [self.fPiecesControl setSelected:!piecesAvailableSegment forSegment:PiecesControlSegmentProgress];
-        self.fPiecesControl.enabled = YES;
-
+        [self syncPiecesControl:piecesAvailableSegment];
         self.fPiecesView.torrent = torrent;
+
+        self.fPiecesControl.enabled = self.shouldTogglePiecesViewAvailability;
     }
```
ckerr, line 323: Select a single magnet link: the control is disabled yet shows a selected segment. With no torrent or several selected, it shows none. Make the disabled state consistent.

Also set `enabled` once, after the branches. In the first branch the expression is always `NO`, which the reader learns only by tracing `PiecesView`'s setter.

```suggestion
        self.fPiecesView.torrent = nil;
    } else {
        self.fPiecesView.torrent = self.fTorrents[0];
    }

    BOOL const enabled = self.shouldTogglePiecesViewAvailability;
    self.fPiecesControl.enabled = enabled;
    if (enabled) {
        [self syncPiecesControl:[NSUserDefaults.standardUserDefaults boolForKey:@"PiecesViewShowAvailability"]];
    } else {
        [self.fPiecesControl setSelected:NO forSegment:PiecesControlSegmentAvailable];
        [self.fPiecesControl setSelected:NO forSegment:PiecesControlSegmentProgress];
    }
```

macosx/InfoActivityViewController.mm
```diff
@@ -236,21 +239,40 @@ - (void)updateInfo
...
+}
 
-    [self.fPiecesView updateView];
+- (IBAction)setPiecesView:(id)sender
+{
+    if (self.shouldTogglePiecesViewAvailability) {
+        BOOL const availability = [sender selectedSegment] == PiecesControlSegmentAvailable;
+        [self setPiecesViewAvailability:availability];
+    }
+}
```
ckerr, line 268: Remove this guard. The only sender is `fPiecesControl`, which is disabled whenever the guard is false. Were it ever to fail, the segment would already have flipped, leaving the control out of step with the stored preference.

```suggestion
- (IBAction)setPiecesView:(id)sender
{
    BOOL const availability = [sender selectedSegment] == PiecesControlSegmentAvailable;
    [self setPiecesViewAvailability:availability];
}
```

macosx/InfoActivityViewController.mm
```diff
@@ -236,21 +239,40 @@ - (void)updateInfo
...
+    [self syncPiecesControl:piecesAvailableSegment];
+
+    [self.fPiecesView updateView];
+}
 
+- (void)syncPiecesControl:(BOOL)piecesAvailableSegment
+{
     [self.fPiecesControl setSelected:piecesAvailableSegment forSegment:PiecesControlSegmentAvailable];
     [self.fPiecesControl setSelected:!piecesAvailableSegment forSegment:PiecesControlSegmentProgress];
+}
```
ckerr, line 260: nit: `piecesAvailableSegment` names a flag, not a segment. The callers call it `availability`.

```suggestion
- (void)setPiecesViewAvailability:(BOOL)availability
{
    [NSUserDefaults.standardUserDefaults setBool:availability forKey:@"PiecesViewShowAvailability"];

    [self syncPiecesControl:availability];

    [self.fPiecesView updateView];
}

- (void)syncPiecesControl:(BOOL)availability
{
    [self.fPiecesControl setSelected:availability forSegment:PiecesControlSegmentAvailable];
    [self.fPiecesControl setSelected:!availability forSegment:PiecesControlSegmentProgress];
}
```

lolgear: Done.

macosx/InfoActivityViewController.mm
lolgear: Done.

macosx/InfoActivityViewController.mm
lolgear: Done.

PR #315, opened 2026-09-07
refactor: mac toolbar controller

nevack:


lolgear:
Let's wait for #224

macosx/MainToolbarController.h
```diff
@@ -0,0 +1,39 @@
...
+
+#import <AppKit/AppKit.h>
+
+@class Torrent;
+
+@protocol MainToolbarControllerDelegate<NSObject>
+
+- (NSArray<Torrent*>*)mainToolbarAllTorrents;
+- (NSArray<Torrent*>*)mainToolbarSelectedTorrents;
+- (NSInteger)mainToolbarSelectedTorrentCount;
```
ckerr, line 13: (opus 5, extra) Name says torrents; Controller implements this as `self.fTableView.numberOfSelectedRows`. `-[TorrentTableView selectedTorrents]` expands a selected group row into its member torrents, so the two quantities differ. Rename to `mainToolbarSelectedRowCount`, or drop it and use `[self.delegate mainToolbarSelectedTorrents].count` at the three call sites (lines 334, 392, 397).

macosx/MainToolbarController.h
```diff
@@ -0,0 +1,39 @@
...
+- (void)mainToolbarStopAllTorrents:(id)sender;
+- (void)mainToolbarResumeAllTorrents:(id)sender;
+- (void)mainToolbarStopSelectedTorrents:(id)sender;
+- (void)mainToolbarResumeSelectedTorrents:(id)sender;
+
+@end
+
+@interface MainToolbarController : NSObject<NSToolbarDelegate, NSToolbarItemValidation>
+
+- (instancetype)initWithDelegate:(id<MainToolbarControllerDelegate>)delegate;
```
ckerr: (opus 5, extra) `[[MainToolbarController alloc] init]` compiles today and yields a nil delegate: the toolbar renders, every button is dead, Pause All/Resume All validate as disabled, no diagnostic. `PowerManager.h` already uses this pattern.

```suggestion
- (instancetype)init NS_UNAVAILABLE;
- (instancetype)initWithDelegate:(id<MainToolbarControllerDelegate>)delegate NS_DESIGNATED_INITIALIZER;
```

macosx/MainToolbarController.mm
```diff
@@ -0,0 +1,443 @@
...
+- (instancetype)initWithDelegate:(id<MainToolbarControllerDelegate>)delegate
+{
+    if ((self = [super init])) {
+        _delegate = delegate;
+    }
+
+    return self;
+}
+
+- (NSToolbar*)createToolbar
```
ckerr, line 52: (opus 5, extra) The toolbar controller hands the toolbar away and keeps no reference, so Controller still reaches through the window for it: `validateVisibleItems` at Controller.mm 1981, 2065, 3331, 3516, and `hideToolBarBezels:` iterating `self.fWindow.toolbar.items` at 4239. Build the toolbar in `initWithDelegate:`, expose `@property(nonatomic, readonly) NSToolbar* toolbar`, and add `-validate` plus a bezel-hiding method here.

As written, a second `createToolbar` call also silently produces a second toolbar sharing the `TRMainToolbar` autosave identifier.

macosx/MainToolbarController.mm
```diff
@@ -0,0 +1,443 @@
...
+- (__kindof ButtonToolbarItem*)toolbarButtonWithIdentifier:(NSString*)ident forToolbarButtonClass:(Class)klass
+{
+    ButtonToolbarItem* item = [[klass alloc] initWithItemIdentifier:ident];
+
+    NSButton* button = [[NSButton alloc] init];
+    button.bezelStyle = NSBezelStyleTexturedRounded;
+    button.stringValue = @"";
+
+    item.view = button;
+    item.target = self;
```
ckerr, line 77: This PR moves every toolbar item's target from `Controller` to `MainToolbarController`. `-[ButtonToolbarItem menuFormRepresentation]` copies that target into the overflow menu item.

On main, `-[Controller validateMenuItem:]` validates those overflow items. It disables Open and Open Address while a sheet is attached. It disables Remove when nothing is selected. It sets the Show/Hide titles for Inspector and Filter Bar. `MainToolbarController` does not implement `validateMenuItem:`, so the overflow items may now stay enabled with fixed titles.

To check, narrow the window until Remove moves into the overflow menu, select nothing, and open the menu. Remove should be disabled.

If it is enabled, either forward `validateMenuItem:` to the delegate or keep `Controller` as the items' target.

macosx/MainToolbarController.mm
```diff
@@ -0,0 +1,443 @@
...
+        ((NSButtonCell*)((NSButton*)item.view).cell).showsStateBy = NSContentsCellMask; // blue when enabled
+
+        item.label = NSLocalizedString(@"Inspector", "Inspector toolbar item -> label");
+        item.paletteLabel = NSLocalizedString(@"Toggle Inspector", "Inspector toolbar item -> palette label");
+        item.toolTip = NSLocalizedString(@"Toggle the torrent inspector", "Inspector toolbar item -> tooltip");
+        item.image = [NSImage imageWithSystemSymbolName:@"info.circle" accessibilityDescription:nil];
+        item.action = @selector(showInfo:);
+
+        return item;
+    } else if ([ident isEqualToString:ToolbarItemIdentifierPauseResumeAll]) {
```
ckerr, line 139: (opus 5, extra) This branch and `PauseResumeSelected` below are 42 near-identical lines, differing only in four symbol names, six localized strings, the identifier pair and the action — including a verbatim second copy of the macOS 13.2 palette-width workaround and its two `setWidth:forSegment:` calls. Extract one helper both branches call, so the next fix to segmented-group setup lands in one place.

macosx/MainToolbarController.mm
```diff
@@ -0,0 +1,443 @@
...
+        item.label = NSLocalizedString(@"Share", "Share toolbar item -> label");
+        item.paletteLabel = NSLocalizedString(@"Share", "Share toolbar item -> palette label");
+        item.toolTip = NSLocalizedString(@"Share torrent file", "Share toolbar item -> tooltip");
+        item.image = [NSImage imageNamed:NSImageNameShareTemplate];
+        item.visibilityPriority = NSToolbarItemVisibilityPriorityLow;
+
+        NSButton* itemButton = (NSButton*)item.view;
+        itemButton.target = self;
+        itemButton.action = @selector(showToolbarShare:);
+        [itemButton sendActionOn:NSEventMaskLeftMouseDown];
```
ckerr, line 258: (opus 5, extra) Line 77 already set this. `NSToolbarItem.target` forwards to the item's view when the view responds to `setTarget:`, so the button's target is assigned by the shared factory.

```suggestion
        NSButton* itemButton = (NSButton*)item.view;
        itemButton.action = @selector(showToolbarShare:);
        [itemButton sendActionOn:NSEventMaskLeftMouseDown];
```

macosx/MainToolbarController.mm
```diff
@@ -0,0 +1,443 @@
...
+                                                                             ((NSControl*)sender).tag;
+    switch (tagValue) {
+    case ToolbarGroupTagPause:
+        [self.delegate mainToolbarStopSelectedTorrents:sender];
+        break;
+    case ToolbarGroupTagResume:
+        [self.delegate mainToolbarResumeSelectedTorrents:sender];
+        break;
+    }
+}
```
ckerr, line 292: (opus 5, extra) `NSMenuItem` is not an `NSControl`, and `-[GroupToolbarItem createMenu:]` sends exactly that from the overflow submenu — the cast works only because both classes independently declare `tag`. `selectedTag` is also -1 when no segment is selected, and neither switch has a `default:`.

```suggestion
// Sender is the group's NSSegmentedControl, or an NSMenuItem from the overflow submenu built by -[GroupToolbarItem createMenu:].
- (NSInteger)groupTagForSender:(id)sender
{
    if ([sender isKindOfClass:[NSSegmentedControl class]]) {
        return ((NSSegmentedControl*)sender).selectedTag;
    }
    if ([sender isKindOfClass:[NSMenuItem class]]) {
        return ((NSMenuItem*)sender).tag;
    }
    return -1;
}

- (void)allToolbarClicked:(id)sender
{
    switch ([self groupTagForSender:sender]) {
    case ToolbarGroupTagPause:
        [self.delegate mainToolbarStopAllTorrents:sender];
        break;
    case ToolbarGroupTagResume:
        [self.delegate mainToolbarResumeAllTorrents:sender];
        break;
    }
}

- (void)selectedToolbarClicked:(id)sender
{
    switch ([self groupTagForSender:sender]) {
    case ToolbarGroupTagPause:
        [self.delegate mainToolbarStopSelectedTorrents:sender];
        break;
    case ToolbarGroupTagResume:
        [self.delegate mainToolbarResumeSelectedTorrents:sender];
        break;
    }
}
```

macosx/MainToolbarController.mm
```diff
@@ -0,0 +1,443 @@
...
+            if (!torrent.active && !torrent.waitingToStart && !torrent.finishedSeeding) {
+                return YES;
+            }
+        }
+        return NO;
+    }
+
+    // enable pause item
+    if ([ident isEqualToString:ToolbarItemIdentifierPauseSelected]) {
+        for (Torrent* torrent in [self.delegate mainToolbarSelectedTorrents]) {
```
ckerr, line 359: (opus 5, extra) `-[TorrentTableView selectedTorrents]` allocates and fills a new array on every call, and `-[GroupToolbarItem validate]` calls `validateToolbarItem:` once per subitem — so the Apply Selected group rebuilds this array twice per validation pass, and validation runs continuously. Snapshot it once in `GroupToolbarItem.validate` and pass it down.

macosx/Controller.mm
```diff
@@ -3637,351 +3610,94 @@ - (NSWindow*)sharingService:(NSSharingService*)sharingService
...
+- (void)mainToolbarStopSelectedTorrents:(id)sender
+{
+    [self stopSelectedTorrents:sender];
+}
 
-    return YES;
+- (void)mainToolbarResumeSelectedTorrents:(id)sender
+{
+    [self resumeSelectedTorrents:sender];
 }
```
ckerr, line 3701: (opus 5, extra) A toolbar click takes three hops to rename one selector: `-[MainToolbarController createFile:]` → `mainToolbarCreateFile:` → `-[Controller createFile:]`. These eight pass-throughs plus the eight shims at MainToolbarController.mm 403-441 are ~90 lines that do nothing but translate names.

Declare the protocol's action methods with the names Controller already uses — `createFile:`, `openShowSheet:`, `openURLShowSheet:`, `removeNoDelete:`, `showInfo:`, `toggleFilterBar:`, `toggleQuickLook:`, `showToolbarShare:` — and Controller's existing IBActions satisfy the protocol directly. This whole block goes away with no behavior change.

macosx/MainToolbarController.mm
```diff
@@ -0,0 +1,443 @@
+// This file Copyright © Transmission authors and contributors.
```
ckerr: (opus 5, extra) Commit messages: `macos: Extract mac toolbar controller logic` puts a scope where CLAUDE.md requires a Conventional Commits type (`refactor(macos): …`), and `Apply codestyle fix` has no type at all and belongs squashed into the first. That commit's body also opens by restating its own subject, which CLAUDE.md rules out.

nevack:
@ckerr I'm inclined to ignore review as this was meant to be a simple refactor to move some code from Controller to dedicated class without behaviour changes.
Is it okay with you?

ckerr:
@nevack I've marked all the pre-existing issues as resolved, agree that it doesn't make sense for a move-only PR.

There's one remaining finding that appears to be new to this PR, so I've left that thread open

ckerr:
@nevack also, sorry for the delay in getting back to this PR, I've been traveling for work and didn't have much free time the last week or so.

PR #316, opened 2026-09-08
fix: set limit for rpc request body size

tearfur:
## Description

Add a maximum request body size limit to the RPC server. Had a look at all the RPC methods, and the largest request body I can spot is torrent metainfo in `torrent_add` requests.

It shares the constant with the web client as both values are set based on the size of torrent files.

Notes: Improved security of the RPC server.

## Checklist

### DIFF
--- a/libtransmission/constants.h
+++ b/libtransmission/constants.h
@@ -29,5 +29,6 @@ inline auto constexpr TrHttpServerRpcRelativePath = std::string_view{ "rpc" };
 inline auto constexpr TrHttpServerWebRelativePath = std::string_view{ "web/" };
 inline auto constexpr TrRpcSessionIdHeader = std::string_view{ TR_PROJ_SHARED_RPC_SESSION_ID_HEADER };
 inline auto constexpr TrRpcVersionHeader = std::string_view{ TR_PROJ_SHARED_RPC_VERSION_HEADER };
+inline auto constexpr TrWebMaxBodyBytes = 10U * 1024U * 1024U;
 
 inline auto constexpr TrBlockSize = uint32_t{ 1024U * 16U };
--- a/libtransmission/rpc-server.cc
+++ b/libtransmission/rpc-server.cc
@@ -617,6 +617,9 @@ auto constexpr ServerStartRetryCount = 10;
 auto constexpr ServerStartRetryDelayIncrement = 5s;
 auto constexpr ServerStartRetryMaxDelay = 60s;
 
+// Arbitrary limit to prevent DoS attacks
+auto constexpr ServerMaxRequestBodyBytes = 10 * 1024 * 1024;
+
 bool bindUnixSocket(
     [[maybe_unused]] struct event_base* base,
     [[maybe_unused]] struct evhttp* httpd,
@@ -767,6 +770,13 @@ void start_server(tr_rpc_server* server)
                 fmt::arg("count", ServerStartRetryCount)));
     } else {
         evhttp_set_gencb(httpd, handle_request, server);
+
+        // N.B. https://github.com/libevent/libevent/issues/321
+        // Some browsers cannot handle HTTP 413 responses unless the server reads
+        // the entire request, which is what EVHTTP_SERVER_LINGERING_CLOSE does.
+        evhttp_set_max_body_size(httpd, ServerMaxRequestBodyBytes);
+        evhttp_set_flags(httpd, EVHTTP_SERVER_LINGERING_CLOSE);
+
         server->httpd.reset(httpd);
 
         tr_logAddInfo(
--- a/libtransmission/rpc-server.cc
+++ b/libtransmission/rpc-server.cc
@@ -34,6 +34,7 @@
 #define ZLIB_CONST
 #include <zlib.h>
 
+#include "libtransmission/constants.h"
 #include "libtransmission/crypto-utils.h" /* tr_ssha1_matches() */
 #include "libtransmission/error.h"
 #include "libtransmission/file-utils.h"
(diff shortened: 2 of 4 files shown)


tests/libtransmission/web-test.cc
```diff
@@ -16,6 +16,7 @@
 
 #include <fmt/format.h>
 
+#include <libtransmission/constants.h> // tr_base64_encode()
```
ckerr, line 19: (Fable 5.1, high) Comment is copied from the line below; this header supplies `TrWebMaxBodyBytes`.
```suggestion
#include <libtransmission/constants.h> // TrWebMaxBodyBytes
```

tests/libtransmission/web-test.cc
```diff
@@ -337,20 +338,20 @@ TEST_F(WebTest, responseBodyLimitReportsCurlError)
...
     ASSERT_TRUE(response.errmsg);
     EXPECT_FALSE(std::empty(*response.errmsg));
-    EXPECT_LE(std::size(response.body), tr_web::FetchOptions::DefaultMaxFileSize);
+    EXPECT_LE(std::size(response.body), TrWebMaxBodyBytes);
 }
 
 TEST_F(WebTest, zeroMaxFileSizeDisablesLimit)
 {
-    auto body = std::string(tr_web::FetchOptions::DefaultMaxFileSize + 1U, 'x');
+    auto body = std::string(TrWebMaxBodyBytes, 'x');
```
ckerr, line 354: (Fable 5.1, high) The `+ 1U` was dropped, so the body no longer exceeds the default limit and the test passes even if `max_file_size = 0` does nothing. (This test change also rides in the `refactor:` commit rather than the `fix:` one.)
```suggestion
    auto body = std::string(TrWebMaxBodyBytes + 1U, 'x');
```

libtransmission/constants.h
```diff
@@ -29,5 +29,6 @@ inline auto constexpr TrHttpServerRpcRelativePath = std::string_view{ "rpc" };
 inline auto constexpr TrHttpServerWebRelativePath = std::string_view{ "web/" };
 inline auto constexpr TrRpcSessionIdHeader = std::string_view{ TR_PROJ_SHARED_RPC_SESSION_ID_HEADER };
 inline auto constexpr TrRpcVersionHeader = std::string_view{ TR_PROJ_SHARED_RPC_VERSION_HEADER };
+inline auto constexpr TrWebMaxBodyBytes = 10U * 1024U * 1024U;
```
ckerr, line 32: (Fable 5.1, high) Base64 inflates `torrent-add` metainfo by 4/3 plus a JSON envelope, so this caps uploaded .torrent files at ~7.5 MiB while the same file fetched by URL passes at 10 MiB. transmission-remote, the Qt client and the web UI all upload via `metainfo`, so adds in that band regress. Give the RPC server its own derived limit and leave tr_web's default alone (raising this one value also widens tracker/blocklist response limits):
```suggestion
inline auto constexpr TrWebMaxBodyBytes = 10U * 1024U * 1024U;
// RPC request bodies carry base64 metainfo (4/3 inflation) inside a JSON envelope.
inline auto constexpr TrRpcMaxBodyBytes = TrWebMaxBodyBytes / 3U * 4U + 64U * 1024U;
```

libtransmission/rpc-server.cc
```diff
@@ -767,6 +768,13 @@ void start_server(tr_rpc_server* server)
                 fmt::arg("count", ServerStartRetryCount)));
     } else {
         evhttp_set_gencb(httpd, handle_request, server);
+
+        // N.B. https://github.com/libevent/libevent/issues/321
+        // Some browsers cannot handle HTTP 413 responses unless the server reads
+        // the entire request, which is what EVHTTP_SERVER_LINGERING_CLOSE does.
+        evhttp_set_max_body_size(httpd, static_cast<ev_ssize_t>(TrWebMaxBodyBytes));
```
ckerr, line 775: (Fable 5.1, high) Use the RPC-specific limit suggested in constants.h:
```suggestion
        evhttp_set_max_body_size(httpd, static_cast<ev_ssize_t>(TrRpcMaxBodyBytes));
```

ckerr, line 775: (Fable 5.1, high) The web UI does not survive this 413: `web/src/remote.js` only special-cases 409/204, so `response.json()` throws on the HTML body, the catch shows "Connection failed" and disables periodic session refresh until reload. Handle 413 in `remote.js` (and/or check `file.size` in `open-dialog.js` before encoding).

ckerr, line 775: (Fable 5.1, high) libevent answers oversized requests itself (`evhttp_connection_incoming_fail` nulls `req->uri`, so `evhttp_handle_request` sends the error page before the gencb). `handle_request` never runs: rejected uploads are not logged, and the 413 lacks the `Server`, `X-Frame-Options` and CSP headers every other response gets. Worth a log line somewhere if operators are expected to diagnose failed adds.

libtransmission/rpc-server.cc
```diff
@@ -767,6 +768,13 @@ void start_server(tr_rpc_server* server)
...
+        evhttp_set_flags(httpd, EVHTTP_SERVER_LINGERING_CLOSE);
```
ckerr, line 776: (Fable 5.1, high) Two cases lingering close does not cover, for the comment above or a follow-up:
- Chunked bodies over the limit take `evhttp_connection_fail_` directly (http.c `DATA_TOO_LONG` branch): the connection is closed mid-upload and the client typically sees a reset, not the 413.
- `Expect: 100-continue` (libcurl's default for POSTs over 1 MiB, i.e. transmission-remote and the Qt/GTK clients): libevent neither sends 100 nor 413 until the whole body is uploaded and drained. Disabling `Expect` in tr_web's RPC client path avoids the wasted upload.

ckerr, line 776: (Fable 5.1, high) `evhttp_set_flags` and `EVHTTP_SERVER_LINGERING_CLOSE` first appear in libevent 2.1.6-beta, but `CMakeLists.txt` declares `EVENT2_MINIMUM 2.1.0`, so a system libevent 2.1.0-2.1.5 no longer compiles. Raise the minimum, or guard this call the way line 499 guards its evhttp API.

libtransmission/constants.h
tearfur: Tens of MiB of RAM should be a breeze for any devices these days, so I don't feel a need to fine tune the size limit for each application like you suggested. If this value is not enough somewhere, we should raise it and call it a day.

You know what, I'll close this PR and open a new one where this limit is configurable in `settings.json`.

PR #323, opened 2026-09-11
fix: Glib::ustring serializer

ckerr:
## Description

While I was reviewing 321, Fable found an issue with the Glib::ustring serializer. This PR fixes it and adds simple scaffolding to add glib/gtk regression tests. 

Marking as `notes:none` since this bug isn't in any releases

## Checklist

### DIFF
--- a/gtk/CMakeLists.txt
+++ b/gtk/CMakeLists.txt
@@ -1,8 +1,10 @@
 find_program(APPSTREAM appstreamcli)
 
-add_executable(${TR_NAME}-gtk WIN32)
+# Everything but main() is a library so that tests/gtk can link the client's
+# code without launching it.
+add_library(${TR_NAME}-gtk-lib STATIC)
 
-target_sources(${TR_NAME}-gtk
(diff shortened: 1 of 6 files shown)


tearfur (commented):
I don't think any code needs a `Glib::ustring` serializer though? Otherwise there would have been a compile error.

ckerr:
Places where this bug exists:

- Actions.cc:55 stores the sort mode picked from the menu.
- PrefsDialog.cc:226 stores the text entries for blocklist URL, RPC username and RPC password.
- PrefsDialog.cc:249 stores the default trackers text view

All three reach the serializer through the prefs setter, which takes its type from the argument. I had an agent run the real conversion on the string "alice" and this displayed the error:

```
stored as: Vector
contents: [ 97 108 105 99 101 ]
read back as string: fails (nullopt)
```

tearfur (approved):
LGTM.

Something for a future PR: Might be a good idea to copy `.clang-tidy` from another test directory too.

PR #324, opened 2026-09-11
chore: bump curl minimum version to 7.80

tearfur:
## Description

Notes: Bump minimum supported curl version to 7.80.

## Checklist


ckerr:
Not a blocker, but you mentioned Curl 7.77.0 earlier, and are at 7.80.0 in this PR. What's the reason for the difference?

tearfur:
I should have made myself clearer.

For the longest time, my target version *was* 7.77, but I also think the recent #292 is kinda important, so now I decided on 7.80.

CMakeLists.txt
```diff
@@ -44,7 +44,7 @@ include(GNUInstallDirs)
 include(TrMacros)
 
 set(CURL_MINIMUM_MAJOR 7)
-set(CURL_MINIMUM_MINOR 28)
+set(CURL_MINIMUM_MINOR 80)
```
ckerr, line 47: Record why 7.80. `ARCHIVE_MINIMUM` below carries the same kind of note.

```suggestion
# 7.80 is where CURLOPT_MAXLIFETIME_CONN landed, the newest option we set unconditionally.
set(CURL_MINIMUM_MAJOR 7)
set(CURL_MINIMUM_MINOR 80)
```

ckerr, line 47: 7.80 drops RHEL/Rocky/Alma 9, which ship curl 7.76.1 and are supported to 2032. Our build docs cover only Debian 12/13, and CI is ubuntu-latest plus alpine, so nothing here catches it.

A 7.61 floor would remove every guard but `MAXLIFETIME_CONN` and keep EL9 building, at the cost of keeping the announcer-http workaround, which needs 7.77. Your call, but let's state it.

libtransmission/web.cc
```diff
@@ -617,9 +610,7 @@ class tr_web::Impl
         (void)curl_easy_setopt(e, CURLOPT_AUTOREFERER, 1L);
         (void)curl_easy_setopt(e, CURLOPT_ACCEPT_ENCODING, "");
         (void)curl_easy_setopt(e, CURLOPT_FOLLOWLOCATION, 1L);
-#if LIBCURL_VERSION_NUM >= 0x075000 /* 7.80.0 */
         (void)curl_easy_setopt(e, CURLOPT_MAXLIFETIME_CONN, MaxlifetimeConn);
```
ckerr: Nothing enforces 7.80 at link time. `curl_multi_wait` (7.28) is the newest function we import, so a 7.79 libcurl loads and drops this option into `CURLE_UNKNOWN_OPTION`, discarded by the cast. #319's warning is the only cover, so the two should land together.

ckerr (approved):

CMakeLists.txt
tearfur: Ack, I always forget about RHEL and its derivatives when dealing with dependency versions.

There's not much reason to raise the curl version floor then.

PR #326, opened 2026-09-12
ci: add AlmaLinux

tearfur:
## Description

Transmission has a user base on RHEL family distributions, but we never had CI runners on them.

This has lead to us (mostly me) using library features that were too new for the versions distributed on RHEL:
- https://github.com/transmission/transmission/issues/7624
- Nearly did it again #324

This PR adds AlmaLinux CI to prevent issues like these.

## Checklist

### DIFF
--- a/.github/actions/install-deps/action.yml
+++ b/.github/actions/install-deps/action.yml
@@ -43,43 +43,53 @@ runs:
 
         if [[ "$PKG_FAMILY" == "dnf" ]]; then
           # Fedora/DNF-based installation
-          dnf install -y \
-            cmake \
-            gettext \
-            ninja-build \
-            pkgconf-pkg-config \
+          BASE_PACKAGES=(
+            cmake
+            gettext
+            ninja-build
+            pkgconf-pkg-config
             xz
+          )
 
           # Transmission-specific libraries with Fedora names
-          dnf install -y \
-            fast_float-devel \
-            fmt-devel \
-            libarchive-devel \
-            libcurl-devel \
-            libevent-devel \
-            libnatpmp-devel \
-            libpsl-devel \
-            miniupnpc-devel \
-            openssl-devel \
-            simdutf-devel \
+          BASE_PACKAGES+=(
+            fast_float-devel
+            fmt-devel
+            libarchive-devel
+            libcurl-devel
+            libevent-devel
+            libnatpmp-devel
+            libpsl-devel
+            miniupnpc-devel
+            openssl-devel
             zlib-devel
+          )
 
           # Compiler packages
           if [[ "${{ inputs.compiler }}" == "clang" ]]; then
-            dnf install -y clang
+            BASE_PACKAGES+=(clang)
           elif [[ "${{ inputs.compiler }}" == "gcc" ]]; then
-            dnf install -y gcc-c++
+            BASE_PACKAGES+=(gcc-c++)
           fi
 
           # Faster linker
           if [[ "${{ inputs.enable-mold }}" == "true" ]]; then
-            dnf install -y mold
+            BASE_PACKAGES+=(mold)
           fi
 
           # Compiler cache
           if [[ "${{ inputs.enable-ccache }}" == "true" ]]; then
-            dnf install -y ccache
+            BASE_PACKAGES+=(ccache)
           fi
+
+          if [ "$DISTRO" = 'fedora' ]; then
+            BASE_PACKAGES+=(simdutf-devel)
+          else
+            dnf install -y epel-release
+            dnf config-manager --set-enabled crb
+          fi
+
+          dnf install -y "${BASE_PACKAGES[@]}"
 
         elif [[ "$PKG_FAMILY" == "apt" ]]; then
           # Debian/Ubuntu/APT-based installation
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -1136,16 +1136,20 @@ jobs:
         if: always()
         run: ccache --show-stats
 
-  fedora:
+  rhel-family:
     needs: [ make-source-tarball, what-to-make ]
     if: ${{ needs.what-to-make.outputs.make-cli == 'true' || needs.what-to-make.outputs.make-daemon == 'true' || needs.what-to-make.outputs.make-gtk == 'true' || needs.what-to-make.outputs.make-qt == 'true' || needs.what-to-make.outputs.make-tests == 'true' || needs.what-to-make.outputs.make-utils == 'true' }}
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
-        version: [ latest ]
+        include:
+          - docker-image: fedora
+            version: latest
+          - docker-image: almalinux
+            version: latest
     container:
-      image: fedora:${{ matrix.version }}
+      image: ${{matrix.docker-image}}:${{ matrix.version }}
     steps:
       - name: Show Configuration
         run: |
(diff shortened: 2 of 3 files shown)


ckerr:
@tearfur please don't merge until I've had a chance to take a look. Getting all the CI cache under the GH free account limit was *difficult* and I want to make sure that if we add another distro that we don't break that :eyes:

tearfur:
Sure thing.

.github/actions/install-deps/action.yml
```diff
@@ -43,44 +43,54 @@ runs:
 
         if [[ "$PKG_FAMILY" == "dnf" ]]; then
           # Fedora/DNF-based installation
```
ckerr: AlmaLinux uses this branch now too. Please update this comment and the "Fedora names" one on line 54.

```suggestion
          # Fedora and RHEL-family (DNF) installation
```

.github/workflows/actions.yml
```diff
@@ -1136,16 +1136,20 @@ jobs:
...
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
-        version: [ latest ]
+        include:
+          - docker-image: fedora
+            version: latest
+          - docker-image: almalinux
+            version: latest
```
ckerr: Both legs have `version: latest`, so both use the ccache key `rhel-family-latest-cache-v2`. On main, only the first leg to finish saves its cache, so AlmaLinux starts cold every run. Please set `key-suffix` on line 1178 to `${{ matrix.docker-image }}-${{ matrix.version }}`.

.github/workflows/actions.yml
```diff
@@ -1200,11 +1204,106 @@ jobs:
             -DRUN_CLANG_TIDY=OFF \
             -DCMAKE_EXE_LINKER_FLAGS=-fuse-ld=mold \
             -DUSE_SYSTEM_DEFAULT=ON \
```
ckerr: Main now requires fmt 11.1.0 or newer (#361). AlmaLinux 10 has 11.0.2, so this leg will fail at configure after a rebase. Please add `USE_SYSTEM_FMT=AUTO` like the other jobs on main.

```suggestion
            -DUSE_SYSTEM_DEFAULT=ON \
            -DUSE_SYSTEM_FMT=AUTO `# distro may not have fmt >= FMT_MINIMUM` \
```

.github/workflows/actions.yml
```diff
@@ -1200,11 +1204,106 @@ jobs:
...
-            -DUSE_SYSTEM_DHT=OFF `# Not packaged in Fedora` \
-            -DUSE_SYSTEM_SIGSLOT=OFF `# Not packaged in Fedora` \
-            -DUSE_SYSTEM_SMALL=OFF `# Not packaged in Fedora` \
-            -DUSE_SYSTEM_UTP=OFF `# Not packaged in Fedora` \
-            -DUSE_SYSTEM_WIDE_INTEGER=OFF `# Not packaged in Fedora`
+            -DUSE_SYSTEM_SIMDUTF=AUTO `# Available on Fedora but not on RHEL/EPEL` \
```
ckerr: `AUTO` lets Fedora silently fall back to the bundled simdutf. Please keep it required on Fedora.

```suggestion
            -DUSE_SYSTEM_SIMDUTF=${{ matrix.docker-image == 'fedora' && 'ON' || 'OFF' }} `# Not packaged in RHEL/EPEL` \
```

.github/workflows/actions.yml
```diff
@@ -1200,11 +1204,106 @@ jobs:
...
+        if: ${{ needs.what-to-make.outputs.make-tests == 'true' }}
+        uses: ./.github/actions/run-tests
+        with:
+          build-dir: obj
+          build-config: Release
+      - name: Install
+        run: cmake --install obj --config Release --strip
+      - uses: actions/upload-artifact@v6
+        with:
+          name: binaries-${{ github.job }}-${{ matrix.version }}
```
ckerr: Both legs upload `binaries-rhel-family-latest`, and this run has two artifacts with that name. Let's add the image name to avoid collision:

```suggestion
          name: binaries-${{ github.job }}-${{ matrix.docker-image }}-${{ matrix.version }}
```

.github/workflows/actions.yml
```diff
@@ -1200,11 +1204,106 @@ jobs:
...
+        uses: ./.github/actions/finalize-ccache
+        with:
+          trim-age: 3d
+      - name: Show ccache stats
+        if: always()
+        run: ccache --show-stats
+
+  # RHEL/AlmaLinux 9's compiler and library versions are very awkward to fit in rhel-family,
+  # so split it into a separate job, and once it goes EOL, drop this job and keep
+  # rhel-family only. https://wiki.almalinux.org/release-notes/
```
ckerr: Please name the actual constraints so we know when this job can rejoin the matrix.

```suggestion
  # AlmaLinux 9 needs its own settings.
  # Its GCC 11 is older than our GCC 12 minimum, so it builds with clang.
  # EPEL 9 has no simdutf, and its fmt is older than FMT_MINIMUM.
  # Drop this job when AlmaLinux 9 reaches EOL: https://wiki.almalinux.org/release-notes/
```

.github/workflows/actions.yml
```diff
@@ -1200,11 +1204,106 @@ jobs:
...
+  almalinux-9:
```
ckerr: This job duplicates about 90 lines of `rhel-family`. Please consider making it a third matrix entry. The image, compiler, GTK version, and fmt/simdutf flags can be include fields, like `debian` does with `matrix.gtk`. Dropping it at EOL then becomes a three-line change.

.github/workflows/actions.yml
```diff
@@ -1200,11 +1204,106 @@ jobs:
...
+    container:
+      image: almalinux:9
+    steps:
+      - name: Show Configuration
+        run: |
+          echo '${{ toJSON(needs) }}'
+          echo '${{ toJSON(runner) }}'
+          cat /etc/os-release
+      - name: Get Composite Actions
+        uses: actions/checkout@v6
```
ckerr: Main now uses `checkout@v7`, `download-artifact@v8`, and `upload-artifact@v7`. Please rebase and match them here and on lines 1271 and 1317.

```suggestion
        uses: actions/checkout@v7
```

.github/workflows/actions.yml
```diff
@@ -1200,11 +1204,106 @@ jobs:
...
+          enable-gtk: ${{ needs.what-to-make.outputs.make-gtk == 'true' && 'gtk3' || 'false' }}
+          enable-qt: ${{ needs.what-to-make.outputs.make-qt == 'true' && 'qt6' || 'false' }}
+          use-sudo: false
+          enable-mold: true
+          enable-ccache: true
+      - name: Setup ccache
+        uses: ./.github/actions/setup-ccache
+        with:
+          in-container: 'true'
+          key-suffix: ${{ matrix.version }}
```
ckerr: This job has no matrix, so `matrix.version` is empty. The suffix does nothing, and the artifact on line 1319 is named `binaries-almalinux-9-`. Please replace the suffix with `max-size`. Your cold build used 0.02 GB, and `.github/CACHE_POLICY.md` asks for about twice the cold-build size. Please also name the artifact `binaries-${{ github.job }}`.

```suggestion
          max-size: 60M
```

.github/workflows/actions.yml
```diff
@@ -1200,11 +1204,106 @@ jobs:
...
+            -DENABLE_QT=${{ (needs.what-to-make.outputs.make-qt == 'true') && 'ON' || 'OFF' }} \
+            -DENABLE_TESTS=${{ (needs.what-to-make.outputs.make-tests == 'true') && 'ON' || 'OFF' }} \
+            -DENABLE_UTILS=${{ (needs.what-to-make.outputs.make-utils == 'true') && 'ON' || 'OFF' }} \
+            -DREBUILD_WEB=OFF \
+            -DENABLE_WERROR=OFF \
+            -DRUN_CLANG_TIDY=OFF \
+            -DCMAKE_EXE_LINKER_FLAGS=-fuse-ld=mold \
+            -DUSE_SYSTEM_DEFAULT=ON \
+            -DUSE_SYSTEM_DHT=OFF `# Not packaged in RHEL` \
+            -DUSE_SYSTEM_FMT=OFF `# FTBFS on clang` \
```
ckerr: On main, this is off because EPEL 9's fmt 8.1.1 is older than `FMT_MINIMUM`, not because of clang. Please use the same flag and comment as the other jobs.

```suggestion
            -DUSE_SYSTEM_FMT=AUTO `# distro may not have fmt >= FMT_MINIMUM` \
```

ckerr (changes requested):
Several little change requests but according to Claude it will still fit in our cache limits 👍

.github/actions/install-deps/action.yml
tearfur: 2611f831145acf3e59da3158b7b905a099b0f745

.github/workflows/actions.yml
tearfur: b3d85141da9834cd34b4e8a4171bcf5badecc760

.github/workflows/actions.yml
tearfur: 365ec728cf2ac356da3cbcaf99a4a0a6610079aa

.github/workflows/actions.yml
tearfur: 4bc6d2ca483a636096c7eb053cf88353c7891528

.github/workflows/actions.yml
tearfur: a16620d20684f934b55e5928545a11d3e6991e47

.github/workflows/actions.yml
tearfur: e959c971ee9ac6c159bd98eb91292b20db604d45

.github/workflows/actions.yml
tearfur: e959c971ee9ac6c159bd98eb91292b20db604d45

.github/workflows/actions.yml
tearfur: Moot because of e959c971ee9ac6c159bd98eb91292b20db604d45.

.github/workflows/actions.yml
tearfur: Moot because of e959c971ee9ac6c159bd98eb91292b20db604d45.

.github/workflows/actions.yml
tearfur: Moot because of e959c971ee9ac6c159bd98eb91292b20db604d45.

ckerr (approved):

PR #329, opened 2026-09-12
refactor: move macOS RPC handling out of Controller

nevack:


macosx/RpcController.mm
```diff
@@ -0,0 +1,140 @@
...
+            return TR_RPC_NOREMOVE; // We'll do the remove manually.
+        });
+    }
+
+    return self;
+}
+
+- (void)rpcCallback:(tr_rpc_callback_type)type forTorrentId:(std::optional<tr_torrent_id_t>)torrent_id
+{
+    @autoreleasepool {
```
lolgear, line 41: This autorelease pool is useless.
You can freely remove it.

It only contains dispatch to queue and block.
So, dispatch_async will put this block on mainQueue and finish.
No objects are added to autorelease pool.

Not sure about an intention of original code.
All handlers ( `addTorrentWithStruct`, `startedStoppedTorrent` etc. ) should be added to autorelease pool because they create many objects, right? Hmmm...

macosx/RpcController.mm
```diff
@@ -0,0 +1,140 @@
...
+- (instancetype)initWithLib:(tr_session*)lib delegate:(id<RpcControllerDelegate>)delegate
+{
+    if ((self = [super init])) {
+        _fLib = lib;
+        _delegate = delegate;
+
+        tr_sessionSetRPCCallback(_fLib, [controller = self](tr_rpc_callback_type const type, std::optional<tr_torrent_id_t> const torrent_id) {
+            [controller rpcCallback:type forTorrentId:torrent_id];
+            return TR_RPC_NOREMOVE; // We'll do the remove manually.
+        });
```
ckerr, line 33: The lambda captures `self` strongly, so the session owns this object and `Controller.fRpcController` owns nothing. Capture weakly and let the property be the owner:

```suggestion
        __weak RpcController* weakSelf = self;
        tr_sessionSetRPCCallback(_fLib, [weakSelf](tr_rpc_callback_type const type, std::optional<tr_torrent_id_t> const torrent_id) {
            [weakSelf rpcCallback:type forTorrentId:torrent_id];
            return TR_RPC_NOREMOVE; // We'll do the remove manually.
        });
```

macosx/RpcController.mm
```diff
@@ -0,0 +1,140 @@
...
+                if (!torrent) {
+                    NSLog(@"No torrent found matching the given torrent id from the RPC callback!");
+                    return;
+                }
+            }
+
+            switch (type) {
+            case TR_RPC_TORRENT_ADDED:
+                if (torrent_id.has_value()) {
+                    if (auto* const torrentStruct = tr_torrentFindFromId(self.fLib, *torrent_id); torrentStruct != nullptr) {
```
ckerr, line 56: A block queued during quit can run after `tr_sessionClose()` and pass a freed session to `tr_torrentFindFromId()`. The old code had the same hazard; this class is the place to fix it.

Add an `-invalidate` that sets `_fLib = nullptr`, call it from `Controller` just before `tr_sessionClose()`, and return early from this block when `fLib` is null. Both run on the main thread, so no lock is needed. Do not clear the callback instead: `tr_sessionSetRPCCallback()` is unsynchronized, and the session thread may be invoking it.

macosx/RpcController.h
```diff
@@ -0,0 +1,30 @@
...
+@class Torrent;
+
+@protocol RpcControllerDelegate<NSObject>
+
+- (Torrent*)rpcController:(RpcController*)controller torrentForId:(tr_torrent_id_t)torrent_id;
+- (void)rpcController:(RpcController*)controller addTorrent:(Torrent*)torrent;
+- (void)rpcController:(RpcController*)controller removeTorrent:(Torrent*)torrent deleteData:(BOOL)deleteData;
+- (void)rpcControllerDidStartOrStopTorrent:(RpcController*)controller;
+- (void)rpcControllerDidChangeTorrent:(RpcController*)controller torrent:(Torrent*)torrent;
+- (void)rpcControllerDidMoveTorrent:(RpcController*)controller torrent:(Torrent*)torrent;
```
ckerr, line 19: Put the sender first, as `rpcController:addTorrent:` above does, and pass the torrent in all three. `DidStartOrStop` drops an argument the old `rpcStartedStoppedTorrent:` had.

```suggestion
- (void)rpcController:(RpcController*)controller didStartOrStopTorrent:(Torrent*)torrent;
- (void)rpcController:(RpcController*)controller didChangeTorrent:(Torrent*)torrent;
- (void)rpcController:(RpcController*)controller didMoveTorrent:(Torrent*)torrent;
```

The call sites in `RpcController.mm` and the implementations in `Controller.mm` need the matching rename.

macosx/RpcController.h
```diff
@@ -0,0 +1,30 @@
...
+#import <Foundation/Foundation.h>
+
+#include <libtransmission/transmission.h>
+
+@class RpcController;
+@class Torrent;
+
+@protocol RpcControllerDelegate<NSObject>
+
+- (Torrent*)rpcController:(RpcController*)controller torrentForId:(tr_torrent_id_t)torrent_id;
```
ckerr, line 14: This method makes the delegate a data source: `RpcController` builds `Torrent` objects yet must ask `Controller` to find them. Adding a torrent is split across the seam, and five of the nine delegate methods only forward.

Choose one side. Either `RpcController` only hops to the main thread and hands `Controller` the event type and torrent id, or it holds the torrent list and needs no lookup.

macosx/RpcController.mm
```diff
@@ -0,0 +1,140 @@
...
+            case TR_RPC_TORRENT_CHANGED:
+                [self changedTorrent:torrent];
+                break;
+
+            case TR_RPC_TORRENT_MOVED:
+                [self movedTorrent:torrent];
+                break;
+
+            case TR_RPC_SESSION_QUEUE_POSITIONS_CHANGED:
+                [self.delegate rpcControllerDidUpdateQueue:self];
```
ckerr, line 84: The other torrent events call `[torrent update]` here before notifying the delegate. This one leaves the refresh (`[Torrent updateTorrents:]`) to `Controller`. Refresh the model on one side of the delegate, not both.

macosx/RpcController.mm
```diff
@@ -0,0 +1,140 @@
...
+{
+    NSString* location = tr_strv_to_utf8_nsstring(tr_torrentGetDownloadDir(torrentStruct));
+
+    Torrent* torrent = [[Torrent alloc] initWithTorrentStruct:torrentStruct location:location lib:self.fLib];
+
+    // Change the location if the group calls for it (this has to wait until after the torrent is created).
+    if ([GroupsController.groups usesCustomDownloadLocationForIndex:torrent.groupValue]) {
+        location = [GroupsController.groups customDownloadLocationForIndex:torrent.groupValue];
+        [torrent changeDownloadFolderBeforeUsing:location determinationType:TorrentDeterminationAutomatic];
+    }
```
ckerr, line 112: This block also lives at `Controller.mm` in `openFiles` and `openMagnet`, and in `Torrent.mm`. Moving one copy into a second class makes the set harder to keep in step. Extract a `Torrent` method, e.g. `-applyGroupDownloadLocation`, and call it from each site. A follow-up PR is fine.

macosx/RpcController.mm
```diff
@@ -0,0 +1,140 @@
...
+                    if (auto* const torrentStruct = tr_torrentFindFromId(self.fLib, *torrent_id); torrentStruct != nullptr) {
+                        [self addTorrentWithStruct:torrentStruct];
+                    }
+                }
+                break;
+
+            case TR_RPC_TORRENT_STARTED:
+            case TR_RPC_TORRENT_STOPPED:
+                [self startedStoppedTorrent:torrent];
+                break;
```
ckerr, line 65: `startedStoppedTorrent:`, `changedTorrent:`, and `movedTorrent:` each have one caller and two lines. The remove, queue, and session cases are already inline. Inline these too and delete the helpers:

```suggestion
            case TR_RPC_TORRENT_STARTED:
            case TR_RPC_TORRENT_STOPPED:
                [torrent update];
                [self.delegate rpcControllerDidStartOrStopTorrent:self];
                break;
```

ckerr (commented):
...I reviewed this before noticing it was a draft. I didn't want to throw the work away, but feel free to ignore 😸

PR #331, opened 2026-09-12
ci: limit Xcode project changes to macOS jobs

nevack:
As one can see in https://github.com/retransmission/retransmission/pull/329/checks and https://github.com/retransmission/retransmission/pull/330/checks changes in xcode project caused to rerun all GHA checking buildability of cli and daemon, including windows/linux/bsd.

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -80,7 +80,8 @@ jobs:
           }
           get_changes core       CMakeLists.txt cmake third-party libtransmission libtransmission-app
           get_changes android    CMakeLists.txt cmake third-party libtransmission android
-          get_changes cli        CMakeLists.txt cmake Transmission.xcodeproj third-party libtransmission cli
+          get_changes cli        CMakeLists.txt cmake third-party libtransmission cli
+          get_changes daemon     CMakeLists.txt cmake third-party libtransmission daemon
           get_changes any-code   CMakeLists.txt cmake Transmission.xcodeproj libtransmission libtransmission-app cli daemon gtk macosx qt utils tests web third-party
           get_changes our-code   CMakeLists.txt cmake Transmission.xcodeproj libtransmission libtransmission-app cli daemon gtk macosx qt utils tests web
           get_changes tidy-config libtransmission/.clang-tidy libtransmission-app/.clang-tidy gtk/.clang-tidy qt/.clang-tidy tests/libtransmission/.clang-tidy tests/qt/.clang-tidy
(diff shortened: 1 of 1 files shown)


lolgear:
I am thinking about total deconstruction of an Xcode project.

Whole daemon could be rewritten in Swift and managed by launchd.

And these cli binaries have no place in the upcoming delightful RPC era.

All releases are shipped through a CMake pipeline and we still have to support them.

nevack:
🙄

nevack:
@ckerr WDYT?

tearfur (approved):

tearfur:
> I am thinking about total deconstruction of an Xcode project.
> 
> Whole daemon could be rewritten in Swift and managed by launchd.
> 
> And these cli binaries have no place in the upcoming delightful RPC era.
> 
> All releases are shipped through a CMake pipeline and we still have to support them.

What? I'm so confused...

lolgear:
> > I am thinking about total deconstruction of an Xcode project.
> > Whole daemon could be rewritten in Swift and managed by launchd.
> > And these cli binaries have no place in the upcoming delightful RPC era.
> > All releases are shipped through a CMake pipeline and we still have to support them.
> 
> What? I'm so confused...

Current setup enforces to support both Xcode project and CMakeLists even for CLI targets and daemon ( these targets exist in Xcode project ). I don't know who uses them except CI. Now even CI doesn't use them.

Actually, in "RPC era" these CLI binaries can be rewritten in more "sane" way for macOS ( open XPC and ask daemon/core to do stuff ).
Also daemon could be rewritten as "true" daemon with `launchd` support.

tearfur:
> Current setup enforces to support both Xcode project and CMakeLists even for CLI targets and daemon ( these targets exist in Xcode project ).

Yes, that's true.

> Now even CI doesn't use them.

No? I think you're only looking at the `xcodebuild` job, but even this job will still be triggered by code changes within the `daemon/` and `cli/` directory.

> Actually, in "RPC era" these CLI binaries can be rewritten in more "sane" way for macOS ( open XPC and ask daemon/core to do stuff ).
> Also daemon could be rewritten as "true" daemon with `launchd` support.

Again, I imagine the rewritten daemon for `launchd` will only work on macOS, not on Linux and Windows, right? So if we go with your suggestion, we'd still have to keep the existing daemon code for these 2 platforms.

That's essentially adding a completely new `libtransmission` user to our project. I for one think it's a bad idea, we're already stretched thin maintaining our current arsenal.

lolgear:
> Again, I imagine the rewritten daemon for `launchd` will only work on macOS, not on Linux and Windows, right? So if we go with your suggestion, we'd still have to keep the existing daemon code for these 2 platforms.
> 
> That's essentially adding a completely new `libtransmission` user to our project. I for one think it's a bad idea, we're already stretched thin maintaining our current arsenal.

But...
Whole idea is to rewrite everything and put `libtransmission` inside `libtransmission-app`, right?

In my head things work like this: ( for macOS at least )

1. daemon consumes `libtransmission-app` and exposes its XPC-interface ( RPC ).
2. client ( macOS app ) connects to it.
3. cli binary also connects to it.

I don't see additional API consumers.

Issue #245, opened 2026-08-22
remove support for legacy data formats discontinued >= 10 years ago


lolgear:
If this app is a brand new start of transmission then why keep all legacy leavings like previous formats of blocklists or groups?

Keeping apps data compatible is ok but in current state, not in whole 20 years history.

ckerr:
@lolgear this issue is too open-ended and broad.  Can you give me a list of legacy formats that you're suggesting be removed?

lolgear:
@ckerr 

Old groups
https://github.com/retransmission/retransmission/blob/64b2814c86028af5acf0c8254ea6040c22e793a0/macosx/TorrentTableView.mm#L62-L67

Old blacklists
https://github.com/retransmission/retransmission/blob/64b2814c86028af5acf0c8254ea6040c22e793a0/macosx/PrefsController.mm#L123-L147

ckerr:
Transmission 2.12 was from 2010, I think it should be safe to remove that one 😆 

Not sure when the old groups were added? If you want to look at the git log to see when that switched over

lolgear:
@ckerr 

Old groups are from 2008...
111ce452d7940f6cbb648be6ca86fa6634473013

But snippet in TorrentTableView is from 2022, hmmm...

ckerr:
I wouldn't want to remove anything that new. 🤷‍♂️

lolgear:
@ckerr 
At least GroupsController could be cleaned up.

lolgear:
@ckerr 
More.

CreatorWindowController
https://github.com/retransmission/retransmission/blob/64b2814c86028af5acf0c8254ea6040c22e793a0/macosx/CreatorWindowController.mm#L126-L135

GroupsController
https://github.com/retransmission/retransmission/blob/64b2814c86028af5acf0c8254ea6040c22e793a0/macosx/GroupsController.mm#L47-L53

PrefsController
https://github.com/retransmission/retransmission/blob/64b2814c86028af5acf0c8254ea6040c22e793a0/macosx/PrefsController.mm#L123-L146

https://github.com/retransmission/retransmission/blob/64b2814c86028af5acf0c8254ea6040c22e793a0/macosx/PrefsController.mm#L170-L175

Controller
https://github.com/retransmission/retransmission/blob/64b2814c86028af5acf0c8254ea6040c22e793a0/macosx/Controller.mm#L469-L480

ckerr:
[...]

Issue #92, opened 2026-07-09
macOS, Qt, and GTK UIs should share i18n strings


ckerr:
Right now we have inconsistent strings across the three platforms, and translators must do 3x the work.

Let's unify these strings into a single set that can be used in all the UIs.

lolgear:
I would like to migrate macOS localization to modern String Catalogs ( xcstrings ).
It at least minifies amount of files, but also shows possible missing localizations in convenient view.
Also it has more support from Xcode like it automates gathering strings in these catalogs.

### Requirement:

Xcode 15 and later.

### Deployment targets ( macOS versions ):

Deployment targets are not affected, Xcode compiles these catalogs to previous localization format.

### But if

If it affects Cmake builds..

### macOS CMake build removal

It would be nice to remove completely macOS build steps from CMake to enable Swift adoption ( and SwiftUI ).
Can be discussed somewhere else but now cmake is obstacle and it can be freely removed.
`xcodebuild` can be used everywhere instead.

ckerr:
I don't have a fully-informed opinion on this but one thing I do know is I intend to use the C++ library `libtransmission-app` in the Qt, GTK, and macOS apps to (1) make it possible for all three to work as remote controls (issues #82, #83) and (2) reduce code duplication between the three UIs.

So if the macOS client adopts Swift,  it must be [version >= 5.9](https://www.swift.org/documentation/cxx-interop/) so that it can talk to C++ . I don't know if there are any compatibility issues for newer Swift versions and older macOS versions. Before we make any changes, we need to make sure we're not accidentally breaking any supported platforms.

lolgear:
@ckerr 

`xcstrings` is the most promising and optimistic scenario in this case.

`Swift 5.9` requirement is not enough because Swift and C++ interop is implemented only for [macOS 13.0](https://github.com/swiftlang/swift-testing/issues/142) and later.
_Sigh_

ckerr:
@lolgear I checked the macOS 13.0 claim in [your Aug 12 comment](https://github.com/retransmission/retransmission/issues/92#issuecomment-5269035426). Swift 5.9 has that limit. Swift 5.10 removed it.

We can adopt Swift with C++ interop and keep the macOS 12.0 floor. We need Swift 5.10 or newer, which means Xcode 15.3 or newer. I suggest we require Xcode 16, because Swift 6 imports more C++ types.

## Swift 5.10 removed the macOS 13.0 limit

[...]

Issue #83, opened 2026-07-05
macOS app should be able to drive remote daemons


ckerr:
This is going to be similar to #82 but for macOS.

Since the macOS client's code is the most unique, this will require a lot more work. This ticket isn't going to happen overnight 😅 

#82 should be done first, but its design should be done in a way that makes this followup possible.

ckerr:
See also https://github.com/jvacek/TransmissionSwift by @jvacek, who appears to be thinking along the same lines.

jvacek:
Hi all! Happy to be here.

For context: my app was designed _primarily_ for running standalone with a remote daemon, but as far as I'm concerned, I don't see much of a difference between a remote one and a local one.

The app is built with this in mind, and things like "reveal in finder" and "open using default app" already work (if the user enables the "path mapping" in the settings.

I could entertain packaging things in a way where the client could also come with the daemon bundled for running it locally. How much of a paradigm shift are you open to? Or did you have a different proposal in mind for how to incorporate my project?

Either way, as long as my usecases are covered (controlling a remote daemon) I'd be happy to help :)

ckerr:
Everything is TBD right now. Some basic context:

- The Retransmission fork is only about a month old. There aren't many people here yet.

- We need more macOS coders, and you showed up as a person who wrote an entire Transmission RPC client in Swift, soooo I thought I'd put Retransmission on your radar. 👋 
  
- The two first big goals I have for Retransmission dovetail together: (1) replace duplicated code between the Qt, GTK, and macOS clients with some shared codebase, and (2) have better feature parity across the three UIs, e.g. right now only the Qt client can work as an RPC client. These two changes will give users more features and the maintainers less duplicated code to maintain.

- The macOS UI's codebase is a mix of ObjC and ObjC++. There has been discussion about migrating to Swift -- eg @Coeur and @DevilDimon have both suggested this in the upstream repo, and @lolgear has talked about it here -- and it looks like this is possible, since Swift >= 5.9 can speak to C++, but maybe not free? Apparently this requires [macOS >= 13.0](https://github.com/retransmission/retransmission/issues/92#issuecomment-5269035426) and Transmission historically has supported a much longer tail. 🤔 But I haven't investigated this yet, I've been focused on Qt<->GTK unification.

[...]

PR #332, opened 2026-09-12
refactor(mac): remove legacy migrations and hidden prefs from Controller

lolgear:
### Description
This PR addresses **#245** by cleaning up legacy code, ancient migration paths, and hidden preferences specifically inside `macosx/Controller.mm`. 

Focusing solely on the Controller allows for a safer and more isolated code review of these core application-level changes.

### Changes in Controller.mm
* Removed outdated migration logic for negative speed limits (`UploadLimit` / `DownloadLimit` < 0).
* Removed the legacy fallback block for transfer history data migration.
* Removed old hidden macOS preferences (`BindAddressIPv4`, `BindAddressIPv6`, and `PeerSocketTOS`).

### ❓ Question for Maintainers
I have completely removed the hidden preferences for network binding and socket TOS. 

Are these advanced `defaults` command options still considered valuable for macOS users, or is it preferred to drop them entirely as part of this legacy cleanup? If they are still needed, I can easily restore just those specific lines while keeping the ancient version migrations deleted.

### DIFF
--- a/macosx/Controller.mm
+++ b/macosx/Controller.mm
@@ -191,14 +191,6 @@ static auto getSettingsFromNSUserDefaults(NSUserDefaults* defaults)
     settings.insert_or_assign(TR_KEY_speed_limit_up, [defaults integerForKey:@"UploadLimit"]);
     settings.insert_or_assign(TR_KEY_speed_limit_up_enabled, static_cast<bool>([defaults boolForKey:@"CheckUpload"]));
 
-    //hidden prefs
-    if ([defaults objectForKey:@"BindAddressIPv4"]) {
-        settings.insert_or_assign(TR_KEY_bind_address_ipv4, [defaults stringForKey:@"BindAddressIPv4"].UTF8String);
-    }
-    if ([defaults objectForKey:@"BindAddressIPv6"]) {
-        settings.insert_or_assign(TR_KEY_bind_address_ipv6, [defaults stringForKey:@"BindAddressIPv6"].UTF8String);
-    }
-
     settings.insert_or_assign(TR_KEY_blocklist_enabled, static_cast<bool>([defaults boolForKey:@"BlocklistNew"]));
     if ([defaults objectForKey:@"BlocklistURL"]) {
         settings.insert_or_assign(TR_KEY_blocklist_url, [defaults stringForKey:@"BlocklistURL"].UTF8String);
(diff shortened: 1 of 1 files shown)


macosx/Controller.mm
```diff
@@ -236,11 +228,6 @@ static auto getSettingsFromNSUserDefaults(NSUserDefaults* defaults)
         settings.insert_or_assign(TR_KEY_peer_port, [defaults integerForKey:@"BindPort"]);
     }
 
-    //hidden pref
-    if ([defaults objectForKey:@"PeerSocketTOS"]) {
```
lolgear, line 240: Not sure about this one.

macosx/Controller.mm
```diff
@@ -191,14 +191,6 @@ static auto getSettingsFromNSUserDefaults(NSUserDefaults* defaults)
     settings.insert_or_assign(TR_KEY_speed_limit_up, [defaults integerForKey:@"UploadLimit"]);
     settings.insert_or_assign(TR_KEY_speed_limit_up_enabled, static_cast<bool>([defaults boolForKey:@"CheckUpload"]));
 
-    //hidden prefs
-    if ([defaults objectForKey:@"BindAddressIPv4"]) {
```
lolgear, line 195: These settings are not documented, right?
So, it is safe to remove ( and later merge #240 to cover needs ), right?

tearfur: They are documented, just horribly, because they are documented in `snake_case`. In general, all macOS settings are like this.

https://github.com/retransmission/retransmission/blob/19581d5a63bcd2a2c154f1761b2dd09a3a8bb624/docs/Editing-Configuration-Files.md?plain=1#L119-L120

They are related to, but *distinct* from the "bind to interface" requested at #240, so they are not obsolete.

tearfur:
I think you need to understand how settings work in (Re)transmission.

All apps except for macOS use `settings.json` (which is our own convention) to store settings. We have centralised functions `tr_sessionSaveSettings` and `tr_sessionLoadSettings` to save and load them in code. Users can manually edit `settings.json` to change settings too.

The macOS app, for reasons I don't know, chose to integrate with the `defaults` command instead. For this reason, *every* setting needs re-implemented in the macOS app, whereas all other apps share the same code for saving/loading settings.

tearfur (approved):
You've persistently asked for my review, ok fine. I'll say up front that I'm not at all familiar with the macOS client, but these are straightforward enough for even me to understand.

All I've done in this review is `git blame` and see that these logic are added around 2009-2010 to accommodate upgrading from older versions, so probably no one needs them anymore.

ckerr:
> The macOS app, for reasons I don't know, chose to integrate with the `defaults` command instead. For this reason, _every_ setting needs re-implemented in the macOS app, whereas all other apps share the same code for saving/loading settings.

This flavor of inconsistency is IMO Transmission's biggest problem, and why I opened tickets like #82, #83, #92. The plan is to gradually move app-level code into libtransmission-app so that there's one implementation instead of three.

nevack:
I don't understand, why this change is needed after all.
Does this code bother you? If so, how?

tearfur:
> I don't understand, why this change is needed after all. Does this code bother you? If so, how?

That's honestly my question too. It's just I don't see a big problem with it either.

PR #333, opened 2026-09-12
refactor(mac): remove legacy migrations from CreatorWindow

lolgear:
### Description
This PR addresses **#245** by removing very old migration and compatibility blocks from `CreatorWindowController.mm`.

### Changes
* Removed migration logic for single trackers (`CreatorTracker`), which dates back to versions before 1.3 (circa 2008).
* Removed fallback compatibility code for `CreatorLocation` strings used in Transmission 2.5 and earlier.

This is a safe, non-breaking cleanup of legacy code that is no longer relevant.


macosx/CreatorWindowController.mm
```diff
@@ -183,14 +173,6 @@ - (void)awakeFromNib
 
     self.fLocation = [[self.fDefaults URLForKey:@"CreatorLocationURL"]
         URLByAppendingPathComponent:[name stringByAppendingPathExtension:@"torrent"]];
```
ckerr: This branch isn't only 2.5-era migration. `Defaults.plist` still registers `CreatorLocation` = `~/Desktop`, while `CreatorLocationURL` has no default and is only written after a torrent is created. So for new installs, and for anyone who hasn't created a torrent yet, `fLocation` is now nil:

- `updateLocationField` sets the location field's `stringValue` to nil
- the location save panel opens without the suggested `<name>.torrent`
- Create fails with `The directory "(null)" does not currently exist.`

Suggest falling back to the Desktop:

```suggestion
    NSURL* locationDirectory = [self.fDefaults URLForKey:@"CreatorLocationURL"];
    if (!locationDirectory) {
        locationDirectory = [NSFileManager.defaultManager URLsForDirectory:NSDesktopDirectory inDomains:NSUserDomainMask].firstObject;
    }
    self.fLocation = [locationDirectory URLByAppendingPathComponent:[name stringByAppendingPathExtension:@"torrent"]];
```

ckerr (changes requested):

macosx/CreatorWindowController.mm
lolgear: Done.

macosx/CreatorWindowController.mm
```diff
@@ -122,16 +122,6 @@ - (instancetype)initWithHandle:(tr_session*)handle path:(NSURL*)path
...
-            //check for single tracker from versions before 1.3
-            NSString* tracker;
-            if ((tracker = [_fDefaults stringForKey:@"CreatorTracker"])) {
-                [_fDefaults removeObjectForKey:@"CreatorTracker"];
-                if (![tracker isEqualToString:@""]) {
-                    [_fTrackers addObject:tracker];
-                    [_fDefaults setObject:_fTrackers forKey:@"CreatorTrackers"];
-                }
-            }
         }
```
ckerr: Nit: this `if` only creates an empty array now that the migration is gone. Please collapse it into one assignment. It's simpler.

```suggestion
        _fTrackers = [[_fDefaults arrayForKey:@"CreatorTrackers"] mutableCopy] ?: [[NSMutableArray alloc] init];
```

macosx/CreatorWindowController.mm
```diff
@@ -179,18 +169,15 @@ - (void)awakeFromNib
     self.fStatusField.stringValue = status_string;
 
     [self updatePiecesField];
+
```
ckerr: This PR adds an unrelated blank line here. Please remove it.

```suggestion
```

macosx/CreatorWindowController.mm
```diff
@@ -179,18 +169,15 @@ - (void)awakeFromNib
...
-        NSString* location = [self.fDefaults stringForKey:@"CreatorLocation"];
-        self.fLocation = [[NSURL alloc]
-            initFileURLWithPath:[location.stringByExpandingTildeInPath
-                                    stringByAppendingPathComponent:[name stringByAppendingPathExtension:@"torrent"]]];
+    NSURL* locationDirectory = [self.fDefaults URLForKey:@"CreatorLocationURL"];
+    if (!locationDirectory) {
+        locationDirectory = [NSFileManager.defaultManager URLsForDirectory:NSDesktopDirectory inDomains:NSUserDomainMask].firstObject;
     }
+    self.fLocation = [locationDirectory URLByAppendingPathComponent:[name stringByAppendingPathExtension:@"torrent"]];
+
```
ckerr: We can replace the Desktop fallback I suggested earlier with a plist default. Please rename the `CreatorLocation` key in `Defaults.plist` to `CreatorLocationURL` and keep its `~/Desktop` value. `URLForKey:` expands a `~` path string into a file URL. This keeps the default in the plist beside `DownloadFolder` and restores the original lines here. Please check that a fresh install still shows `~/Desktop` in the location field.

```suggestion
    self.fLocation = [[self.fDefaults URLForKey:@"CreatorLocationURL"]
        URLByAppendingPathComponent:[name stringByAppendingPathExtension:@"torrent"]];
```

macosx/Defaults.plist
```diff
@@ -32,8 +32,6 @@
 	<false/>
 	<key>CheckUpload</key>
 	<false/>
```
ckerr: Please register the Desktop default under `CreatorLocationURL` here. It pairs with the suggestion on `CreatorWindowController.mm` line 175.

```suggestion
	<false/>
	<key>CreatorLocationURL</key>
	<string>~/Desktop</string>
```

ckerr (changes requested):
(Opus 5.5 max)

macosx/CreatorWindowController.mm
lolgear: Code golf
```objective-c
_fTrackers = [[_fDefaults arrayForKey:@"CreatorTrackers"] ?: @[] mutableCopy]
```

macosx/CreatorWindowController.mm
lolgear:
```
(lldb) po [self.fDefaults URLForKey:@"CreatorLocationURL"]
file:///Users/dmitry/Desktop/
```

And registration domain shows ~/Desktop/

```objective-c
auto item = [[NSUserDefaults standardUserDefaults] volatileDomainForName:NSRegistrationDomain]
```

PR #335, opened 2026-09-12
refactor: replace deprecated NSNetService with Network framework

lolgear:
### Summary
This PR refactors `BonjourController` by migrating from the deprecated `NSNetService` API to modern Network framework primitives (`nw_listener_t`). Additionally, it cleans up architecture bottlenecks by removing the singleton instance.

### Key Changes
- **Network Framework Migration:** Replaced `NSNetService` with `nw_listener_t`, configured for standard unencrypted TCP/HTTP.
- **Host Name Resolution:** Swapped out `[NSHost currentHost].localizedName` (which was causing sandbox/performance issues or deprecation warnings) with a clean fallback using `SCDynamicStoreCopyComputerName` and `[[NSProcessInfo processInfo] hostName]`.
- **Architecture Cleanup:** Removed the static singleton pattern (`defaultController` / `defaultControllerExists`) from `BonjourController`. It is now properly instantiated and managed as a standard property inside `Controller.mm`.

### Testing Checklist
- [ ] Verify that the Web UI is properly discovered via Bonjour on the local network when "RPCWebDiscovery" is enabled.
- [ ] Verify that disabling the Web UI or quitting the app successfully triggers the `[bonjourController stop]` cancellation handler.

### DIFF
--- a/macosx/BonjourController.h
+++ b/macosx/BonjourController.h
@@ -4,10 +4,7 @@
 
 #import <Foundation/Foundation.h>
 
-@interface BonjourController : NSObject<NSNetServiceDelegate>
-
-@property(nonatomic, class, readonly) BonjourController* defaultController;
-@property(nonatomic, class, readonly) BOOL defaultControllerExists;
+@interface BonjourController : NSObject
 
 - (void)startWithPort:(int)port;
 - (void)stop;
--- a/macosx/BonjourController.mm
+++ b/macosx/BonjourController.mm
@@ -1,68 +1,102 @@
-// This file Copyright © Transmission authors and contributors.
-// It may be used under the MIT (SPDX: MIT) license.
-// License text can be found in the licenses/ folder.
-
 #include "libtransmission/macros.h"
 
 #import "BonjourController.h"
+#import <Network/Network.h>
+#import <SystemConfiguration/SystemConfiguration.h>
 
 static NSUInteger const kBonjourServiceNameMaxLength = 63;
 
 @interface BonjourController ()
 
-@property(nonatomic) NSNetService* fService;
+@property(nonatomic, strong) nw_listener_t fListener;
 
 @end
 
 @implementation BonjourController
 
-static BonjourController* fDefaultController = nil;
-
-+ (BonjourController*)defaultController
-{
-    static dispatch_once_t onceToken;
-    dispatch_once(&onceToken, ^{
-        fDefaultController = [[BonjourController alloc] init];
-    });
-
-    return fDefaultController;
-}
-
-+ (BOOL)defaultControllerExists
-{
-    return fDefaultController != nil;
-}
-
 - (void)startWithPort:(int)port
 {
     [self stop];
 
-    NSMutableString* serviceName = [NSMutableString
-        stringWithFormat:@TR_PROJ_APPNAME_CAPITALIZED " (%@ - %@)", NSUserName(), [NSHost currentHost].localizedName];
+    // 1. Get the localized computer name (Computer Name) instead of using NSHost
+    NSString* computerName = nil;
+    CFStringRef computerNameRef = SCDynamicStoreCopyComputerName(NULL, NULL);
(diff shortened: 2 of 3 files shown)


nevack:
I've tried to do the same in the past and it did not work.
Did you check it? Did you verify with `dns-sd` that the service is being advertised correctly? (`man dns-sd`)

AFAIK
There's no equivalent API in Network.framework for advertising Bonjour services without listening for connections.
Real alternative is to use low level dns-sd APIs.

See: https://developer.apple.com/documentation/technotes/tn3151-choosing-the-right-networking-api#Bonjour-service-discovery

macosx/BonjourController.h
```diff
@@ -4,10 +4,7 @@
 
 #import <Foundation/Foundation.h>
 
-@interface BonjourController : NSObject<NSNetServiceDelegate>
-
-@property(nonatomic, class, readonly) BonjourController* defaultController;
-@property(nonatomic, class, readonly) BOOL defaultControllerExists;
```
ckerr: (Opus 5.5 xhigh) PrefsController.mm:1096-1099 still uses `defaultController` and `defaultControllerExists`, so the macOS CI builds fail to compile. Keep these class properties, or pass the instance that Controller.mm creates into PrefsController. PrefsController needs it because `setRPCEnabled:`, `setRPCPort:`, and the discovery checkbox start and stop the advertisement.

macosx/BonjourController.mm
```diff
@@ -1,68 +1,102 @@
-// This file Copyright © Transmission authors and contributors.
-// It may be used under the MIT (SPDX: MIT) license.
-// License text can be found in the licenses/ folder.
-
 #include "libtransmission/macros.h"
```
ckerr: (Opus 5.5 xhigh) This PR deletes the file's copyright and license header. Restore it.

```suggestion
// This file Copyright © Transmission authors and contributors.
// It may be used under the MIT (SPDX: MIT) license.
// License text can be found in the licenses/ folder.

#include "libtransmission/macros.h"
```

macosx/BonjourController.mm
```diff
@@ -1,68 +1,102 @@
...
 
 #import "BonjourController.h"
+#import <Network/Network.h>
+#import <SystemConfiguration/SystemConfiguration.h>
```
ckerr: (Opus 5.5 xhigh) The app doesn't link Network or SystemConfiguration, so the link will fail on the `nw_*` and `SCDynamicStoreCopyComputerName` symbols. Add both frameworks to `target_link_libraries` in macosx/CMakeLists.txt and to the Transmission target's Frameworks phase in Transmission.xcodeproj. Both builds compile Objective-C++ with `-fno-modules`, so these imports don't link the frameworks by themselves.

macosx/BonjourController.mm
```diff
@@ -1,68 +1,102 @@
...
+    char portString[6];
+    snprintf(portString, sizeof(portString), "%d", port);
 
-- (void)stop
-{
-    [self.fService stop];
-    self.fService = nil;
-}
+    // Create a network listener on the specified port
+    self.fListener = nw_listener_create_with_port(portString, parameters);
```
ckerr: (Opus 5.5 xhigh) This listener binds the RPC port, which libtransmission's web server also binds. Register the service with `DNSServiceRegister()` as nevack suggested, because it advertises a port without binding it. At launch the web server binds first. This listener then either fails and advertises nothing, or it shares the port and cancels every connection it receives. When the prefs window changes the port or enables RPC, the two race to bind it. If this listener wins, the web server can't bind and the Web UI stays down.

ckerr (changes requested):
(Opus 5.5 xhigh)

- This PR lacks the `notes:none` label. Add it, since the refactor intends no user-facing change.
- The title covers only the NSNetService replacement. The PR also removes the singleton and changes the host-name lookup. Split those into their own PRs, or retitle to cover them.

PR #339, opened 2026-09-14
refactor(mac): nuke redundant text field validation and dead code.

lolgear:
### Description
This PR cleans up the macOS UI code by removing redundant text field validation delegates (`control:textShouldBeginEditing:` and `control:didFailToFormatString:`) from `GlobalOptionsPopoverViewController` and `PrefsController`. It also removes the unused `updateRatioStopFieldOld` method.

### Changes
- Removed `fInitialString` property and its delegate tracking logic.
- Removed unused `updateRatioStopFieldOld` method in `PrefsController`.


macosx/PrefsController.mm
```diff
@@ -59,8 +59,6 @@ @interface PrefsController ()<NSWindowRestoration>
 @property(nonatomic) IBOutlet NSView* fRemoteView;
 @property(nonatomic) IBOutlet NSView* fGroupsView;
```
ckerr, line 61: Restore this with the delegate methods below.

```suggestion
@property(nonatomic, copy) NSString* fInitialString;

```

macosx/PrefsController.mm
```diff
@@ -743,23 +734,6 @@ + (NSDate*)timeSumToDate:(NSInteger)sum
     return [NSCalendar.currentCalendar dateFromComponents:comps];
 }
```
ckerr, line 736: These methods are live, not redundant. `PrefsWindow.xib` makes File's Owner the `delegate` of every number-formatted text field, so AppKit calls `control:didFailToFormatString:errorDescription:` whenever input fails the formatter: `abc` in the port field, for example. This handler beeps and restores the previous value. Without it, AppKit still rejects the string but leaves the bad text in place, and the field refuses to end editing until the user corrects it by hand.

Restore them:

```suggestion
- (BOOL)control:(NSControl*)control textShouldBeginEditing:(NSText*)fieldEditor
{
    self.fInitialString = control.stringValue;

    return YES;
}

- (BOOL)control:(NSControl*)control didFailToFormatString:(NSString*)string errorDescription:(NSString*)error
{
    NSBeep();
    if (self.fInitialString) {
        control.stringValue = self.fInitialString;
        self.fInitialString = nil;
    }
    return NO;
}

```

macosx/GlobalOptionsPopoverViewController.mm
```diff
@@ -15,8 +15,6 @@ @interface GlobalOptionsPopoverViewController ()
 @property(nonatomic) IBOutlet NSTextField* fRatioStopField;
 @property(nonatomic) IBOutlet NSTextField* fIdleStopField;
```
ckerr, line 17: Restore this with the delegate methods below.

```suggestion
@property(nonatomic, copy) NSString* fInitialString;

```

macosx/GlobalOptionsPopoverViewController.mm
```diff
@@ -135,21 +133,4 @@ - (IBAction)setIdleStopLimit:(id)sender
     [NSNotificationCenter.defaultCenter postNotificationName:@"UpdateGlobalOptions" object:nil];
 }
```
ckerr, line 135: Same problem. `GlobalOptionsPopover.xib` makes this controller the `delegate` of all four fields, and the speed-limit formatters cap input at 99999. Type `100000`: the old code beeps and reverts; the new code leaves the bad value in the field and never fires `setDownSpeedLimit:`.

Restore them:

```suggestion
- (BOOL)control:(NSControl*)control textShouldBeginEditing:(NSText*)fieldEditor
{
    self.fInitialString = control.stringValue;

    return YES;
}

- (BOOL)control:(NSControl*)control didFailToFormatString:(NSString*)string errorDescription:(NSString*)error
{
    NSBeep();
    if (self.fInitialString) {
        control.stringValue = self.fInitialString;
        self.fInitialString = nil;
    }
    return NO;
}

```

ckerr:
I didn't mean to merge this one, I meant to submit review comments. Apparently I need step away from the keyboard and get some more caffeine. :roll_eyes: 

@lolgear I don't think there's a way to reopen this PR, you'll need to open a new PR. Sorry for the noise.

tearfur:
Water is the way. 😄

PR #341, opened 2026-09-14
refactor(mac): nuke legacy AppleEventManager.

lolgear:
### Description
This PR modernizes how the macOS application handles incoming magnet URLs and file-drop events. 

It replaces the legacy and verbose `NSAppleEventManager` setup (handling `kAEGetURL` and `kAEOpenContents`) with the modern, native native `application:openURLs:` delegate method. Additionally, dead AppleScript-facing `open:` logic has been removed.

### Changes
- Removed manual Apple Event handlers registration from `init` and `applicationDidFinishLaunching:`.
- Implemented `application:openURLs:` to process incoming URLs cleanly.
- Deleted obsolete `handleOpenContentsEvent:replyEvent:` and `open:` methods.


macosx/Controller.mm
```diff
@@ -1036,11 +1023,6 @@ - (void)URLSession:(nonnull NSURLSession*)session
 
 #pragma mark -
 
-- (void)application:(NSApplication*)app openFiles:(NSArray<NSString*>*)filenames
-{
-    [self openFiles:filenames addType:AddTypeManual forcePath:nil];
-}
```
ckerr, line 1042: This change deletes `application:openFiles:`, but `performDragOperation:` still calls it at :3408. Please change that call to:

```objc
            [self openFiles:filesToOpen addType:AddTypeManual forcePath:nil];
```

The app will otherwise throw an unrecognized-selector exception when a user drops a `.torrent` file on the main window. The call still compiles because `NSApplicationDelegate` declares the method as optional.

macosx/Controller.mm
```diff
@@ -1260,20 +1242,6 @@ - (void)openCreatedFile:(NSNotification*)notification
     [self openFiles:@[ dict[@"File"] ] addType:AddTypeCreated forcePath:dict[@"Path"]];
 }
 
-- (void)openFilesWithDict:(NSDictionary*)dictionary
-{
-    [self openFiles:dictionary[@"Filenames"] addType:static_cast<AddType>([dictionary[@"AddType"] intValue]) forcePath:nil];
-}
```
ckerr, line 1266: You deleted `openFilesWithDict:` here, but `Controller.h:35` still declares it. Please delete that declaration too. Clang now warns that the method definition is missing.

ckerr (changes requested):

macosx/Controller.mm
lolgear: Oh... So embarrassing. :(
Replaced it.

macosx/Controller.mm
lolgear: Done.

macosx/Controller.mm
```diff
@@ -916,6 +905,24 @@ - (void)applicationWillTerminate:(NSNotification*)notification
     tr_sessionClose(self.fLib);
 }
 
+- (void)application:(NSApplication*)application openURLs:(NSArray<NSURL*>*)urls
+{
+    // AppKit sends document opens here as file URLs.
+    // It does not call `application:openFiles:` when this method exists.
```
ckerr: This comment explains the file URLs but not the magnet links that reach the `else` branch. Please mention both.

```suggestion
    // AppKit sends both opened files and magnet links here.
    // It does not call `application:openFiles:` when this method exists.
```

ckerr (commented):
I had one minor comment suggestion, but the rest LGTM

ckerr (approved):

PR #342, opened 2026-09-14
cleanup: nuke carbon

lolgear:
### Description
This PR completely eliminates the legacy `Carbon` framework dependency from the macOS build. 

The application used Carbon's `GetCurrentKeyModifiers()` to detect if the Option key was held down on launch. This has been modernized using AppKit's native `NSEvent.modifierFlags`, allowing us to safely remove Carbon from both code imports and `CMakeLists.txt`.

### Changes
- Replaced Carbon-based `GetCurrentKeyModifiers()` with `NSEvent.modifierFlags`.
- Removed `@import Carbon;` and `#import <Carbon/Carbon.h>` from `Controller.mm`.
- Dropped `-framework Carbon` from `macosx/CMakeLists.txt`


macosx/CMakeLists.txt
```diff
@@ -405,7 +405,6 @@ target_link_libraries(${TR_NAME}-mac
         vdkqueue
         ${SPARKLE_FRAMEWORK}
         "-framework AppKit"
-        "-framework Carbon"
```
ckerr, line 408: The Xcode target still links Carbon, and the nightly ships the Xcode build (`nightly.yml:77`). Delete these lines from `Transmission.xcodeproj/project.pbxproj`:

- :141, the `Carbon.framework in Frameworks` build file
- :1001, the `Carbon.framework` file reference
- :1655, its entry in the `Transmission` target's Frameworks phase
- :2017, its entry in the Frameworks group

macosx/Controller.mm
```diff
@@ -916,6 +903,13 @@ - (void)applicationWillTerminate:(NSNotification*)notification
     tr_sessionClose(self.fLib);
 }
 
+- (void)application:(NSApplication*)application openURLs:(NSArray<NSURL*>*)urls
+{
+    for (NSURL* url in urls) {
+        [self openURL:url.absoluteString];
+    }
+}
```
ckerr, line 911: Implementing `application:openURLs:` stops AppKit from calling `application:openFiles:`. `NSApplication.h` says so: "If this is implemented, application:openFiles: and application:openFile: will not be called." AppKit also sends this method every document type that has no `NSDocument` class, which includes `.torrent`.

So Finder double-clicks, Dock drops, and `open -a` now deliver `.torrent` files here as `file://` URLs. `openURL:` treats them as downloads: `NSURLSession` copies the file to a temp path and adds the copy as `AddTypeURL` (:973), not `AddTypeManual`. The user loses:

- the prefilled download folder. `location` is nil (:1064-1067), so the Add window opens a folder sheet, and Cancel drops the torrent.
- the "trash original torrent file" preference (:1025-1027).
- torrent files that lack a `.torrent` extension (:933).

Send file URLs to `openFiles:` in one batch, as `performDragOperation:` does (:3389-3408). Batch them because each `openFiles:` call ends in `fullUpdateUI`.

```suggestion
- (void)application:(NSApplication*)application openURLs:(NSArray<NSURL*>*)urls
{
    // AppKit sends document opens here as file URLs.
    // It does not call `application:openFiles:` when this method exists.
    NSMutableArray<NSString*>* filenames = [NSMutableArray arrayWithCapacity:urls.count];
    for (NSURL* url in urls) {
        if (url.fileURL) {
            [filenames addObject:url.path];
        } else {
            [self openURL:url.absoluteString];
        }
    }

    if (filenames.count > 0) {
        [self openFiles:filenames addType:AddTypeManual forcePath:nil];
    }
}
```

`application:openFiles:` (:1013) then has one caller, :3408. Delete it and call `openFiles:addType:forcePath:` there.

To test, double-click a `.torrent` in Finder with default preferences. The Add window should show the torrent's own folder as the destination.

macosx/Controller.mm
```diff
@@ -1265,15 +1239,6 @@ - (void)openFilesWithDict:(NSDictionary*)dictionary
     [self openFiles:dictionary[@"Filenames"] addType:static_cast<AddType>([dictionary[@"AddType"] intValue]) forcePath:nil];
```
ckerr, line 1239: Optional. `openFilesWithDict:` now has one caller, `openShowSheet:` (:1259). The dictionary existed to fit `performSelectorOnMainThread:withObject:`; a block can capture the values directly. Delete this method and its declaration (`Controller.h:35`), and replace :1259-1265 with:

```objc
            AddType const addType = sender == self.fOpenIgnoreDownloadFolder ? AddTypeShowOptions : AddTypeManual;
            dispatch_async(dispatch_get_main_queue(), ^{
                [self openFiles:filenames addType:addType forcePath:nil];
            });
```

ckerr (changes requested):

macosx/Controller.mm
lolgear: Done in first PR in stack ( unfortunately it is impossible ( not so easy ) to organize PRs in stack in forked repository ).

#341

macosx/Controller.mm
lolgear: Done in #341

macosx/Controller.mm
ckerr: Thanks. I left two follow-up comments on #341. Once #341 merges, please rebase this branch onto main so that it carries only the Carbon commit.

macosx/CMakeLists.txt
ckerr: The Xcode project still links Carbon at this head. Please delete the four `project.pbxproj` lines listed above.

ckerr (approved):

PR #343, opened 2026-09-14
cleanup: nuke sound delegate

lolgear:
### Description
This PR simplifies the sound notification logic in the macOS controller by leveraging native `NSSound` capabilities. 

Instead of manually tracking whether audio is active via `fSoundPlaying` and conforming to `NSSoundDelegate`, the code now directly inspects the built-in `sound.isPlaying` property. This allows us to safely strip out the delegate methods, remove the state variable, and clean up the controller's interface declaration.

### Changes
- Replaced manual `fSoundPlaying` tracking with `sound.isPlaying` checks.
- Removed `fSoundPlaying` property and `sound:didFinishPlaying:` delegate method.
- Dropped `NSSoundDelegate` conformance from `Controller.h`.
- Formatted the long, single-line protocol list in `Controller.h` into a readable, multiline layout.


macosx/Controller.mm
```diff
@@ -2293,10 +2254,8 @@ - (void)torrentFinishedSeeding:(NSNotification*)notification
...
     if (!self.fSoundPlaying && [self.fDefaults boolForKey:@"PlaySeedingSound"]) {
-        NSSound* sound;
-        if ((sound = [NSSound soundNamed:[self.fDefaults stringForKey:@"SeedingSound"]])) {
-            sound.delegate = self;
-            self.fSoundPlaying = YES;
+        NSSound* sound = [NSSound soundNamed:[self.fDefaults stringForKey:@"SeedingSound"]];
+        if (sound && !sound.isPlaying) {
             [sound play];
         }
     }
```
ckerr, line 2261: `fSoundPlaying` no longer exists, so this line does not compile; the macOS CI jobs fail on it. Drop the term.

The `isPlaying` guard is also unnecessary. `-play` returns `NO` and does nothing when the sound is already playing, and a nil `sound` ignores the message.

```suggestion
    if ([self.fDefaults boolForKey:@"PlaySeedingSound"]) {
        [[NSSound soundNamed:[self.fDefaults stringForKey:@"SeedingSound"]] play];
    }
```

macosx/Controller.mm
```diff
@@ -916,6 +901,13 @@ - (void)applicationWillTerminate:(NSNotification*)notification
     tr_sessionClose(self.fLib);
 }
 
+- (void)application:(NSApplication*)application openURLs:(NSArray<NSURL*>*)urls
+{
+    for (NSURL* url in urls) {
+        [self openURL:url.absoluteString];
+    }
+}
```
ckerr, line 909: Once the delegate implements `application:openURLs:`, AppKit stops calling `application:openFiles:` ([docs](https://developer.apple.com/documentation/appkit/nsapplicationdelegate/application(_:open:))). Every `.torrent` opened from Finder, the Dock, or `open` now arrives here as a `file://` URL. `openURL:` treats it as a remote download and adds it as `AddTypeURL`, which:

- ignores the `DeleteOriginalTorrent` preference; only the temp copy is deleted;
- drops the torrent's folder as the default location, forcing the add window when `DownloadLocationConstant` is off;
- rejects a valid torrent that lacks the `.torrent` extension;
- opens N files as N separate tasks, each with its own `fullUpdateUI`.

Send file URLs down the file path, as `performDragOperation:` does:

```suggestion
- (void)application:(NSApplication*)application openURLs:(NSArray<NSURL*>*)urls
{
    NSMutableArray<NSString*>* files = [NSMutableArray arrayWithCapacity:urls.count];
    for (NSURL* url in urls) {
        if (url.isFileURL) {
            [files addObject:url.path];
        } else {
            [self openURL:url.absoluteString];
        }
    }

    if (files.count > 0) {
        [self openFiles:files addType:AddTypeManual forcePath:nil];
    }
}
```

AppKit no longer calls `application:openFiles:`. Remove it, and have `performDragOperation:` call `openFiles:addType:forcePath:` too.

macosx/Controller.mm
```diff
@@ -916,6 +901,13 @@ - (void)applicationWillTerminate:(NSNotification*)notification
     tr_sessionClose(self.fLib);
 }
 
+- (void)application:(NSApplication*)application openURLs:(NSArray<NSURL*>*)urls
+{
+    for (NSURL* url in urls) {
```
ckerr, line 906: The old handler passed the Apple Event's raw string to `openMagnet:`, which tolerates malformed magnets. Now AppKit must first build an `NSURL`. On macOS 12–13, `URLWithString:` rejects unescaped spaces and brackets, so a magnet sent outside LaunchServices (e.g. `osascript -e 'open location "magnet:…"'`) is dropped silently. The case is narrow; test it on 12 or 13 before merging.

macosx/Controller.mm
```diff
@@ -2244,11 +2207,9 @@ - (void)torrentFinishedDownloading:(NSNotification*)notification
...
-        if (!self.fSoundPlaying && [self.fDefaults boolForKey:@"PlayDownloadSound"]) {
-            NSSound* sound;
-            if ((sound = [NSSound soundNamed:[self.fDefaults stringForKey:@"DownloadSound"]])) {
-                sound.delegate = self;
-                self.fSoundPlaying = YES;
+        if ([self.fDefaults boolForKey:@"PlayDownloadSound"]) {
+            NSSound* sound = [NSSound soundNamed:[self.fDefaults stringForKey:@"DownloadSound"]];
+            if (sound && !sound.isPlaying) {
                 [sound play];
             }
```
ckerr, line 2214: Same simplification here.

```suggestion
            [[NSSound soundNamed:[self.fDefaults stringForKey:@"DownloadSound"]] play];
```

Note the behavior change. `fSoundPlaying` kept any two notification sounds from overlapping. `soundNamed:` returns one shared instance per name, so `isPlaying` gates a sound only against itself; the download and seeding sounds (Glass and Submarine by default) can now play at once. If that matters, remember the last sound played and check its `isPlaying`. If it does not, say so in the description.

macosx/CMakeLists.txt
```diff
@@ -405,7 +405,6 @@ target_link_libraries(${TR_NAME}-mac
         vdkqueue
         ${SPARKLE_FRAMEWORK}
         "-framework AppKit"
-        "-framework Carbon"
```
ckerr, line 408: `Transmission.xcodeproj/project.pbxproj` still links `Carbon.framework` (lines 141, 1001, 1655, 2017). Remove those entries too; CI and the nightlies build from the Xcode project.

macosx/Controller.mm
```diff
@@ -1265,15 +1237,6 @@ - (void)openFilesWithDict:(NSDictionary*)dictionary
     [self openFiles:dictionary[@"Filenames"] addType:static_cast<AddType>([dictionary[@"AddType"] intValue]) forcePath:nil];
```
ckerr, line 1237: With `open:` gone, `openShowSheet:` is this method's only caller. Call `openFiles:addType:forcePath:` there directly, and delete `openFilesWithDict:` and its declaration in `Controller.h`.

lolgear:
@ckerr 
Yes, this one is also in PR stack.

1. #341 
2. #342 
3. #343

ckerr:
@lolgear I don't know what to do with that comment. Are you saying all these review comments are addressed by other PRs in the stack...? :smile_cat:

lolgear:
Well, GitHub is a hell.
I prefer to have my own room with my own branches ( even if it is a public fork ).
And it is impossible to do something "similar" to "stacked PRs" or "PRs stack". Not sure how they call this feature.

So these three "nuke" PRs are in one "cleanup: nuke something" stack.

And first one is with AppleEventManager (#341 )

macosx/Controller.mm
ckerr: not a blocker; this is a valid review finding but it's an edge case.

PR #346, opened 2026-09-16
ci: enable dependabot for GHA

tearfur:


.github/dependabot.yml
```diff
@@ -0,0 +1,13 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
```
ckerr: `directory: "/"` covers only `.github/workflows` and a root `action.yml`. The nine composite actions under `.github/actions/*/action.yml` stay untracked — among them `hendrikmuhs/ccache-action@v1.2.23`, the one exact-version pin in the repo. The drift has already begun: workflows run `actions/github-script@v9`, `finalize-ccache` still `@v8`. `.github/actions` triggers the whole build matrix in `actions.yml`, so it is the last tree to leave unwatched.

Use the glob-capable plural:

```suggestion
    directories:
      - "/"
      - "/.github/actions/*"
```

.github/dependabot.yml
```diff
@@ -0,0 +1,13 @@
...
+    schedule:
+      interval: "weekly"
+    labels:
+      - "notes:none"
+      - "type:ci"
+    groups:
+      github-actions:
+        patterns:
+          - "*"
```
ckerr: `patterns: ["*"]` with no `update-types` puts breaking majors in the same pull request as safe bumps. One bad major then blocks the rest, and Dependabot cannot split a group after the fact. Leave majors ungrouped and they arrive one per PR, each reviewable on its own.

```suggestion
    groups:
      github-actions:
        patterns:
          - "*"
        update-types:
          - "minor"
          - "patch"
```

.github/dependabot.yml
```diff
@@ -0,0 +1,13 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
```
ckerr: Every action in the repo is pinned to a floating major tag — `actions/checkout@v6`, `actions/cache@v5`, `vmactions/freebsd-vm@v1`. Dependabot rewrites a ref only when the pinned string changes, so a fix shipped as v6.0.1 under the same `v6` tag is invisible and most weeks yield nothing. The updates that do arrive are the breaking ones.

Pin to full commit SHAs with a trailing `# v6.0.1` comment to track every release; otherwise drop the schedule to `monthly`, which matches a major-only cadence.

.github/dependabot.yml
```diff
@@ -0,0 +1,13 @@
...
+    directory: "/"
+    schedule:
+      interval: "weekly"
```
ckerr: Add a cooldown. Dependabot otherwise proposes a release within days of publication, before a compromised tag is retracted, and merging that one-line diff runs the new code under the `actions: write` token granted in `actions.yml`. `default-days` is supported for the github-actions ecosystem.

```suggestion
    schedule:
      interval: "weekly"
    cooldown:
      default-days: 7
```

.github/dependabot.yml
```diff
@@ -0,0 +1,13 @@
...
+    labels:
+      - "notes:none"
+      - "type:ci"
```
ckerr: Pin the commit prefix. PRs here squash-merge with the title as the subject, and CLAUDE.md requires Conventional Commits. Dependabot titles a grouped update "Bump the github-actions group with 3 updates" and infers a prefix only from bump history this repo does not have.

```suggestion
    commit-message:
      prefix: "ci"
    labels:
      - "notes:none"
      - "type:ci"
```

.github/dependabot.yml
```diff
@@ -0,0 +1,13 @@
+version: 2
+updates:
```
ckerr: Follow-up, not this PR: `web/package-lock.json`, which `webapp.yml` builds with `npm ci`, and the `third-party/` submodules remain unmonitored. The upstream npm stream is visible in this repo's history and has stopped. An `npm` entry on `/web` plus a `gitsubmodule` entry would restore it.

.github/dependabot.yml
```diff
@@ -0,0 +1,13 @@
+version: 2
```
ckerr: State the intent. Every neighbor in `.github/` opens with it — see the header of `pr-new-label.yml` and the step comments in `finalize-ccache/action.yml`. Without it the next reader cannot tell whether the grouping and the label pair were chosen or copied.

```suggestion
# Keep GitHub Actions versions current. Updates are grouped so that review
# happens in one pull request, and labeled as CI changes that stay out of
# the release notes.
version: 2
```

PR #350, opened 2026-09-17
refactor: simplify variant-json.cc

ckerr:
## Description

Cleanups in `libtransmission/variant-json.cc`, one topic per commit.

Parser:

- Remove two debug size assertions. The map assertion failed on duplicate keys, which `tr_variant::Map` merges.
- Always copy strings. RapidJSON passes `copy=false` only when parsing in situ, which this parser never does. `inplace()` now says so.
- Intern keys as they are parsed instead of buffering them until their value arrives.
- Keep the parse stack in a `small::max_size_vector`, so a parse no longer allocates for it.
- Size `prealloc_guess_` to fit every stack depth. This removes two index guards.

Writer:

- Merge the compact and pretty paths into one template. This removes the CTAD workaround for GCC bug 85790 and the deduction guide.
- Pass `std::string` to RapidJSON's `String()` and `Key()` instead of going through a member-pointer helper.
- Share `sorted_entries()` with the benc writer through a module-private `variant-common.h`. `LIBTRANSMISSION_VARIANT_MODULE` gates that header, so the defines nothing needed are gone.

Smaller fixes: `make_map()`, `substr()` for the error excerpt, and corrected comments.

Behavior is unchanged except in assert-enabled builds, which no longer abort on duplicate keys or on an empty key.

Tested on Linux: core, gtk3, gtk4, qt5, and qt6 build and pass their tests, and a Debug clang-tidy build is clean. The Xcode project lists the new header but was not built.

## Checklist

🤖 Generated with [Claude Code](https://claude.com/claude-code)


lolgear:
This PR will trash #224.

libtransmission/variant-json.cc
```diff
@@ -196,12 +178,15 @@ struct json_to_variant_handler : public rapidjson::BaseReaderHandler<> {
     /* A very common pattern is for a container's children to be similar,
      * e.g. they may all be objects with the same set of keys. So when
      * a container is popped off the stack, remember its size to use as
-     * a preallocation heuristic for the next container at that depth. */
-    std::array<size_t, MaxDepth> prealloc_guess_{};
+     * a preallocation heuristic for the next container at that depth.
+     * Indexed by the stack size, which push_stack() caps at MaxDepth. */
+    std::array<size_t, MaxDepth + 1> prealloc_guess_{};
```
tearfur, line 183: I'm not sure what I was thinking back then. I was wasting space on index 0, wasn't I?

Even looking at it now, there's no reason to leave index 0 unused. I think we should index by `size - 1` instead.

ckerr: Wasting an entire 8 bytes! :sob:

tearfur: Wouldn't you know! Imagine how many more torrents we could have downloaded..

tearfur (approved):

PR #352, opened 2026-09-18
fix: determine current_dir using a wanted file

### LINKED ISSUE #300: Move Data may report success while leaving partial torrent data in its old directory

tearfur:
## Description

Fixes #300.
Alternative to #301. I think this PR has less complications than #301 *and* is simpler.

Notes: Fixed edge case where moving torrent data reports no errors but files are left in their old location.

@h1de0x @ckerr what do you think?

## Checklist


h1de0x:
Thanks for putting this together. Choosing a wanted file makes sense for the case where file 0 is unselected, but I don't think we can treat this as a replacement for #301 yet.

The first wanted file can also be absent while another file exists in `downloadDir`. In that case, this still falls back to `incompleteDir`, and Move Data can report success while leaving the existing data behind. It also doesn't handle files already split across both directories.

I'd prefer to keep #301 open while we work through its remaining review questions, and consider this as a possible improvement to `refresh_current_dir()`.

tearfur:
@h1de0x So if I'm grasping the big picture correctly, this PR is going to fix the bug for cases when all factors outside of Retransmission worked perfectly (i.e. no filesystem errors), and your PR is going to cover for cases when things went south (e.g. a move was completed halfway).

If we're able to sort out all of the complications in your PR, I'm all for it.

Edit: Adding more context for the record, IIUC torrent data files are either going to be all in the download dir, or all in the incomplete dir (except when moving of course), hence this PR is enough to fix #300 assuming there are no filesystem errors.

h1de0x:
@tearfur  I don't think filesystem errors are required here. Following the code, this looks possible through normal file selection and a restart:

1. Download only a later file B, leaving an earlier file A unselected and absent. Assume they don't share a piece, so A doesn't get a `.part` file.
2. Let B finish and move successfully to `downloadDir`.
3. Stop the torrent, select A as well, then close and restart the client before A is created.

On startup, `init()` restores `incomplete_dir_` from the session settings. A is now the first wanted file, but it doesn't exist, so this PR still makes `refresh_current_dir()` fall back to `incompleteDir`.

At that point, all existing data is still in `downloadDir`, yet Move Data searches only `incompleteDir` and can report success without moving B. If downloading resumes instead, A gets created in incomplete while B remains in download, producing split data without a failed move.

I haven't run this sequence end to end yet, but the initialization, resume loading, and file creation paths appear to allow it. So I wouldn't draw the distinction between the PRs as normal operation versus filesystem failures.

h1de0x:
That said, this is a fairly specific sequence, and I don't know how often it happens in practice. Your approach does address the original case with much less code. I mainly wanted to clarify that filesystem errors aren't required for the remaining issue, rather than suggest that this edge case alone justifies all the added complexity in #301.

tearfur:
@h1de0x Yeah, you're making some really good points here. This PR can only be called half-baked at best, as there's still plenty of cases it doesn't cover.

AI managed to find a few cases where this PR fails, and among them I think `#3` is exactly what you brought up.

<details>
<summary>gpt-5.6-luna output</summary>

## Finding `#1` premise is reachable

Let:

- `D = download_dir_`
- `I = incomplete_dir_`, with `I` non-empty and normally distinct from `D`
- `A` = the lowest-index wanted file
- `B` = a later wanted file

The relevant state is:

```text
A is wanted
A has no matching path in D or I
B is wanted
B has a matching path under D
```

The PR’s `refresh_current_dir()` only probes `A`. It does not inspect later wanted files:

- `/home/hyy/git_src/transmission/libtransmission/torrent.cc:2285-2305`
- `/home/hyy/git_src/transmission/libtransmission/file-piece-map.h:109-117`

`find_file()` searches `D` first and then `I`, and regards either the normal filename or the `.part` filename as present:

- `/home/hyy/git_src/transmission/libtransmission/torrent.cc:1120-1167`
- `/home/hyy/git_src/transmission/libtransmission/torrent-files.cc:86-105`

Therefore, if `find_file(A)` returns no result, the PR unconditionally selects `I`, even if `B` is present under `D`.

A meaningful split-directory case requires `D` and `I` to be different. The default session settings make them equal unless the user configures a separate incomplete directory:

- `/home/hyy/git_src/transmission/libtransmission/session-settings.h:130-131`

---

## 1. Fresh torrent add with later-file data already in `download_dir`

This is directly possible without a prior restart or filesystem error.

### Path

1. Configure incomplete-directory support with distinct `D` and `I`.
2. Add a multi-file torrent with:
   - `A` wanted but absent;
   - `B` wanted and already present under `D`.
3. Create the torrent.

Initialization sets `I` when incomplete-directory support is enabled, sets the default or builder-selected `D`, applies builder/resume wanted flags, and then refreshes the current directory:

- `/home/hyy/git_src/transmission/libtransmission/torrent.cc:895-903`
- `/home/hyy/git_src/transmission/libtransmission/torrent.cc:919-951`
- `/home/hyy/git_src/transmission/libtransmission/torrent-builder.cc:104-114`

At the final refresh:

```text
first_wanted_file() == A
find_file(A) == none
```

so `current_dir_` becomes `I`:

- `/home/hyy/git_src/transmission/libtransmission/torrent.cc:2290-2304`

The RPC add path can supply both the download directory and wanted/unwanted lists before `tr_torrentNew()`:

- `/home/hyy/git_src/transmission/libtransmission/rpcimpl.cc:1598-1644`
- `/home/hyy/git_src/transmission/libtransmission/rpcimpl.cc:1474-1479`

For a new metainfo torrent, `on_metainfo_completed()` refreshes again and then verifies:

- `/home/hyy/git_src/transmission/libtransmission/torrent.cc:814-825`

The missing `A` causes `is_new_torrent_a_seed()` to return false, but that only causes verification; it does not correct `current_dir_`:

- `/home/hyy/git_src/transmission/libtransmission/torrent.cc:768-799`

Because `B` exists, `has_any_local_data()` is true, so the torrent is not treated as having lost all data:

- `/home/hyy/git_src/transmission/libtransmission/torrent.cc:1169-1175`
- `/home/hyy/git_src/transmission/libtransmission/torrent.cc:167-175`
- `/home/hyy/git_src/transmission/libtransmission/torrent.cc:600-615`

### Minimal preconditions

- At least two files.
- `I` non-empty and distinct from `D`.
- `A` wanted and absent from both directories.
- `B` wanted and present under `D`.
- A normal torrent-add/init path that calls the final refresh.

No selection change, restart, or filesystem failure is required.

---

## 2. Magnet metadata arrival

The same state is reachable when a magnet initially has no metainfo.

Before metadata exists, the refresh code chooses `I` because the `!has_metainfo()` branch is unconditional:

- `/home/hyy/git_src/transmission/libtransmission/torrent.cc:2290-2294`

When metadata arrives, `set_metainfo()` rebuilds the file map and wanted bitfield, then calls `on_metainfo_completed()`:

- `/home/hyy/git_src/transmission/libtransmission/torrent.cc:1029-1041`
- `/home/hyy/git_src/transmission/libtransmission/torrent.cc:802-817`

If, at that point, `A` is absent and `B` is already under `D`, the new refresh selects `I`.

### Minimal preconditions

- Magnet torrent has obtained metainfo.
- `I` is enabled/non-empty.
- `A` is the first wanted file and has no normal or `.part` path.
- `B` is a later wanted file present under `D`.

The pre-metadata state does not itself have a “first wanted file”; the problematic state begins when metadata is installed and the refresh runs.

---

## 3. Normal wanted-selection sequence followed by restart

This is the concrete no-filesystem-error sequence described in the review discussion.

### Step 1: Download a later file while an earlier file is skipped

Start with:

```text
A unwanted
B wanted
```

The first wanted file is therefore `B`, not `A`. If `B` is downloaded under `I`, then it can complete while `A` remains absent.

The DND state can be set through the public C API or RPC:

- `/home/hyy/git_src/transmission/libtransmission/torrent.cc:1826-1830`
- `/home/hyy/git_src/transmission/libtransmission/rpcimpl.cc:1105-1114`
- `/home/hyy/git_src/transmission/libtransmission/rpcimpl.cc:1211-1219`

### Step 2: B completes and moves to D

When the wanted data reaches a done state, `recheck_completeness()` schedules a move from `I` to `D`:

- `/home/hyy/git_src/transmission/libtransmission/torrent.cc:1768-1807`

`tr_torrent_files::move()` searches only its `old_parent`:

- `/home/hyy/git_src/transmission/libtransmission/torrent-files.cc:144-169`

Thus it finds and moves `B` from `I` to `D`. `A` is absent and is skipped. The relocation then clears the per-torrent incomplete directory:

- `/home/hyy/git_src/transmission/libtransmission/torrent.cc:1072-1113`

The result is:

```text
A absent
B present in D
A still unwanted
I cleared for this in-memory torrent
```

This is normal behavior for a partial seed containing only the selected later file.

### Step 3: Select A

Set `A` wanted while the torrent is stopped or still loaded. `set_files_wanted()` changes only the wanted bitfield, invalidates completion sizing, and calls `recheck_completeness()`; it deliberately does not call `refresh_current_dir()`:

- `/home/hyy/git_src/transmission/libtransmission/torrent.cc:1928-1947`

Since the current directory is still `D`, selecting `A` does not immediately switch the current directory to `I`. This is an important distinction: **the selection change alone does not create the bad `current_dir_` value.**

### Step 4: Stop/save and restart

The resume file stores the DND list and destination:

- `/home/hyy/git_src/transmission/libtransmission/resume.cc:188-235`
- `/home/hyy/git_src/transmission/libtransmission/resume.cc:870-908`

On restart, `tr_torrent::init()` reconstructs `I` from the enabled session setting, loads the wanted flags, and performs the final refresh:

- `/home/hyy/git_src/transmission/libtransmission/torrent.cc:895-903`
- `/home/hyy/git_src/transmission/libtransmission/torrent.cc:937-951`

The resumed wanted state is now:

```text
A wanted
B wanted
```

`A` is still absent, while `B` remains in `D`. The final refresh therefore selects `I`.

This sequence needs no failed move and no external filesystem intervention. To guarantee that downloading `B` did not create an `A` path, it is sufficient for `A` and `B` not to share a piece, or otherwise for no write/preallocation to have created `A`.

</details>

I've opened #353 in place of this PR.

PR #353, opened 2026-09-19
fix: determine current_dir using first found file

### LINKED ISSUE #300: Move Data may report success while leaving partial torrent data in its old directory

tearfur:
## Description

Fixes #300.
Supersedes #352.

Change `tr_torrent::refresh_current_dir` so that the new `current_dir` value is determined by the first data file found.

This can become quite expensive if there are a lot of files in the torrent and only files towards the end of the file list existed in the filesystem, but it is not a big problem because `tr_torrent::refresh_current_dir` is not a hot path. 

Notes: Fixed edge case where moving torrent data reports no errors but files are left in their old location.

## Checklist


h1de0x:
@tearfur thanks for looking into those cases. I ended up with the same approach: use the first existing file, regardless of whether it is selected for download.

I've prepared an update for #301 that also avoids some duplicate filesystem work. During resume loading, `load_checked_pieces()` already searches every file, so it retains the first matching root and passes that result to `refresh_current_dir()`. It also records when nothing was found, so that case doesn't trigger another full search. The result is temporary and isn't reused for later location changes.

The update also removes the second `refresh_current_dir()` call when adding a new `.torrent`, while keeping the refresh when magnet metadata arrives.

Although this isn't a hot path during normal operation, it runs during torrent initialization, so the extra filesystem work can add up across a large session. With roughly 15,000 torrents in my session, I'd rather reuse the work we're already doing. I haven't measured the startup improvement yet, but the duplicate searches are avoidable.

ckerr (approved):

tearfur:
On a brief look, seems like #301 now has made some changes that are equivalent to this PR.

I'll do a proper code review over there soon.

PR #355, opened 2026-09-20
refactor(test): extract common test code to fixture methods

tearfur:
## Description

Some test refactors extracted from a change that I've abandoned. I thought they are good enough for a PR on their own.

## Checklist


tests/libtransmission/inout-test.cc
```diff
@@ -21,27 +21,27 @@
 
 #include "test-fixtures.h"
```
ckerr, line 18: This file no longer uses `<chrono>`, `<future>`, `<memory>` or `<utility>`. Please delete those includes.

tests/libtransmission/inout-test.cc
```diff
@@ -21,27 +21,27 @@
...
-        auto future = promise->get_future();
-        session_->run_in_session_thread([fn = std::move(fn), promise]() {
-            fn();
-            promise->set_value();
-        });
-        ASSERT_EQ(std::future_status::ready, future.wait_for(std::chrono::milliseconds{ MaxWaitMsec }));
-    };
-
     auto err = tr_error_code_t{};
     auto const write_block = [session = session_, tor, &err]() {
```
ckerr: This comment described the lambda that this PR deletes, we should move the comment to `write_block`.

That lambda calls `tr_ioWrite()`. We can drop "block until it's done" because the helper's comment now says so.

```suggestion
    auto err = tr_error_code_t{};
    // tr_ioWrite() must run in the session thread
    auto const write_block = [session = session_, tor, &err]() {
```

tests/libtransmission/move-test.cc
```diff
@@ -37,10 +37,49 @@ class IncompleteDirTest
...
+            data.block = block_index;
+            session_->run_in_session_thread(TestIncompleteDirThreadfunc, data);
+
+            // save_block() may return before the write finishes
+            auto const test = [tor, block_index]() {
+                return tor->has_block(block_index);
+            };
+            EXPECT_TRUE(waitFor(test, MaxWaitMsec));
+        }
+    }
```
ckerr: This struct, factory and static lambda only carry `tor` and a block index to the session thread. We can remove all three if we use a capturing lambda in `blockingRunInSessionThread()`. The suggestion builds the zero buffer inside the lambda so that nothing dangles if the wait times out.

```suggestion
    void completeBlockSpan(tr_torrent* const tor, tr_block_span_t const span)
    {
        auto const [begin, end] = span;
        for (tr_block_index_t block_index = begin; block_index < end; ++block_index) {
            auto const save_zeroes = [tor, block_index]() {
                if (tor->on_block_received(block_index)) {
                    auto const zeroes = std::vector<uint8_t>(tr_block_info::BlockSize);
                    tor->save_block(block_index, std::make_unique<tr::LocalData::BlockData>(zeroes));
                }
            };
            blockingRunInSessionThread(save_zeroes, std::chrono::milliseconds{ MaxWaitMsec });

            // save_block() may return before the write finishes
            auto const test = [tor, block_index]() {
                return tor->has_block(block_index);
            };
            EXPECT_TRUE(waitFor(test, MaxWaitMsec));
        }
    }
```

tests/libtransmission/test-fixtures.h
```diff
@@ -79,7 +80,7 @@ inline bool waitFor(std::function<bool()> const& test, std::chrono::milliseconds
     }
 }
 
-inline bool waitFor(std::function<bool()> const& test, int msec)
+inline bool waitFor(std::function<bool()> const& test, std::chrono::milliseconds::rep const msec)
```
ckerr, line 83: This line changes `int` to `rep`, but every integer caller still passes an `int`. Please revert it or move it to its own PR.

tests/libtransmission/test-fixtures.h
```diff
@@ -445,6 +446,23 @@ class SessionTest : public SandboxedTest
...
+    // Runs `func` on the session thread and waits for it to finish.
+    void blockingRunInSessionThread(std::function<void()> const& func, std::chrono::milliseconds const msec)
+    {
+        auto promise = std::promise<void>{};
+        auto const future = promise.get_future();
+        session_->run_in_session_thread([&func, &promise]() {
+            func();
+            promise.set_value();
+        });
+        ASSERT_EQ(future.wait_for(msec), std::future_status::ready);
```
ckerr: On timeout this returns while the lambda is still queued. The lambda then calls a dead `func` — often a temporary `std::function` built at the call site — and sets a dead `promise`. The code this replaces in inout-test.cc avoided that by owning both. Do the same here. This also restores expected-first argument order.

```suggestion
        // On timeout the queued lambda outlives this call, so it owns what it uses.
        auto const promise = std::make_shared<std::promise<void>>();
        auto const future = promise->get_future();
        session_->run_in_session_thread([func, promise]() {
            func();
            promise->set_value();
        });
        ASSERT_EQ(std::future_status::ready, future.wait_for(msec));
```

tests/libtransmission/test-fixtures.h
```diff
@@ -445,6 +446,23 @@ class SessionTest : public SandboxedTest
...
+    }
+
+    void blockingRunInSessionThread(std::function<void()> const& func, std::chrono::milliseconds::rep const msec)
+    {
+        blockingRunInSessionThread(func, std::chrono::milliseconds{ msec });
+    }
```
ckerr: Two fixtures now define the same `inSessionThread` wrapper only to pass a 5 second timeout. We can delete both wrappers and this overload if we give the overload above a default timeout.

```cpp
void blockingRunInSessionThread(std::function<void()> const& func, std::chrono::milliseconds const msec = 5s)
```

Then `InOutTest` can go back to a `using` alias. The `MaxWaitMsec` constants in inout-test.cc and torrent-disk-io-test.cc can go too. If you also name the base method `inSessionThread`, the existing call sites keep working.

tests/libtransmission/test-fixtures.h
```diff
@@ -445,6 +446,23 @@ class SessionTest : public SandboxedTest
...
 
+    // Runs `func` on the session thread and waits for it to finish.
+    void blockingRunInSessionThread(std::function<void()> const& func, std::chrono::milliseconds const msec)
+    {
+        auto promise = std::promise<void>{};
+        auto const future = promise.get_future();
+        session_->run_in_session_thread([&func, &promise]() {
+            func();
+            promise.set_value();
+        });
```
ckerr: The queued lambda captures `func` and `promise` by reference, and it can outlive both. Please capture `func` by value and hold the promise in a `shared_ptr`, as the old inout-test.cc code did. If the wait times out, this function returns and destroys both objects. The session thread then runs the lambda and uses them after they are gone.

```suggestion
        // The queued lambda owns `func` and the promise.
        // It can outlive this call if the wait times out.
        auto const promise = std::make_shared<std::promise<void>>();
        auto const future = promise->get_future();
        session_->run_in_session_thread([func, promise]() {
            func();
            promise->set_value();
        });
```

tests/libtransmission/test-fixtures.h
ckerr: This assertion puts the actual value first, but the rest of this header puts the expected value first. Please swap the arguments.

```suggestion
        ASSERT_EQ(std::future_status::ready, future.wait_for(msec));
```

ckerr (commented):
I had a few nits eg removing unused `#include` calls and simplifications suggestions, but mostly LGTM.

tests/libtransmission/test-fixtures.h
tearfur: `rep` is `int64_t`, the underlying type of `std::chrono::milliseconds`, so existing callers will continue working.

I just didn't want our API to artificially limit the range of values that can be passed in.

tests/libtransmission/test-fixtures.h
tearfur: 0bee81f8a9a3021d6caa96fcd189ee53396fd144

tests/libtransmission/test-fixtures.h
tearfur: dd3344be9a99b11f2c5e0f313e4b52aa180a4e8c

tests/libtransmission/move-test.cc
tearfur: d090e971bb6a2a44f4bd73a3d6c6014e6dda7d35 I've made a few changes on top of your suggestion.

1. Use `TrBlockInfo` instead of `tr_block_info::BlockSize`. We can include `constants.h` instead of `block-info.h` this way, which is lighter.
2. We are waiting for `test` to return true anyway, so running `save_zeroes` in session thread don't need to be blocking.

tests/libtransmission/inout-test.cc
tearfur: 4c757c135a9384a501eefa64d7d79c04ee4850d0

tests/libtransmission/inout-test.cc
tearfur: 033ca477ba707635268414a2e892fa50417f9a4b

ckerr (approved):

PR #356, opened 2026-09-20
cmake install: add missing public headers

lolgear:
## Description
This PR updates the list of public headers installed by `libtransmission` in `CMakeLists.txt`. 

The changes include:
* **Adding missing headers** (such as `announce-list.h`, `block-info.h`, `constants.h`, `types.h`, etc.) to ensure that dependent projects have access to all necessary public definitions.

## Type of Change
- [x] Refactoring / Build system update (non-breaking change)

## Checklist
- [x] Headers are sorted alphabetically
- [x] Verified that all newly added headers exist in the project directory


ckerr (approved):

libtransmission/CMakeLists.txt
```diff
@@ -316,22 +316,30 @@ if(INSTALL_LIB)
...
+            block-info.h
+            constants.h
+            digest.h
             error-types.h
             error.h
             file.h
             log.h
+            macros.h
             makemeta.h
+            net.h
```
tearfur, line 329: Like this one.

https://github.com/retransmission/retransmission/blob/4d608b9ec78c41ef58694111e406c501c9b3e234/libtransmission/net.h#L7-L9

tearfur (commented):
I was too slow, I meant to stop this PR.

How did you come up with this list? Some of these aren't meant to be included outside of libtransmission.

lolgear:
@tearfur 

Well..

I'm playing with Swift support via Swift package manager ( XCFramework wrapper for static library ).

1. Build library via cmake install.
2. Wrap it into XCFramework.
3. Put it into Swift package manager.

After that I also added modulemap and press "Build" in additional SPM target..

Long story short.

Xcode is failing building artifacts for Swift because of these includes:

[ERROR] log.h:15 -> includes private header 'libtransmission/shared-string.h'
[ERROR] log.h:16 -> includes private header 'libtransmission/types.h'
[ERROR] makemeta.h:18 -> includes private header 'libtransmission/announce-list.h'
[ERROR] makemeta.h:19 -> includes private header 'libtransmission/block-info.h'
[ERROR] makemeta.h:22 -> includes private header 'libtransmission/torrent-files.h'
[ERROR] makemeta.h:23 -> includes private header 'libtransmission/types.h'
[ERROR] tr-buffer.h:18 -> includes private header 'libtransmission/net.h'
[ERROR] transmission.h:22 -> includes private header 'libtransmission/types.h'
[ERROR] web-utils.h:15 -> includes private header 'libtransmission/digest.h'

[ERROR] block-info.h:10 -> includes private header 'libtransmission/constants.h'

libtransmission/CMakeLists.txt
lolgear: tr_buffer.h:20

ckerr:
TBH I've been thinking we should just remove the whole header installation step altogether. We have a monorepo so who are we installing the headers for?

PR #357, opened 2026-09-20
build(mac): SPM and XCFramework

lolgear:
## Description

This PR illustrates an approach with SPM package and XCFramework that wraps static library.
Static library is produced by cmake install.

## Steps

1. Build static library and its third-party dependencies.
2. Build fat binary with all libraries included.
3. Put everything into XCFramework.
4. Expose this XCFramework as target in SPM package.
5. Integrate this SPM package into xcodeproj.

One thing to remember:
We still need to have good header search paths, because library has includes of other libraries in its public interface.


regenerate.sh
```diff
@@ -0,0 +1,156 @@
...
+rm -rf build_cmake
+rm -rf build_output
+
+mkdir -p build_cmake
+cd build_cmake
+
+echo "🛠 CMake Configuration..."
+
+cmake -G "Ninja" .. \
+  -DCMAKE_BUILD_TYPE=Debug \
```
lolgear, line 126: It works only for `Debug` now.

lolgear: Not sure that current setup requires Release configuration for Xcode builds.

But A.I. suggests something like:

```swift
// swift-tools-version:5.9
import PackageDescription

let package = Package(
    name: "YourLibraryWrapper",
    products: [
        .library(name: "YourLibrary", targets: ["YourLibraryTarget"])
    ],
    targets: [
        // Umbrella target
        .target(
            name: "YourLibraryTarget",
            dependencies: [
                .target(name: "YourLibRelease", condition: .when(configuration: .release)),
                .target(name: "YourLibDebug", condition: .when(configuration: .debug))
            ]
        ),
        // Binary frameworks for each configuration
        .binaryTarget(
            name: "YourLibRelease",
            path: "build/Release/YourLib.xcframework"
        ),
        .binaryTarget(
            name: "YourLibDebug",
            path: "build/Debug/YourLib.xcframework"
        )
    ]
)
```

lolgear:
For Swift we need to expose interface via module map.

```map
module LibTransmission {
    umbrella "."
    
    export *
    module * { export * }
}
```

But even with this module map we need to configure paths to find headers ( of third-party libraries ).

Transmission.xcodeproj/project.pbxproj
```diff
@@ -738,25 +219,11 @@
 /* End PBXCopyFilesBuildPhase section */
 
 /* Begin PBXFileReference section */
-		0A6169A50FE5C9A200C66CE6 /* bitfield.cc */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.cpp.cpp; path = bitfield.cc; sourceTree = "<group>"; };
-		0A6169A60FE5C9A200C66CE6 /* bitfield.h */ = {isa = PBXFileReference; explicitFileType = sourcecode.cpp.h; path = bitfield.h; sourceTree = "<group>"; };
-		0A89346B736DBCF81F3A4851 /* torrent-metainfo.cc */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.cpp.cpp; path = "torrent-metainfo.cc"; sourceTree = "<group>"; };
-		0A89346B736DBCF81F3A4853 /* torrent-metainfo.h */ = {isa = PBXFileReference; explicitFileType = sourcecode.cpp.h; path = "torrent-metainfo.h"; sourceTree = "<group>"; };
+		0A5783AD30532E7B007CDDC9 /* TransmissionSPM */ = {isa = PBXFileReference; lastKnownFileType = wrapper; path = TransmissionSPM; sourceTree = "<group>"; };
```
ckerr: (Opus 5.5 xhigh) This file reference points Xcode at `TransmissionSPM/`, which only regenerate.sh creates. Please install Ninja and run regenerate.sh in the `macos-xcodebuild-universal-git-26` job before `xcodebuild`, and add the step to the Xcode section of docs/Building-Transmission.md. Until then, a fresh clone fails with "Missing package product 'LibTransmission'", as that CI job does now. The nightly macOS build comes from that job.

Transmission.xcodeproj/project.pbxproj
```diff
@@ -1580,74 +692,35 @@
...
-			buildActionMask = 2147483647;
-			files = (
-			);
-			runOnlyForDeploymentPostprocessing = 0;
-		};
 		4DDBB71709E16BAE00284745 /* Frameworks */ = {
 			isa = PBXFrameworksBuildPhase;
 			buildActionMask = 2147483647;
 			files = (
-				4D9A2BF009E16D21002D0FF9 /* libtransmission.a in Frameworks */,
```
ckerr: (Opus 5.5 xhigh) This PR removes `libtransmission.a` from transmissioncli, transmission-daemon, transmission-remote, transmission-create, transmission-edit, and transmission-show, so each of them now fails to link. Please delete these six targets, or link the LibTransmission product into each one. CMake already builds these tools in CI. If the targets go, please delete `third-party/macosx-libevent-*.h` too, because only the deleted "Copy libevent headers" phase used them. If transmission-daemon stays, please restore that phase for it.

regenerate.sh
```diff
@@ -0,0 +1,156 @@
+# : << DOC
```
ckerr: (Opus 5.5 xhigh) Minor: the `#!/bin/bash` lines sit inside the function bodies, so the file has no shebang. Please move one to line 1 and mark the file executable. As written, `./regenerate.sh` fails with a permission error, and `zsh regenerate.sh` passes `$THIRD_PARTY_LIBS` to libtool as one argument.

```suggestion
#!/bin/bash
# : << DOC
```

regenerate.sh
```diff
@@ -0,0 +1,156 @@
...
+# Generated (CMake) xcodeproj is something very tasty and jealous. Zero maintenance.
+
+# DOC
+
+function install_as_spm {
+#!/bin/bash
+set -e
+
+# Change to the repository root directory
+cd "$(dirname "$0")"
```
ckerr: (Opus 5.5 xhigh) This PR deletes the "Generate version file" phase, which was the only step that wrote `libtransmission/version.h`. Please run update-version-h.sh here. Xcode still reads that file when it preprocesses the Info.plist of Transmission and QuickLookExtension, and `macosx/AboutWindowController.mm` includes it. A fresh clone fails to build without it, and an older checkout keeps a stale version number. The app's version then also matches the library it links, since both come from the same run.

```suggestion
cd "$(dirname "$0")"
sh update-version-h.sh
```

regenerate.sh
```diff
@@ -0,0 +1,156 @@
...
+
+echo "🧹 Cleaning up old build..."
+rm -rf build_cmake
+rm -rf transmission-spm
```
ckerr: (Opus 5.5 xhigh) Minor: the script cleans and creates `transmission-spm/`, but it writes the package to `TransmissionSPM/`. Please use `TransmissionSPM` here and on lines 77 and 107. As written, each run starts with the previous xcframework still in place and leaves an empty `transmission-spm/` behind.

```suggestion
rm -rf TransmissionSPM
```

regenerate.sh
```diff
@@ -0,0 +1,156 @@
...
+mkdir -p build_cmake
+cd build_cmake
+
+echo "🛠 Configuring CMake (Debug, enabling INSTALL_LIB)..."
+# Enable INSTALL_LIB and explicitly set the installation directory to the temporary build_install folder"
+
+cmake -G "Ninja" .. \
+  -DCMAKE_BUILD_TYPE=Debug \
+  -DCMAKE_OSX_ARCHITECTURES="arm64;x86_64" \
+  -DENABLE_IPO=OFF \
```
ckerr: (Opus 5.5 xhigh) This configure step builds libtransmission without `BUILD_MAC_CLIENT`, which the Xcode project defined for every target. Please add it here. Without it, `tr_getWebClientDir()` in libtransmission/platform.cc searches XDG directories instead of `Transmission.app/Contents/Resources/public_html`. With "Enable remote access" on, "Launch Web Interface" then shows "Couldn't find Retransmission's web interface files!"

```suggestion
  -DENABLE_IPO=OFF \
  -DCMAKE_CXX_FLAGS=-DBUILD_MAC_CLIENT \
```

regenerate.sh
```diff
@@ -0,0 +1,156 @@
...
+  -DINSTALL_LIB=ON \
+  -DCMAKE_INSTALL_PREFIX=../build_install \
+  -DENABLE_DAEMON=OFF -DENABLE_GTK=OFF -DENABLE_QT=OFF -DENABLE_MAC=OFF \
+  -DENABLE_UTILS=OFF -DENABLE_CLI=OFF -DENABLE_TESTS=OFF -DREBUILD_WEB=OFF
```
ckerr: (Opus 5.5 xhigh) This configure step lets CMake pick up Homebrew packages, which the Xcode build never used. Please pass `-DUSE_SYSTEM_DEFAULT=OFF -DENABLE_NLS=OFF` to match the Xcode build. USE_SYSTEM_DEFAULT defaults to AUTO, so CMake uses any Homebrew copy of libevent, libpsl, libarchive, and similar libraries that it finds, and those stay out of libtransmission_combined.a. ENABLE_NLS defaults to ON, so a Homebrew gettext makes libtransmission call libintl, and the Xcode app links without libintl. Either way, the app fails to link.

```suggestion
  -DENABLE_UTILS=OFF -DENABLE_CLI=OFF -DENABLE_TESTS=OFF -DREBUILD_WEB=OFF \
  -DUSE_SYSTEM_DEFAULT=OFF -DENABLE_NLS=OFF
```

regenerate.sh
```diff
@@ -0,0 +1,156 @@
...
+
+# Perform native install — CMake will automatically collect the library and headers into the build_install folder
+cmake --install .
+
+cd ..
+
+echo "📦 Merging native core library with external third-party dependencies..."
+
+# The main core library is now guaranteed to be in build_install/lib/libtransmission.a# We just need to mix in the third-party libraries (dht, libevent, etc.) with it
+THIRD_PARTY_LIBS=$(find build_cmake/third-party -name "*.a")
```
ckerr: (Opus 5.5 xhigh) Minor: this `find` returns each ExternalProject library twice, once from its build tree and once from its install prefix. Please skip the build-tree copies. As written, libtool merges two copies each of libarchive, libdht, libminiupnpc, libnatpmp, and libpsl.

```suggestion
THIRD_PARTY_LIBS=$(find build_cmake/third-party -name "*.a" -not -path "*/src/_*-build/*")
```

ckerr (commented):
(Opus 5.5 xhigh)

Please fix everything except the three comments marked Minor before this leaves draft. Until then, a fresh clone fails to build.

**Design question.** Xcode no longer rebuilds libtransmission. The app compiles against the headers in the source tree but links whatever archive regenerate.sh built last. After each libtransmission change, a developer must rerun the script, and it rebuilds every dependency from scratch. If a developer skips the rerun after a change to a struct like `tr_stat`, the app and the library disagree on its layout, and the build still succeeds. Could Xcode run an incremental `cmake --build` from a Run Script phase or an External Build System target? Each Xcode configuration could then build its matching CMake build type.

**Debug-only library.** To answer your question in the regenerate.sh thread: Release builds do need a Release library. With the Debug archive, the nightly ships an unoptimized libtransmission, and a Release build ships one with `TR_ASSERT` enabled. SwiftPM's `TargetDependencyCondition` accepts only platforms and traits, so the Package.swift from that thread fails to compile.

**Rebase.** `project.pbxproj` conflicts with main, which has changed it heavily since this branch, mostly in #224. Please redo the Xcode edits on current main rather than merging the conflict by hand.

**Other notes**
- The header of regenerate.sh says the SPM option doesn't work for the QuickLook extension, but this PR links LibTransmission into QuickLookExtension. Please confirm the extension still previews a torrent, or update the header.
- Only the deleted archive target used `third-party/libarchive-config/`. Please delete it too.
- Please delete the unused `build_static_library` and move the comparison of options from the header of regenerate.sh into the PR description.
- Please add the `notes:none` label, since users see no change.
- The Windows x64 failure comes from a libarchive configure error and is unrelated to this PR.

PR #366, opened 2026-09-21
fix(gtk): open torrent folder when it contains a single file

h1de0x:
## Description

The GTK Open Folder action opened the download directory for single-file torrents, even when the file was inside the torrent's own folder. This change opens the torrent folder in that case.

Notes: The GTK client now opens the torrent's own folder when a single-file torrent uses a directory structure.

## Checklist


gtk/Session.cc
```diff
@@ -1228,7 +1228,7 @@ void Session::open_folder(tr_torrent_id_t torrent_id) const
     if (auto const* tor = find_torrent(torrent_id); tor != nullptr) {
         auto const current_dir = tr_torrentGetCurrentDir(tor);
 
-        if (tr_torrentFileCount(tor) == 1) {
+        if (tr_torrentFileCount(tor) == 1 && !tr_torrentView(tor).is_folder) {
             gtr_open_file(current_dir);
         } else {
```
ckerr: This regresses single-file torrents whose `name` contains a `/`, e.g. `Show/ep1.mkv`. The file is stored at `Show/ep1.mkv`, so `is_folder` is true, and the `else` branch calls `gtr_open_file(current_dir, "Show/ep1.mkv")`. That opens the file itself in its default app, or, while the download is still a `.part` file, a path that doesn't exist. Before this change it opened the download dir.

Taking the top-level folder from the file's own subpath handles that case as well as the one this PR targets:

```suggestion
        if (tr_torrentFileCount(tor) == 1) {
            auto const subpath = std::string_view{ tr_torrentFile(tor, 0).name };
            if (auto const slash = subpath.find('/'); slash != std::string_view::npos) {
                gtr_open_file(current_dir, subpath.substr(0, slash));
            } else {
                gtr_open_file(current_dir);
            }
        } else {
```

h1de0x:
Thanks for catching this! I overlooked that `is_folder` being true doesn't guarantee that the torrent name refers to a directory.

ckerr (approved):

